PR #237, opened 2013-09-29
Fixed [904706] Thimble pages not scrollable on iOS

stanica:
This patch makes Thimble makes scrollable on iPads without introducing multiple scrollbars on desktop browsers. iPhone support was not included.


sedge:
Closing as part of #622

PR #417, opened 2014-04-17
Fix bug 971878 - Add Preset Makerstrap templates

mlveis:
Added a button named "New from template" and when the user selects it, it links to a bunch of makerstrap templates.


sedge:
This is no longer in scope for thimble

PR #428, opened 2014-05-30
Bug 1015639 - Update webmaker-auth-client

cadecairos:


humphd:
Given that we're updating to a newer login system based on the new Webmaker ID server, I'm going to close this. See #497.

PR #429, opened 2014-06-02
Bug 1019066 - store webmaker userid in project records, include migratio...

cadecairos:
...n sql and js for old projects


humphd:
@cadecairos @jbuck can you two comment on this, and whether we need to update and carry this change forward?

jbuck:
I'd recommend the idea for your version of thimble, but it's not worth applying this patch.

PR #446, opened 2014-10-13
Bug 1068585 - Add 'getvalue' action to tutorial protocol.

toolness:
This is a potential solution to [bug 1068585](https://bugzilla.mozilla.org/show_bug.cgi?id=1068585). If the tutorial pane sends its parent window a JSON-stringified message with the following content:

```json
{
  "type": "tutorial",
  "action": "getvalue"
}
```

The parent window will immediately respond with the following JSON-stringified message:

```json
{
  "type": "tutorial",
  "action": "value",
  "value": "contents of codeMirror.getValue() will go here"
}
```


sedge:
Closing as part of #622

PR #472, opened 2015-03-02
Use minified CSS/JS resources from CDNs

humphd:
I'm working on making our Brackets/Thimble smaller, and there are a bunch of things that we can do just to Thimble.  This is the first, and it's a pretty easy fix.


jbuck:
@humphd r+

PR #495, opened 2015-05-06
uplifted versions + fixes for things that broke

### LINKED ISSUE #491: Thimble is incompatible with node 0.12 at the moment
There are a lot of dependencies that are pegged to versions that are incompatible with node 0.12, meaning that people on modern setups cannot run Thimble.

Part of the uplifting process here can be done by updating the dependencies in the package, but some of those will result in needing to change the actual app code due to changes in express 4 (although from testing, express 3.22.2 seems node 0.12 compatible at least and so uplifting express can be done as followup)

Pomax:
fixes #491


Pomax:
@sedge if you can verify this works, this might also be a template for lifting up bramble from requiring node 0.10.x to being runnable on node 0.12.x instead

humphd:
@Pomax++ for fixing things here.  <3

sedge:
I'm getting a `500` on `http://localhost:3500/stylesheets/userbar-overrides.css ` resulting in this:

screen shot 2015-05-07 at 11 17 28 am

sedge:
I'm seeing two errors.

First, a failure on `node-gyp rebuild` while trying to install this dependancy chain:

`node-expat@2.0.0 install /Users/k.sedge/Documents/Projects/webmaker-suite/thimble.webmaker.org/node_modules/webmaker-download-locales/node_modules/xml2json/node_modules/node-expat`

Second, a `500` on `GET /stylesheets/userbar-overrides.css` which is probably related to a less compilation error:

```bash
 LESS File error : '../../bower_components/bootstrap/less/variables.less' wasn't found. Tried - bower_components/bower_components/bootstrap/less/variables.less,../../bower_components/bootstrap/less/variables.less
  LESS File : bower_components/makerstrap/less/makerstrap.less 9:0
GET /stylesheets/userbar-overrides.css 500 81.247 ms - 745
```

This is the app looks like:

screen shot 2015-05-07 at 11 17 28 am

Pomax:
I updated the package.json to include webmaker-download-locales as a dependency, so it should not try to use your globally installed version anymore (which is likely the culprit for the first error). Looking into the LESS issue now.

Pomax:
LESS errors were being caused by a [relative dir location bug in makerstrap](https://github.com/mozilla/makerstrap/issues/90). Last commit sidesteps that by loading makerstrap the not-prebuilt-way.

sedge:
r+, nice one.

Pomax:
next question is whether the #bramble branch can be lifted as easily.

Issue #504, opened 2015-05-12
Tech Debt: Fix Localization


JordanTheriault:
Both thimble and brackets have localization, but it is not enabled and they do not communicate with each other.

If the language is changed within Thimble, it should also change the language within Brackets.

humphd:
I sketched out a solution in https://github.com/humphd/brackets/issues/245

sedge:
Fixed in https://github.com/mozilla/thimble.webmaker.org/pull/516

PR #516, opened 2015-05-18
Issue504 locale

JordanTheriault:
Added functionality which upon language change in thimble adds the locale code to Backet's query string. Addresses #504.

### DIFF
--- a/routes/index.js
+++ b/routes/index.js
@@ -62,9 +62,16 @@ module.exports = function(utils, env, nunjucksEnv, appName) {
         }));
       }
 
+      //We add the localization code to the query params through a URL object
+      //and set search prop to nothing forcing query to be used during url.format()
+      var urlObj = url.parse(req.url, true);
+      urlObj.search = "";
+      urlObj.query["locale"] = req.localeInfo.lang;
+      var thimbleUrl = url.format(urlObj);
+
       // We forward query string params down to the editor iframe so that
       // it's easy to do things like enableExtensions/disableExtensions
-      var queryString = url.parse(req.url).search || '';
+      var queryString = url.parse(thimbleUrl).search || '';
 
       res.render('index.html', {
         appname: appName,


routes/index.js
```diff
@@ -62,6 +62,14 @@ module.exports = function(utils, env, nunjucksEnv, appName) {
         }));
       }
 
+      //We add the localization code to the query params through a URL object
+      //and set search prop to nothing forcing query to be used during url.format()
+      var urlObj, thimbleUrl;
```
humphd: Break this onto two separate lines for increased visibility:

```js
var urlObj;
var thimbleUrl;
```

routes/index.js
```diff
@@ -62,6 +62,14 @@ module.exports = function(utils, env, nunjucksEnv, appName) {
...
+      urlObj = url.parse(req.url, true);
+      urlObj.search = "";
+      urlObj.query["locale"] = req.localeInfo.lang;
+      thimbleUrl = url.format(urlObj);
```
humphd: Actually, just declare this here:

```js
var thimbleUrl = url.format(urlObj);
```

routes/index.js
```diff
@@ -62,6 +62,14 @@ module.exports = function(utils, env, nunjucksEnv, appName) {
...
+
       // We forward query string params down to the editor iframe so that
       // it's easy to do things like enableExtensions/disableExtensions
       var queryString = url.parse(req.url).search || '';
```
humphd: Should this not be `thimbleUrl` instead of `req.url`?

JordanTheriault: Yes, good catch. I must have accidentally removed the change when squashing.

routes/index.js
```diff
@@ -62,9 +62,17 @@ module.exports = function(utils, env, nunjucksEnv, appName) {
         }));
       }
 
+      //We add the localization code to the query params through a URL object
+      //and set search prop to nothing forcing query to be used during url.format()
+      var thimbleUrl;
```
humphd: Don't declare this here, do it all at once on line 71

routes/index.js
```diff
@@ -62,9 +62,17 @@ module.exports = function(utils, env, nunjucksEnv, appName) {
...
+      var urlObj = url.parse(req.url, true);
+      urlObj.search = "";
```
humphd: Why do you clear this?  Won't we lose things like `?enableExtensions=...` and the like?

humphd: OK, I understand this now, it's good.

humphd:
One nit, and this is r+.  Get @sedge to sign off on it, too.

sedge:
r+

PR #575, opened 2015-06-18
Client side implementation of persistence to publish.webmaker.org

gideonthomas:


public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -0,0 +1,77 @@
...
+      options.data = JSON.stringify({
+        path: path,
+        buffer: data
+      });
+
+      send();
+    });
+  }
+
+  function pushFileDelete() {}
```
humphd: Console log that this isn't implemented yet.

public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -0,0 +1,77 @@
...
+
+  function pushFileRename() {}
```
humphd: Same here

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,168 @@
+define([], function() {
+  "use strict";
+
+  var Path = Bramble.Filer.Path;
+  var FilerBuffer = Bramble.Filer.Buffer;
+  var self = {};
```
humphd: This isn't a great pattern.  Don't call this `self`, just have everything be local functions, and then do:

```js
return {
  load: load,
  ...
};
```

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,168 @@
...
+
+  // Taken from http://stackoverflow.com/questions/6965107/converting-between-strings-and-arraybuffers
+  function convertToBuffer(string) {
```
humphd: `convertToArrayBuffer` since we also have `Filer.Buffer`.

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,168 @@
...
+      bufView[i] = string.charCodeAt(i);
+    }
+    return buf;
+  }
+
+  self.load = function(project, defaultTemplate, callback) {
+    var fs = Bramble.getFileSystem();
+    var shell = new fs.Shell();
+    var projectFilesUrl = window.location.protocol + "//" + window.location.host + "/initializeProject";
+    var request = new XMLHttpRequest();
```
humphd: Why do you mix using jquery XHR and native?

humphd: It might be nice to use native XHR, so we can move this whole thing into a worker thread in the future.

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,168 @@
...
+              callback(err);
+              return;
+            }
+
+            completed++;
+            checkProjectLoaded();
+          });
+        }
+
+        fs.stat(parent, function(err) {
```
humphd: Don't bother with the `stat` call, just do `mkdirp` always, and if it's there already, don't fail.

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,168 @@
...
+        });
+      }
+
+      if(!length) {
+        // TODO: What should we do here? :P
+        callback(new Error("No files to load"));
+        return;
+      }
+
+      files.forEach(function(file) {
```
humphd: Will we have to worry about races at all, with two writes both trying to make the project dir or something?

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,168 @@
...
+      tags: [],
+      description: ""
+    };
+  };
+
+  self.generateDefaultFiles = function(defaultTemplate) {
+    return [
+      {
+        path: "/New Project/index.html",
+        buffer: convertToBuffer(defaultTemplate || "")
```
humphd: Why would we ever want an empty string for the HTML file?

public/friendlycode/js/fc/ui/bramble-project.js
```diff
@@ -0,0 +1,13 @@
...
+
+  var self = {};
+
+  self.updateMeta = function(project) {
+    // Update the project title in the userbar
+    // see views/userbar.html
+    $("#project-title").text(project.title || "New Project");
+  };
+
+  return self;
```
humphd:
```js
return {
  updateMeta: updateMeta
};
```

public/friendlycode/js/fc/ui/bramble-proxy.js
```diff
@@ -22,6 +22,7 @@ define(["backbone-events", "fc/prefs", "fc/bramble-ui-bridge"],
     var that = this;
     var lastLine = 0;
     var scrollInfo;
+    var _instance;
```
humphd: This seems unused.

public/friendlycode/js/fc/load-project-files.js
gideonthomas: It shouldn't once we add the condition that ignores the `EEXIST` error for `mkdirp`

public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -0,0 +1,105 @@
...
+      }
+
+      console.log("Successfully deleted ", path);
+    });
+    request.fail(function(jqXHR, status, err) {
+      console.error("Failed to send request to delete the file to the server with: ", err);
+    });
+  }
+
+  function pushFileRename() {}
```
sedge: `fn(url, csrf, fs, oldPath, newPath)`

public/friendlycode/js/friendlycode.js
```diff
@@ -77,9 +81,25 @@ define(function(require) {
...
+        ProjectFiles.load(makeDetails, defaultContent, callback);
+        return;
+      }
+
+      ProjectFiles.load(makeDetails, callback);
+    };
+
+    FileSystemSync.init(makeDetails && makeDetails.title, {
+      createOrUpdate: options.appUrl + "/updateProjectFile",
+      del: options.appUrl + "/deleteProjectFile"
```
sedge: `rename: options.appUrl + /newRoute`

app.js
```diff
@@ -206,6 +206,22 @@ app.get('/',
...
+        middleware.checkForAuth,
+        routes.getProject);
+
+app.get('/project/:projectId', routes.openProject);
+
+app.get('/projectExists/:projectName', routes.projectExists);
+
+app.get('/newProject/:projectName', routes.newProject);
+
+app['delete']('/deleteProject/:projectId', routes.deleteProject);
```
sedge: What's this about?

sedge: If you're concerned about the keyword don't be. `app.delete()` should work fine.

public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -0,0 +1,104 @@
...
+        "X-Csrf-Token": csrfToken
+      },
+      type: "PUT",
+      url: url
+    };
+
+    function send() {
+      var request = $.ajax(options);
+      request.done(function() {
+        if(request.readyState !== 4) {
```
sedge: You don't need this, since `.done` callbacks are executed when the parent promise resolves. In other words, this will only execute when the request is finished.

public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -0,0 +1,104 @@
...
+      });
+      request.fail(function(jqXHR, status, err) {
+        console.error("Failed to send request to persist the file to the server with: ", err);
+      });
+    }
+
+    fs.readFile(path, function(err, data) {
+      if(err) {
+        // TODO: handle errors
+        throw err;
```
humphd: Can we `console.error("[Bramble]...", err);` here instead, so that it looks like we've dealt with it on the console vs. just blowing up?  Also, how bad an error is this, since it would just mean that somehow we thought we needed to deal with this file, and actually we don't (i.e., it's not there now).  Maybe we can just `console.warn` and bail here.

public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -0,0 +1,104 @@
+define(["jquery"], function($) {
+  "use strict";
+
+  var FileSystemSync = {};
+
+  function pushFileChange(url, csrfToken, fs, path) {
```
humphd: Because `Bramble.getFileSystem()` is globally available, maybe you don't need to bother passing `fs` around all the time?

public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -0,0 +1,104 @@
...
+        "X-Csrf-Token": csrfToken
+      },
+      type: "PUT",
+      url: url,
+      data: JSON.stringify({
+        "path": path
+      })
+    });
+    request.done(function() {
+      if(request.readyState !== 4) {
```
sedge: Same as above

public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -0,0 +1,104 @@
...
+    function send() {
+      var request = $.ajax(options);
+      request.done(function() {
+        if(request.readyState !== 4) {
+          return;
+        }
+
+        if(request.status !== 201 && request.status !== 200) {
+          // TODO: handle error case here
+          console.error("Server did not persist file");
```
humphd: Prefix your console messages with `[Bramble]`.  Also, specify the `path` here, so we know which file it's talking about, and also the `request.status` you actually got:

```js
console.error("[Bramble] unable to persist `" + path + "` to server. Server responded with status `" + request.status + "`");
```

public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -0,0 +1,104 @@
...
+          return;
+        }
+
+        console.log("Successfully persisted ", path);
```
humphd: I understand why you want this for debugging, but let's not do this.  What is our plan around events on `bramble` for success/error cases with syncing file changes to the server?  We could do something like:

```js
bramble.trigger("syncFinished", path); // when it works
...
bramble.trigger("syncFailed", path, reason); // when it fails
```

And then clients can do whatever with that info.

app.js
gideonthomas: No coz jshint complains

public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -0,0 +1,104 @@
...
+    });
+    request.fail(function(jqXHR, status, err) {
+      console.error("Failed to send request to delete the file to the server with: ", status);
+    });
+  }
+
+  function pushFileRename() {}
+
+  FileSystemSync.init = function(projectName, persistanceUrls, csrfToken) {
+    if(!projectName) {
```
humphd: Why would this happen?  Can you add a comment.  Is this for anonymous user flow?

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,215 @@
+define(["jquery"], function($) {
+  "use strict";
+
+  var Path = Bramble.Filer.Path;
+  var FilerBuffer = Bramble.Filer.Buffer;
+
+  // Taken from http://stackoverflow.com/questions/6965107/converting-between-strings-and-arraybuffers
+  function convertToArrayBuffer(string) {
```
humphd: For follow-up, this is a great candidate function for moving to a worker, and making async.

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,215 @@
...
+    for(var i = 0, strLen = string.length; i < strLen; i++) {
+      bufView[i] = string.charCodeAt(i);
+    }
+    return buf;
+  }
+
+  function load(project, options, callback) {
+    var fs = Bramble.getFileSystem();
+    var shell = new fs.Shell();
+    var projectFilesUrl = window.location.protocol + "//" + window.location.host + "/initializeProject";
```
humphd: Can you not do protocol-less and just use `"//" + window.location.host + "/initializeProject";`?

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,215 @@
...
+    var request = new XMLHttpRequest();
```
humphd: I still think we should standardize on always using `$.ajax` or always using `XHR` native.  Why not just stick with `$.ajax` here, since you already have jquery?

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,215 @@
+define(["jquery"], function($) {
+  "use strict";
+
+  var Path = Bramble.Filer.Path;
```
sedge: Are these files being jshinted?

public/friendlycode/js/fc/load-project-files.js
sedge: Why aren't you using jquery here too?

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,215 @@
...
+    }
+    return buf;
+  }
+
+  function load(project, options, callback) {
+    var fs = Bramble.getFileSystem();
+    var shell = new fs.Shell();
+    var projectFilesUrl = window.location.protocol + "//" + window.location.host + "/initializeProject";
+    var request = new XMLHttpRequest();
+    var root = project && project.title;
```
humphd: I'd rename this `projectTitle` here, and not have it become `root` until further down, when it actually becomes a path.

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,215 @@
...
+    if(!root) {
+      callback(new Error("No project specified"));
+      return;
+    }
+
+    root = Path.join("/", root);
+
+    if(typeof options !== "function") {
+      defaultProject = true;
+      defaultPath = options.isNew ? "/" + project.title : "/New Project";
```
humphd: Prefer `Path.join` to concatenating dir separators.

sedge: How can this be `/New Project`? If `project.title` doesn't exist then it bails on line 29.

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,215 @@
...
+      project.dateCreated = project.dateCreated || (new Date()).toISOString();
+      project.dateUpdated = project.dateCreated;
```
humphd: Question: why do we not just deal with this on the server?  Why do you have clients pass you dates like this vs. just stamping it into the db when you touch those records on the server?

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,215 @@
...
+      files = generateDefaultFiles(options.defaultTemplate, defaultPath);
+      updateFs();
+      return;
+    }
+
+    callback = options;
+
+    function relative(path) {
```
humphd: For follow-up later, please look at my comments in https://github.com/filerjs/filer/issues/357

public/friendlycode/js/fc/load-project-files.js
sedge: This seems fishy. Isn't it the equivalent of:

```js
if (options.isNew) {
  defaultPath = "/" + project.title;
} else {
  defaultPath = "New Project"
}
```

public/friendlycode/js/fc/load-project-files.js
gideonthomas: So that we can account for the client's timezone as opposed to the server's timezone

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,215 @@
...
+
+    request.onreadystatechange = updateFs;
+    request.responseType = "json";
+    request.open("GET", projectFilesUrl, true);
+    request.send();
+  }
+
+  function generateDefaultProject(title) {
+    return {
+      title: title || "New Project",
```
humphd: It might be wise to move constants to another module, `Constants.js`, so we can easily modify them without hunting.

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,215 @@
...
+      tags: [],
+      description: ""
+    };
+  }
+
+  function generateDefaultFiles(defaultTemplate, projectPath) {
+    return [
+      {
```
humphd: When you only have a single item in the array, it's better to do:

```js
return [{
...
}];
```

public/friendlycode/js/fc/ui/bramble-project.js
```diff
@@ -0,0 +1,13 @@
+define(["jquery.min"], function($) {
+  "use strict";
+
+  function updateMeta(project) {
+    // Update the project title in the userbar
+    // see views/userbar.html
+    $("#project-title").text(project.title || "New Project");
```
humphd: Same deal here, better to have `Constants.js` and reuse it everywhere.  Down the road we'll also need to ponder how we localize this.

public/friendlycode/js/fc/ui/bramble-proxy.js
```diff
@@ -120,32 +122,16 @@ define(["backbone-events", "fc/prefs", "fc/bramble-ui-bridge"],
...
-                   "  <body>\n"                   +
-                   "    <p>Hello World</p>\n"     +
-                   "  </body>\n"                  +
-                   "</html>";
-
-        fs.writeFile("/project/index.html", html, function(err) {
-          if (err) {
-            throw err;
-          }
+        console.log(Bramble.Filer.Path.normalize(config.root));
```
humphd: Remember to remove this.

public/friendlycode/js/friendlycode.js
```diff
@@ -48,6 +51,7 @@ define(function(require) {
       window: window,
       currentPage: pageToLoad
     });
+    ProjectUI.updateMeta(makeDetails);
```
humphd: We should be careful using the word `make` still, since we've dropped that idea in the new system.  We only have projects now.

public/friendlycode/js/friendlycode.js
```diff
@@ -77,9 +81,36 @@ define(function(require) {
...
+        return;
+      }
+
+      if(makeDetails.isNew) {
+        makeDetails = ProjectFiles.generateDefaultProject(makeDetails.title);
+        ProjectFiles.load(makeDetails, {
+          isNew: true,
+          defaultTemplate: defaultContent,
+          csrfToken: $("meta[name='csrf-token']").attr("content"),
+          persistenceURL: options.appUrl + "/updateProjectFile"
```
humphd: `Path.join`

public/friendlycode/js/friendlycode.js
humphd: Also, `Path.join`

public/friendlycode/js/friendlycode.js
```diff
@@ -77,9 +81,36 @@ define(function(require) {
...
+
+    FileSystemSync.init(makeDetails && makeDetails.title, {
+      createOrUpdate: options.appUrl + "/updateProjectFile",
+      del: options.appUrl + "/deleteProjectFile"
+    }, $("meta[name='csrf-token']").attr("content"));
+
     if (!pageManager.currentPage()) {
       setTimeout(function() {
-        editor.panes.codeMirror.init(defaultContent);
+        editor.panes.codeMirror.init(defaultContent, initFs);
```
humphd: I realize you're hacking this into the existing stuff, but for sure `editor.panes.codeMirror` is going away.  Think about how we'll do this with the new stuff.

public/scripts/open-project.js
```diff
@@ -0,0 +1,115 @@
...
+
+    function generateUrl(path, date) {
+      var url = window.location;
+      var qs = url.search;
+      if(qs.length < 1) {
+        qs = "?";
+      }
+      qs += "now=" + encodeURIComponent(date);
+
+      return url.protocol + "//" + url.host + "/newProject/" + encodeURIComponent(path) + qs;
```
humphd: You're doing this kind of thing in a few places, I wonder if you should move this to a `utils` module and reuse?  Also, see comments above about maybe dropping `protocol`.

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,215 @@
...
+          },
+          type: "PUT",
+          url: options.persistenceURL,
+          data: JSON.stringify({
+            path: path,
+            buffer: data
+          })
+        });
+        request.done(function() {
+          if(request.readyState !== 4) {
```
sedge: nuke this check

public/scripts/open-project.js
```diff
@@ -0,0 +1,115 @@
+(function() {
+  require(["jquery"], function($) {
+    var projects = document.querySelectorAll("tr.bramble-user-project");
+
+    function generateUrl(path, date) {
+      var url = window.location;
+      var qs = url.search;
+      if(qs.length < 1) {
+        qs = "?";
+      }
```
humphd: if `qs.length > 1`, don't you need to add a `&` in order to concat the stuff on the next line to the existing bits?

public/scripts/open-project.js
```diff
@@ -0,0 +1,115 @@
...
+      qs += "now=" + encodeURIComponent(date);
+
+      return url.protocol + "//" + url.host + "/newProject/" + encodeURIComponent(path) + qs;
+    }
+
+    $("#project-submit").on("click", function(e) {
+      e.preventDefault();
+
+      var projectName = document.getElementById("project-name");
```
humphd: This feels sloppy.  We're mixing state info with UI presentation.  I realize we don't want to over-engineer how we do this, but should we make a module that holds this state, and have the UI use it, and you use it here?

public/friendlycode/js/fc/load-project-files.js
```diff
@@ -0,0 +1,215 @@
...
+      if(!length) {
+        // TODO: What should we do here? :P
+        callback(new Error("No files to load"));
+        return;
+      }
+
+      files.forEach(function(file) {
+        // TODO: Make this configurable
+        if(!filePathToOpen || Path.extname(filePathToOpen) !== ".html") {
+          filePathToOpen = relative(file.path);
```
sedge: So the last .html file it finds becomes the one we open? Shouldn't this just be hardcoded as `index.html`?

public/scripts/open-project.js
```diff
@@ -0,0 +1,115 @@
...
+      if (value.length < 1) {
+        return;
+      }
+
+      var request = $.ajax({
+        type: "GET",
+        url: "/projectExists/" + encodeURIComponent(value),
+        statusCode: {
+          404: function() {
+            if (request.status === 404) {
```
humphd: Why do you do this check?  Isn't the point of doing `statusCode: { 404: function() {...} }` so that you know it's that code?

public/stylesheets/projects-style.css
```diff
@@ -0,0 +1,128 @@
+body {
+	font-family: 'Open Sans', sans-serif;
+	font-size: 1em;
+}
+
+#container {
+	width: 80%;
+	margin: 0px auto;
```
humphd: remove `px` on 0 values.

public/stylesheets/projects-style.css
```diff
@@ -0,0 +1,128 @@
...
+
+table#project-list {
+	width: 100%;
+	border: 0;
+	padding: 0;
+	border-collapse: collapse;
+	cursor: pointer;
+	-webkit-user-select: none; /* Chrome/Safari */
+	-moz-user-select: none; /* Firefox */
+	-ms-user-select: none; /* IE10+ */
```
humphd: Add unprefixed at the end.

public/stylesheets/projects-style.css
```diff
@@ -0,0 +1,128 @@
...
+}
+tr.project:hover span.project-information {
+	color: #A1A1A1;
+}
+tr.project:hover > td.side-fill {
+	border-bottom: 2px solid #EFEFEF;
+}
+
+tr.project > td:first-child {
+	padding-left: 0px;
```
humphd: remove px

public/stylesheets/projects-style.css
```diff
@@ -0,0 +1,128 @@
...
+}
+tr.project td:last-child {
+	padding-right: 0px;
```
humphd: remove px

public/stylesheets/projects-style.css
```diff
@@ -0,0 +1,128 @@
...
+td.side-fill {
+	border-bottom: 0;
+	width: 10px;
+}
+td.button-fill {
+	width: 40px;
+}
+
+td.project-title {
+	padding-left: 0px;
```
humphd: remove px

routes/index.js
```diff
@@ -4,12 +4,19 @@
...
 var langmap = i18n.getAllLocaleCodes();
-var url = require("url");
-var env = require('../lib/environment');
-var Cryptr = require("cryptr");
 var request = require("request");
-
-var cryptr = new Cryptr(env.get("SESSION_SECRET"));
+var config = require("./config");
+
+// Bramble routes
```
humphd: This is a lot better, nice work.

views/layout.html
```diff
@@ -17,6 +17,7 @@
     <link rel="stylesheet" href="{{ HTTP_STATIC_URL }}friendlycode/css/details-form.css">
     <link rel="stylesheet" href="https://ajax.googleapis.com/ajax/libs/jqueryui/1.10.3/themes/smoothness/jquery-ui.min.css">
     <link rel="stylesheet" href="{{ HTTP_STATIC_URL }}stylesheets/userbar.css">
+    <link rel="stylesheet" href="{{ HTTP_STATIC_URL }}stylesheets/projects-style.css">
```
humphd: For follow-up, we should merge all these CSS files into a single file.

views/projects.html
```diff
@@ -0,0 +1,50 @@
...
+
+    <div class="project-button" id="project-0">Create New</div>
+
+    <div id="modal_wrapper">
+      <div id="modal_window">
+
+      <div style="text-align: right;"><a id="modal_close" href="#">Close <b>X</b></a></div>
+
+      <p><label>Project Name<br>
+      <input id="project-name" type="text" autofocus required size="48" name="name" value=""></label></p>
```
humphd: size=48?  Where did this magic number come from?

humphd:
One last comment, I see a lot of ```//TODO : handle error```.  You should consider how you'll deal with this, if at all, and remove these if you have no plan to deal with them.  Probably logging them is enough in many cases.

routes/check-if-project-exists.js
```diff
@@ -0,0 +1,81 @@
+var request = require("request");
+
+module.exports = function(config) {
+  return function(req, res) {
+    var publishURL = config.publishURL;
+    var token = config.cryptr.decrypt(req.session.token);
```
sedge: As long as the user middleware runs on these routes, you don't need to decrypt the token:

https://github.com/mozilla/thimble.webmaker.org/blob/bramble/lib/middleware.js#L530-L533

routes/check-if-project-exists.js
```diff
@@ -0,0 +1,81 @@
...
+
+    if(!req.session.user) {
```
sedge: This should probably be refactored into a piece of middleware and attached to the routes you want it in.

routes/check-if-project-exists.js
```diff
@@ -0,0 +1,81 @@
...
+      }
+
+      if(response.statusCode !== 200 &&  response.statusCode !== 201) {
+        console.error('Error retrieving user: ', response.body);
+        res.send(500);
+        // deal with failure
+        return;
+      }
+
+      var publishUser = req.session.publishUser = body;
```
sedge: I'd attach this to `req.session.user.publishId` instead.

routes/create-or-update-file.js
```diff
@@ -0,0 +1,65 @@
+var request = require("request");
+
+module.exports = function(config) {
+  return function(req, res) {
+    if(!config.authorized(req, res)) {
```
sedge: This is probably an anti-pattern in express. I'd declare this as middleware that runs before this route handler.

sedge: An elegant way to incorporate this change would be for the `module.exports` function to return an array:

```js
module.exports = function(config) {
  return [
    config.authorized,
    function(req, res) { 
      // ...
    }
  ];
};
```

routes/create-or-update-file.js
```diff
@@ -0,0 +1,65 @@
...
+      res.send(400);
+      return;
+    }
+
+    var fileReceived = {
+      path: req.body.path,
+      buffer: req.body.buffer.data,
+      project_id: req.session.project.meta.id
+    };
+    var existingFile = req.session.project.files[fileReceived.path];
```
sedge: We have to be mindful of what we're doing here: Every extra file added means a larger cookie for the client.  I'm not sure we can avoid it at the moment though...

routes/delete-file.js
```diff
@@ -0,0 +1,49 @@
+var request = require("request");
+
+module.exports = function(config) {
+  return function(req, res) {
+    if(!config.authorized(req, res)) {
```
sedge: See comments above

routes/delete-project.js
```diff
@@ -0,0 +1,100 @@
...
+            console.error("Error deleting project file: ", response.body);
+            callback(response.statusCode);
+            return;
+          }
+
+          callback();
+        });
+      }
+
+      async.eachSeries(files, deleteFile, function(err) {
```
sedge: This is unnecessary. Deleting the project should be enough.

routes/get-current-project.js
```diff
@@ -0,0 +1,47 @@
...
+        // TODO: handle error
+        console.error('Error retrieving user\'s project files: ', response.body);
+        res.send(404);
+        return;
+      }
+
+      var files = JSON.parse(body);
+      req.session.project.files = {};
+      files.forEach(function(file) {
+        var fileMeta = JSON.parse(JSON.stringify(file));
```
sedge: This makes me think we need a way to get just the representation without the data. Is it worth it to add that functionality? It wouldn't be too hard.

routes/home.js
```diff
@@ -0,0 +1,63 @@
+var url = require("url");
+
+module.exports = function(config) {
+  return function(req, res) {
+    var makedetails = "{}";
+    var project = req.session.project && req.session.project.meta;
+      console.log(req.session.project.isNew);
```
sedge: nuke this

routes/home.js
```diff
@@ -0,0 +1,63 @@
+var url = require("url");
+
+module.exports = function(config) {
+  return function(req, res) {
+    var makedetails = "{}";
```
sedge: `var makedetails = {}`

routes/index.js
sedge: +1, the world makes a little more sense now.

routes/set-current-project.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    var projectId = req.params.projectId;
+    if(!projectId) {
+      // TODO: handle error
+      console.error('No project ID specified');
+      return;
+    }
+
+    // TODO: UI implementation (progress bar/spinner etc.)
```
sedge: This todo doesn't belong on the server.

views/projects.html
sedge: @JordanTheriault Can you comment?

public/friendlycode/js/friendlycode.js
gideonthomas: I can't use `Path.join` because it normalizes the result...so I end up with something like:
`Path.join("http://localhost:3500/", "updateProjectFile") = "http://localhost:3500/localhost:3500/updateProjectFile"`

public/friendlycode/js/fc/filesystem-sync.js
gideonthomas: It saves me a line of code in each function :)

public/friendlycode/js/fc/filesystem-sync.js
gideonthomas: I have an object called `fsync` that I think should be emitting these events. But I think we should wait until @sedge's refactor so that we can use node's `EventEmitter` to make our lives easier.

public/friendlycode/js/friendlycode.js
gideonthomas: I think we should deal with that once we start gutting out thimble!

public/scripts/open-project.js
gideonthomas: This is actually just getting whatever the user types in the input box for the name of a new project

public/friendlycode/js/fc/load-project-files.js
gideonthomas: For now...no! Some of the seeds in `publish.webmaker.org` do not have `index.html` files. So as a temporary solution, we're selecting the last html file that was loaded.

views/projects.html
sedge: This was code borrowed along with the modal javascript.  As far as I understood, it wasn't meant to be final.

routes/check-if-project-exists.js
gideonthomas: I think I'm going to leave this for now until we decide where we store user prefs

routes/create-or-update-file.js
gideonthomas: Yeah that's true! I'm not a huge fan of storing file lists on the server, but for now I don't think there is a clean way around it. Luckily, we're not storing the buffers.

routes/get-current-project.js
gideonthomas: Either way would be fine!

routes/home.js
gideonthomas: I can't do that because when you `decodeURIComponent`, `{ }` will return `[object object]` whereas `"{ }"` will return `"{ }"` which can be JSON parsed.

public/friendlycode/js/friendlycode.js
gideonthomas: I can't do `Path.join` because it also normalizes the result. So, `Path.join("http://localhost:2015/", "updateProjectFile")` will result in `http://localhost:2015/localhost:2015/updateProjectFile`

sedge:
r+.

Nit: Remove TODOS unless there are bugs attached to them.

PR #578, opened 2015-06-19
Fixed several UI issues in safari, altered theme switching animation, fixed filetree arrow

JordanTheriault:


public/friendlycode/css/editor.less
```diff
@@ -265,37 +264,32 @@
...
     margin-right: 10px;
     img {
-      position: relative;
-      left: -4px;
-    }
-    img#moon-green {
-      display: none;
+      position: absolute;
+      top: 100px;
+      left:165px;
```
humphd: space after `:`

views/friendlycode/templates/nav-options.html
```diff
@@ -1,7 +1,7 @@
 <div class="editor-nav">
 
   <!-- FILE TREE -->
-  <div class="filetree-pane-nav">
+  <div class="filetree-pane-nav" style="display:none;" class="hide">
```
humphd: Why do you need both the inline style attrib and the class?

humphd:
A few small things, r+ on code.

views/friendlycode/templates/nav-options.html
JordanTheriault: Not necessary, but added for consistency.
This element needs an inline display:none in order to hide it on load (the class adds the CSS too late, and creates a bad load visual).
 Everything else uses class="hide" except this element.

JordanTheriault:
I'm going to wait until after the demo to merge.

PR #595, opened 2015-06-23
Ignore css files generated during node app

JordanTheriault:
Updated to ignore:
public/friendlycode/css/editor.css
public/stylesheets/userbar.css


gideonthomas:
I don't think we actually need this any more with the upgrade to node 12

JordanTheriault:
Confirmed, just checked. Closing pr.

PR #624, opened 2015-06-29
Fix #600 - Update express to 4.x

### LINKED ISSUE #600: Update express
We are using an ancient version of express that has been deprecated. Initially, this was not a big problem but we recently ran into an issue: 
The built-in body parser in express (which no longer exists in later versions) enforces a limit on HTTP message body sizes which has no easy way of being overridden. Hence, uploading images > 100kb is practically impossible. Updating express will allow us to specify the limit on message body sizes.

humphd:


humphd:
r+ from @gideonthomas on irc, who has tested this locally.

PR #628, opened 2015-06-29
Get Publishing working

gideonthomas:


public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -23,6 +23,11 @@ define(["jquery"], function($) {
       request.done(function() {
         if(request.status !== 201 && request.status !== 200) {
           console.error("[Bramble] Server did not persist ", path, ". Server responded with status ", request.status);
+          return;
+        }
+
+        if(context._callback) {
+          context._callback();
```
sedge: `context` is the scope of `pushFileChange` right? Where might `_callback` be set?

public/friendlycode/js/fc/publish.js
```diff
@@ -0,0 +1,172 @@
...
+        "Accept": "application/json"
+      },
+      type: "PUT",
+      url: url + "/unpublish",
+      data: JSON.stringify(project)
+    });
+    request.done(function() {
+      if(request.status !== 200) {
+        console.error("[Bramble] Server was unable to unpublish project, responded with status ", request.status);
+        unpublishBtn.off("click", unpublish).on("click", unpublish);
```
sedge: You've already removed the listener on line 37 haven't you?

public/friendlycode/js/fc/publish.js
```diff
@@ -0,0 +1,172 @@
...
+
+      $("#publish-link > a")
+        .attr("href", "")
+        .text("");
+      $("#publish-changes").addClass("hide");
+      $("#publish-live").addClass("hide");
+    });
+    request.fail(function(jqXHR, status, err) {
+      console.error("[Bramble] Failed to send request to unpublish project to the server with: ", err);
+      unpublishBtn.off("click", unpublish).on("click", unpublish);
```
sedge: Ditto

public/friendlycode/js/fc/publish.js
```diff
@@ -0,0 +1,172 @@
...
+        .text(project.publishUrl);
+      $("#publish-changes").addClass("hide");
+      $("#publish-live").removeClass("hide");
+      unpublishBtn.on("click", unpublish);
+    }
+
+    if(fileSync) {
+      fileSync.setCallback(function() {
+        $("#publish-changes").removeClass("hide");
+        updatePublishBtn.off("click", publish).on("click", publish);
```
sedge: Nit: I know why you're doing it, but this on/off stuff is kinda gross. Is there a way to check if the event is registered? It would make this stuff easier to read.

public/friendlycode/js/fc/publish.js
gideonthomas: yep good catch! I'm going to probably disable the publish button as well so that we don't run into timing issues if you click both buttons at the same time

public/friendlycode/js/fc/publish.js
gideonthomas: Tried to search for that haha but couldn't find anything! After doing some research, I found that this was the cleanest way to do it

routes/publish.js
```diff
@@ -0,0 +1,71 @@
+var request = require("request");
+
+function isEmpty(val) {
+  return val === null || val === undefined;
+}
+
+module.exports = function(config) {
+  return function(req, res) {
+    var project = JSON.parse(JSON.stringify(req.session.project.meta));
```
sedge: Try/catch

gideonthomas: I don't think that's necessary if we are the ones controlling what the object looks like

routes/publish.js
```diff
@@ -0,0 +1,71 @@
...
+    request({
+      method: "PUT",
+      uri: config.publishURL + "/projects/" + req.session.project.meta.id,
+      headers: {
+        "Authorization": "token " + req.user.token
+      },
+      body: project,
+      json: true
+    }, function(err, response, body) {
+      if(err) {
```
sedge: Mayb

routes/publish.js
```diff
@@ -0,0 +1,71 @@
...
+        res.status(500).send({error: err});
```
sedge: I'd log the error and send back a generic 500 message.

routes/publish.js
```diff
@@ -0,0 +1,71 @@
...
+        return;
+      }
+
+      if(response.statusCode !== 201) {
```
sedge: I'd also combine these two cases under one response code, unless there's a reason you need the publish server's response back.

routes/publish.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    project.description = req.body.description;
+    // Uncomment the line below once https://github.com/mozilla/publish.webmaker.org/issues/98 is done
+    // project.public = req.body.public;
+    project.date_updated = req.body.dateUpdated;
+
+    request({
+      method: "PUT",
+      uri: config.publishURL + "/projects/" + req.session.project.meta.id,
```
sedge: So every `publish` call from thimble includes the entire project? Do we really need to do this? I feel like the front end should prevent a publish action if the state of the project is dirty.

routes/unpublish.js
```diff
@@ -0,0 +1,70 @@
...
+module.exports = function(config) {
+  return function(req, res) {
+    var project = JSON.parse(JSON.stringify(req.session.project.meta));
+    delete project.id;
+    delete project.publish_url;
+
+    if(!req.body || isEmpty(req.body.description) || isEmpty(req.body.dateUpdated) || isEmpty(req.body.public)) {
+      res.status(400).send({ error:  "Request body missing data" });
+      return;
+    }
```
sedge: Nit: Everything up to this line could be a separate piece of middleware.

routes/unpublish.js
```diff
@@ -0,0 +1,70 @@
...
+    }, function(err, response, body) {
+      if(err) {
+        res.status(500).send({error: err});
+        return;
+      }
+
+      if(response.statusCode !== 201) {
+        res.status(response.statusCode).send({error: response.body});
+        return;
+      }
```
sedge: Correction - up to this line.

If we choose to keep the "update the project" step that is.

Otherwise, up to line 18.

sedge:
This is pretty good. Some nits, and a major question about the "update things" step in each process.

routes/publish.js
gideonthomas: It would help if the client knew whether it was a 404 or not

routes/publish.js
gideonthomas: This is done primarily because of how the UX is designed. The dialog that shows the "publish" button also has fields for description and "isPublic" which need to be updated as well...since this update is triggered by a single action (one mouse click on "Publish"), we have to do it on the server

sedge: Gotcha. Makes sense.

public/friendlycode/js/fc/publisher.js
```diff
@@ -0,0 +1,203 @@
...
+  function unpublishedChangesPrompt() {
+    var interface = this.interface;
+    var publish = this.handlers.publish;
+    interface.published.changed.removeClass("hide");
+    interface.buttons.update
+      .off("click", publish)
+      .on("click", publish);
+  }
+
+  function Publisher(editor, options) {
```
sedge: why not `function Publisher(bramble, options) {` ?

public/friendlycode/js/fc/publisher.js
```diff
@@ -0,0 +1,203 @@
...
+      url: host + route,
+      data: JSON.stringify({
+        description: publisher.interface.description.val() || " ",
+        public: publisher.public,
+        dateUpdated: (new Date()).toISOString()
+      })
+    });
+  };
+
+  Publisher.prototype.update = function(publishUrl, allowUnpublish) {
```
sedge: `updateDialog`

public/friendlycode/js/fc/publisher.js
```diff
@@ -0,0 +1,203 @@
...
+        changed: $("#publish-changes"),
+        container: $("#publish-live")
+      }
+    };
+  }
+
+  Publisher.prototype.init = function(project) {
+    var publisher = this;
+    var interface = publisher.interface;
+    publisher.public = true;
```
sedge: This seems to manage the project's state as well as the state of publishing itself. Are you sure we want to mix those? It makes this somewhat confusing.

public/friendlycode/js/fc/publisher.js
```diff
@@ -0,0 +1,203 @@
...
+  };
+
+  Publisher.prototype.publish = function() {
+    var publisher = this;
+    var interface = publisher.interface;
+
+    function setState(done) {
+      var buttons = interface.buttons;
+      var handlers = publisher.handlers;
+      var toggle = done ? "on" : "off";
```
sedge: This is a clever way of cleaning up the event listener stuff.

public/friendlycode/js/fc/publisher.js
```diff
@@ -0,0 +1,203 @@
...
+      contentType: "application/json",
+      headers: {
+        "X-Csrf-Token": publisher.csrfToken,
+        "Accept": "application/json"
+      },
+      type: "PUT",
+      url: host + route,
+      data: JSON.stringify({
+        description: publisher.interface.description.val() || " ",
+        public: publisher.public,
```
sedge: This might be clearer if the value was stored/toggled in the html element, and pulled from that element here.

public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -1,5 +1,6 @@
 define(["jquery"], function($) {
   var FileSystemSync = {};
+  var noop = function() {};
```
sedge: I can't figure out where these callbacks are set/unset.

sedge:
Awesome stuff. Can't wait to see this land :+1:

PR #638, opened 2015-06-30
Stay Calm template for user testing, minor fix to node error handling

humphd:
I've also slipped in a minor fix to the error handling middleware we use, which was wrong before.

This temporarily overrides how we do the default project content so @flukeout can do user testing, see https://github.com/mozilla/thimble.webmaker.org/issues/558#issuecomment-117259560

### DIFF
--- a/lib/errorhandling.js
+++ b/lib/errorhandling.js
@@ -14,7 +14,8 @@ function renderError( err, res ) {
   });
 }
 
-module.exports.errorHandler = function(err, req, res) {
+module.exports.errorHandler = function(err, req, res, next) {
+  //jshint unused:vars
   if (typeof err === "string") {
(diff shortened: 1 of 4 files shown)


lib/errorhandling.js
```diff
@@ -14,7 +14,7 @@ function renderError( err, res ) {
   });
 }
 
-module.exports.errorHandler = function(err, req, res) {
+module.exports.errorHandler = function(err, req, res, next) {
```
gideonthomas: why do we pass in next if we're not calling it anywhere?

humphd: It needs 4-args vs. 3 in order for express to treat it as error handling middleware, see http://expressjs.com/guide/error-handling.html

gideonthomas:
r=me with travis

PR #640, opened 2015-06-30
Show last opened project instead of project list on page refresh

gideonthomas:


routes/delete-project.js
```diff
@@ -27,6 +27,10 @@ module.exports = function(config) {
         return;
       }
 
+      if(req.session.project && req.session.project.meta.id == projectId) {
```
humphd: Do you think it's necessary to also add a check for `req.session.project.meta` before you try to get `req.session.project.meta.id`?

humphd:
One question, r=me otherwise.

routes/delete-project.js
gideonthomas: I don't think so! We have control over that code, so if the `project` exists (which it may or may not depending on whether or not the user has never opened a project in thimble) we know that it will contain all the necessary properties!

PR #654, opened 2015-07-02
Fix #652 - add crown.svg and thimble.svg to default project content

### LINKED ISSUE #652: Update temporary starting project
Please update to include the files in
https://dl.dropboxusercontent.com/u/109902/bramble/stay-calm.zip

Changes
* Crown image file now included in project, and HTML references it
* Thimble logo image file also included in project to facilitate testing task of image source replacement

humphd:


humphd:
One interesting thing this reveals is that since an SVG is really an XML doc, Brackets is happy to let you open it:

screenshot svg

I wonder if we should have it instead show you the image in the Image Viewer, which looks like this (I made another screenshot as a PNG and dragged it in to show what it looks like):

screenshot 2015-07-02 17 01 48

flukeout:
@humphd I've changed the images to be .png and also updated the index.html file to reference the correct starting image (crown.png), available here...
https://dl.dropboxusercontent.com/u/109902/bramble/stay-calm.zip

humphd:
We're going to go with SVG files for now, and I've fixed the problem above with showing XML vs. an image preview in https://github.com/humphd/brackets/pull/376 (not reviewed/landed yet) on the Brackets side.  I'm going to land this as is.

PR #657, opened 2015-07-03
Fix #650 - Make the entire editor theme toggle element clickable

### LINKED ISSUE #650: Make the entire editor theme toggle element clickable
Right now you can only toggle a theme if you click the part of the toggle that is not currently selected. Users had trouble using the toggle because of this.

This guy...
image

humphd:
This makes it so that clicking either theme option toggles the theme.  I've also updated the CSS pointer, which I remember reading you wanted somewhere (can't find bug).

### DIFF
--- a/public/friendlycode/css/editor.less
+++ b/public/friendlycode/css/editor.less
@@ -280,6 +280,7 @@
   padding: 7px 6px 5px 14px;
   width: 45px;
   .flex-container;
+  cursor: pointer;
 
   div#theme-dark {
     z-index: 11;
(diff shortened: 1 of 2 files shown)


JordanTheriault:
Clicking in dead centre of the toggle where the two circles do not intercept is a dead spot, but using #editor-pane-nav-toggle-theme does not work in this context.

r+, this button is on my list to redo and this patch works well for now.

humphd:
After landing @JordanTheriault's recent fix, I rebased and landed this.

PR #694, opened 2015-07-13
Fix #681 - Save all files before publishing anything

### LINKED ISSUE #681: Publishing doesn't publish the same version of the project as in the editor & preview window
Steps to reproduce
* Start a new project
* Make a small text change to the Carry On poster
* Immediately hit Publish
* Follow the link
* The changes you made are not reflected in the published version

Why? Probably because we don't auto-save except every 30 seconds and when you change files in the editor. 

Solution: Fire the auto-save call right before publishing to make sure we get the latest changes.

gideonthomas:


public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -7,6 +7,12 @@ define(["jquery"], function($) {
     }
   }
 
+  function trigger(callbacks, args) {
```
humphd: `triggerCallbacks`

public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -19,29 +25,30 @@ define(["jquery"], function($) {
...
       request.always(function() {
         context.queueLength--;
         hideFileState(context.queueLength);
+        trigger(context._callbacks.afterEach, [error, path]);
       });
     }
 
     fs.readFile(path, function(err, data) {
       if(err) {
+        context.queueLength--;
```
humphd: Why only in the error case?

gideonthomas: it happens in the success case as well, just not immediately!

public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -103,22 +114,124 @@ define(["jquery"], function($) {
...
+      if(fsync.queueLength > 0) {
+        return;
+      }
+
+      resetAfterEach(maybeContinue);
+      bramble.off("fileChange", fsync.handlers.change);
+      bramble.on("fileChange", cacheFilesSaved);
+
+      bramble.saveAll(function() {
+        bramble.off("fileChange", cacheFilesSaved);
```
humphd: Is there a risk of losing track of file saves during the time that the `cacheFilesSaved` callback is off *and* the `fsync.handlers.change` callback is also not wired up yet (i.e., between now and when `finish()` gets called?  Do we risk any loss of info in that period if a save occurs?

public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -89,6 +96,10 @@ define(["jquery"], function($) {
 
   function FSync() {
     this.queueLength = 0;
+    this._callbacks = {
+      beforeEach: [],
```
humphd: Document what these are before and after.  It's a bit unclear.

public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -103,22 +114,124 @@ define(["jquery"], function($) {
...
+        return;
+      }
+
+      fsync.afterEach(maybeFinish);
+      filesSaved.forEach(function(path) {
+        fsync.handlers.change(path);
+      });
+    }
+
+    function cacheFilesSaved(path) {
```
humphd: I wonder if this name should indicate that you're caching a file *path* and not the file itself, so there's no confusion.

public/friendlycode/js/fc/filesystem-sync.js
gideonthomas: Well, one (and right now the only) scenario that I can imagine is if you click on "Publish", and while it is still publishing, you make changes in the editor (which will be auto-saved locally) and then once it's published, you do not make any more changes. Instead, you shutdown your browser, in which case it will be stored in IndexedDB but not persisted and will probably be overwritten (with whats on the server) when you open up your project again. **BUT** this is a very unlikely edge-case that will happen only if you time it correctly (stop changing anything as soon as publishing has ended) and want to unintuitively focus out of the Publish dialog box while publishing is occuring.

tldr; I don't think we should worry about this

public/friendlycode/js/fc/filesystem-sync.js
```diff
@@ -103,22 +114,124 @@ define(["jquery"], function($) {
...
+      fsync.afterEach(maybeContinue);
+    } else {
+      maybeContinue();
+    }
+  };
+
+  // Add a callback to execute before every sync
+  // Each callback receives a `path` as an argument
+  FSync.prototype.beforeEach = function(callback) {
+    if(typeof callback !== "function") {
```
humphd: I don't think this check is necessary; same is true below for `afterEach`.  I might change the names, though: `FSync.prototype.add{before|after}EachCallback`

public/friendlycode/js/fc/publisher.js
```diff
@@ -63,6 +63,7 @@ define(function(require) {
   Publisher.prototype.init = function(project) {
     var publisher = this;
     var interface = publisher.interface;
```
humphd: `interface` is a reserved word in JS.

humphd:
I've done a round on the code, it's looking good minus my comments.  @sedge, since you know publish better than me, can you try this too?

sedge:
Code looks good - testing now.

sedge:
@gideonthomas walked me through it. r+

PR #703, opened 2015-07-16
Use user id and project name as part of file paths for indexeddb storage

gideonthomas:


gideonthomas:
This allows us to maintain different file paths for storage on bramble and in the publish.wm.org database. Thimble server acts as the moderator for this. Every incoming file path from publish-->Thimble server will be the file path as is. On Thimble server, the file path will be stored (in the session) based on whatever publish sends it. Every outgoing path from Thimble server-->Thimble client will be the stored path in Thimble server **prepended by `/:userID/projects/:projectID`** so that bramble can use that path to store all the project files

gideonthomas:
This patch depends on https://github.com/mozilla/publish.webmaker.org/pull/115 landing

public/scripts/editor/js/fc/load-project-files.js
```diff
@@ -22,56 +22,26 @@ define(["jquery", "constants", "text!fc/stay-calm/index.html", "text!fc/stay-cal
     var projectFilesUrl = "//" + window.location.host + "/initializeProject";
     var request;
     var projectTitle = project && project.title;
+    var root = project && project.root;
     var defaultProject = false;
-    var defaultPath;
     var files;
 
-    if(!projectTitle) {
+    if(!projectTitle || !root) {
```
sedge: Is it possible to have one but not the other?

public/scripts/editor/js/fc/load-project-files.js
```diff
@@ -187,9 +157,11 @@ define(["jquery", "constants", "text!fc/stay-calm/index.html", "text!fc/stay-cal
     });
   }
 
-  function generateDefaultProject(title) {
+  function generateDefaultProject(title, root) {
+    title = title || Constants.ANON_PROJECT_NAME;
```
sedge: Nit: Why this change?

public/scripts/editor/js/fc/load-project-files.js
```diff
@@ -187,9 +157,11 @@ define(["jquery", "constants", "text!fc/stay-calm/index.html", "text!fc/stay-cal
...
     return {
-      title: title || Constants.ANON_PROJECT_NAME,
+      root: root || Path.join("/", title),
```
sedge: It's doubly odd since you only do it for the one parameter.

public/scripts/editor/js/fc/load-project-files.js
gideonthomas: realistically, this condition should never eval to true because that would be very bad, I wonder if there is even a point checking for this

public/scripts/editor/js/fc/load-project-files.js
gideonthomas: I don't, next line uses it as well

public/scripts/editor/js/fc/load-project-files.js
gideonthomas: I use it in two places below

sedge:
r+

PR #709, opened 2015-07-17
Rename + Default project name + Project file loading fixes

gideonthomas:
This patch does a couple of things:
* Instead of prompting the user to enter a **unique** project name when creating a new project, it directly takes the user to the editor when "New Project" is clicked and uses the title "Untitled Project"
* Allows a user to rename an existing or new project and saves this change to publish.webmaker.org if the user is authenticated
* Fixes a couple of loading bugs in terms of when files were loaded/persisted for different cases such as an unauthenticated user, an authenticated user creating a new project and an authenticated user opening an existing project


routes/new-project.js
```diff
@@ -5,21 +5,16 @@ var utils = require("./utils");
 module.exports = function(config) {
   return function(req, res) {
     var qs;
-    var projectName = req.params.projectName;
-    if(!projectName) {
-      res.status(400).send({error: "No project name specified"});
-      return;
-    }
-
     var cur = req.query.now || (new Date()).toISOString();
```
sedge: Nit: `now` vs `cur`

routes/new-project.js
```diff
@@ -5,21 +5,16 @@ var utils = require("./utils");
...
     var project = {
-      title: projectName,
+      title: config.constants.NEW_PROJECT,
```
sedge: Storing constants like this seems odd. Why not require it in this file?

routes/new-project.js
```diff
@@ -5,21 +5,16 @@ var utils = require("./utils");
...
       user_id: req.session.publishUser.id,
       date_created: cur,
       date_updated: cur
     };
 
     delete req.query.now;
+    delete req.query.cacheBust;
```
sedge: ugh, too bad we have to do this.

gideonthomas: +1

routes/new-project.js
gideonthomas: Yeah you're right! I think I'm actually gonna make this an env variable (so that its consistent on the front-end + backend) and create a separate file for constants

routes/rename-project.js
```diff
@@ -0,0 +1,25 @@
+var utils = require("./utils");
+
+module.exports = function(config) {
+  return function(req, res) {
+    var project = JSON.parse(JSON.stringify(req.session.project.meta));
```
sedge: I don't understand what you're doing here. It looks like you convert an object to a string, and then back to an object.

gideonthomas: This is the easiest way to clone an object in js (without the prototype). This will let me make changes to the object without changing the original object

routes/config.js
```diff
@@ -14,5 +18,6 @@ module.exports = {
   publishURL: env.get("PUBLISH_HOSTNAME"),
   editorHOST: editorHOST,
   editorURL: env.get("NODE_ENV") === "development" ? env.get("BRAMBLE_URI") + "/src" : env.get("BRAMBLE_URI") + "/dist",
-  cryptr: new Cryptr(env.get("SESSION_SECRET"))
+  cryptr: new Cryptr(env.get("SESSION_SECRET")),
+  constants: constants
```
sedge: This isn't so bad. I still find it odd though.

routes/rename-project.js
sedge: Riiiight. You've explained this before :P

public/scripts/editor/js/bramble-editor.js
```diff
@@ -67,6 +52,21 @@ define(function(require) {
...
       Bramble.mount(config.root, config.open);
-    });
+    }
+
+    if(!authenticated) {
+      // Anonymous (not signed in) user
+      makeDetails = ProjectFiles.generateDefaultProject();
+    } else if(makeDetails.isNew) {
+      // Authenticated user creating a new project
+      makeDetails = ProjectFiles.generateDefaultProject(makeDetails.title, makeDetails.root);
```
sedge: The title is also a default in this case isn't it?

public/scripts/editor/js/fc/project-rename.js
```diff
@@ -0,0 +1,89 @@
...
+    this.renameButton = $("#navbar-rename-project");
+    this.titleBar = new InputField(this.container, true);
+    this.titleBar.id = "project-title";
+    this.titleBar.val(projectTitle);
+    this.fs = Bramble.getFileSystem();
+
+    this.container.one("click", rename.bind(this));
+  }
+
+  function init(appUrl, authenticated, csrfToken, projectTitle) {
```
sedge: I'm not sure I like this abstraction. Shouldn't you just be returning the constructor? It would also make the semantics less confusing:

`ProjectRename.init(...)` vs `new ProjectRenameUtility(...)`

public/scripts/editor/js/bramble-editor.js
gideonthomas: Yeah, but it is sent by the server

public/scripts/editor/js/fc/project-rename.js
gideonthomas: I mainly did this coz I couldn't think of good names for renaming, but yeah I can return the constructor

public/scripts/editor/js/fc/project-rename.js
```diff
@@ -0,0 +1,89 @@
...
+    context.titleBar[isSave ? "save" : "edit"]();
+    context.titleBar[isSave ? "removeClass" : "addClass"]("text-input");
+    context.container[isSave ? "addClass" : "removeClass"]("shadow");
+  }
+
+  function persist(title, callback) {
+    var appUrl = this.appUrl;
+    var csrfToken = this.csrfToken;
+
+    if(!this.authenticated) {
```
sedge: If we're letting them change the title when they aren't authenticated, does that mean that we store that somewhere in case they log in?

sedge: i.e. so it can be persisted then?

gideonthomas: Yeah probably, we'll have to think more about the unauthenticated-->authenticated transition coz there's a lot to consider (this being one of them)

public/scripts/editor/js/fc/bramble-input-field.js
```diff
@@ -0,0 +1,139 @@
...
+           $("<span></span>") : $("<input type=\"text\" />");
+  }
+
+  // This creates a morphable input field and takes in the
+  // container (should be a div as a DOM element or JQuery object)
+  // that will contain the input field. The constructor also takes
+  // a second optional boolean argument that will force the input field
+  // to be added before any other elements in the container (by default,
+  // it is appended to the container)
+  // The input field starts of as a static placeholder
```
sedge: *starts* ***off*** *as*

public/scripts/editor/js/fc/project-rename.js
gideonthomas: Sorry, I misread. We can get the value whenever we want from the textfield. What we do with that when we want to log in is a different issue altogether

public/scripts/editor/js/fc/bramble-input-field.js
```diff
@@ -0,0 +1,139 @@
...
+  function InputField(container, prepend) {
+    // Internal values
+    this._container = container instanceof $ ? container : $(container);
+    this._element = generateElement(InputField.Modes.STATIC);
```
sedge: Why is this separated out? It looks like you only use the `generateElement` function here.

gideonthomas: Didn't want to put that logic inside the constructor and clutter it up

public/scripts/editor/js/fc/bramble-input-field.js
```diff
@@ -0,0 +1,139 @@
...
+    var value = context.val();
+    var element;
+
+    context.mode = InputField.Modes.STATIC;
+
+    element = create(context);
+    context._element.replaceWith(function() {
+      return element;
+    });
+    context._element = element;
```
sedge: This looks a little weird. Is this line necessary?

gideonthomas: Apparently so, it seems like the `replaceWith` method creates a new jquery object, so I have to update the reference.

public/scripts/editor/js/fc/bramble-input-field.js
sedge: Ah, okay. I tend to only factor out logic if the concerns change or if there's redundancy.

public/scripts/editor/js/fc/bramble-ui-bridge.js
```diff
@@ -1,4 +1,7 @@
-define(["jquery", "fc/publisher", "fc/bramble-keyhandler"], function($, Publisher, KeyHandler) {
+define(function(require) {
+  var $ = require("jquery");
+  var Publisher = require("fc/publisher");
+  var KeyHandler = require("fc/bramble-keyhandler");
```
sedge: I've noticed you do this all over your patch. This was good thinking.

public/scripts/editor/js/fc/load-project-files.js
```diff
@@ -16,136 +21,147 @@ define(["jquery", "constants", "text!fc/stay-calm/index.html", "text!fc/stay-cal
...
+          return;
+        }
 
-      function checkProjectLoaded() {
-        if(completed === length) {
-          callback(null, {
-            root: root,
-            open: filePathToOpen
-          });
+        if(config.persist) {
```
sedge: What does this represent?

public/scripts/editor/js/fc/load-project-files.js
```diff
@@ -16,136 +21,147 @@ define(["jquery", "constants", "text!fc/stay-calm/index.html", "text!fc/stay-cal
...
+    // generate default files for the project. Once that is done,
+    // we can directly update the Bramble filesystem
+    if(!authenticated || options.isNew) {
+      project.dateCreated = (new Date()).toISOString();
+      project.dateUpdated = project.dateCreated;
+
+      // Persist the new project's files to the server for an
+      // authenticated user
+      if(options.isNew) {
+        config.persist = true;
```
sedge: Ah, I see it now. I'd change this to `authenticated`

sedge: Hmm, but not every updateFs() call will persist. Ignore my comments

public/scripts/editor/js/fc/load-project-files.js
gideonthomas: If we should persist the file or not

public/scripts/editor/js/fc/load-project-files.js
gideonthomas: I'll add some more comments to make it clearer

public/scripts/editor/js/fc/bramble-input-field.js
```diff
@@ -0,0 +1,139 @@
...
+
+    listenerList.forEach(function(listener) {
+      listener.apply(context, args);
+    });
+  }
+
+  function create(context) {
+    var element;
+
+    if(context.mode === InputField.Modes.STATIC) {
```
sedge: Reuse `generateElement()` here.

public/scripts/editor/js/fc/bramble-input-field.js
```diff
@@ -0,0 +1,139 @@
...
+    triggerListeners(context, context._listeners.change, [InputField.EventTypes.SAVE]);
+  };
+
+  // Morph the field into an editable input field
+  InputField.prototype.edit = function() {
+    var context = this;
+    var value = context.val();
+    var element;
+
+    context.val("");
```
sedge: What if you want to cancel the rename? Is that possible at this point? Or has the old value been lost?

public/scripts/editor/js/fc/bramble-input-field.js
```diff
@@ -0,0 +1,139 @@
...
+    context.mode = InputField.Modes.EDIT;
+
+    element = create(context);
+    context._element.replaceWith(function() {
+      return element;
+    });
+    context._element = element;
+    context._element.focus().val(value);
```
sedge: Why the `.focus()`?

public/scripts/editor/js/fc/bramble-input-field.js
sedge: Is this step necessary? You replace the value on line 95

sedge: After this element is destroyed...

public/scripts/editor/js/fc/bramble-input-field.js
gideonthomas: Because creating an input element in the DOM does not automatically give it focus. So, you wouldn't want the user to have to click into the input field to type after clicking on the rename button

public/scripts/editor/js/fc/bramble-input-field.js
```diff
@@ -0,0 +1,139 @@
...
+
+  InputField.prototype.removeClass = function() {
+    this._element.removeClass.apply(this._element, arguments);
+  };
+
+  // Polyfill: Allow listeners to listen for events triggered by
+  // the input field
+  InputField.prototype.on = function(eventName, listener) {
+    if(!this._listeners[eventName]) {
+      this._listeners[eventName] = [];
```
sedge: Don't we only have two event types that we trigger?

public/scripts/editor/js/fc/bramble-input-field.js
gideonthomas: Let me try!

public/scripts/editor/js/fc/bramble-input-field.js
gideonthomas: Yeah, but this entire module is meant to be extensible so that features can be added later on. This makes it future-proof

public/scripts/editor/js/fc/bramble-input-field.js
gideonthomas: Good catch!

routes/utils.js
```diff
@@ -31,8 +32,43 @@ function updateProject(config, token, data, callback) {
...
+    if(response.statusCode !== 200) {
+      callback(response.body, response.statusCode);
+      return;
+    }
+
+    var files = JSON.parse(body);
+    session.project.files = {};
+    files.forEach(function(file) {
+      var fileMeta = JSON.parse(JSON.stringify(file));
+      delete fileMeta.buffer;
```
sedge: This is no good. We need to add a way for this not to be returned in the first place.

gideonthomas: What do you mean?

sedge: Filed: https://github.com/mozilla/publish.webmaker.org/issues/117

sedge:
I haven't tested, but r+ on code

PR #710, opened 2015-07-20
Fix #645 - Don't use CDNs, improve requirejs config/loading

### LINKED ISSUE #645: Let's not use CDNs
We load jquery and jquery ui through CDNs for our app. But considering our offline use case, we should prefer not to do this especially since we use jquery extensively and thimble breaks completely if I have no internet connectivity.

humphd:
The way that requirejs config/loading was being done was odd, since it spread it across multiple top-level script loads in order to do the ssooverride stuff on its own.  I've fixed this so they share a single config.

I had to use jquery vs. zepto since it lacks the `fade*()` methods we use, and I don't want to have to do a custom build to get them.  I don't think it's worth trying to save ~5-10K for all the trouble.


sedge:
r+ on the code

PR #716, opened 2015-07-20
Fix #711 - Enter and Esc keys should work on project rename field

### LINKED ISSUE #711: When editing the Project Title, "enter" key should save the edits.
Referring to this state here...

image

If I hit **Enter**, Thimble should attempt to save my changes.

humphd:
The event state becomes a bit trickier when you can click or press enter, so I have to do more work to manually add/remove event listeners, which makes the `project-rename.js` code a bit more complicated.

One thing I notice about this UX for another bug: there is no non-keyboard way to cancel a rename once you start.  I wonder if we should have a way to do "Save" or "Cancel" (checkmark and an x?).  cc @flukeout


gideonthomas:
r=me

PR #721, opened 2015-07-21
Fix #626: add support for multipart uploads

### LINKED ISSUE #626: add support for multipart uploads
Right now the upload limit is set to 3mb (in brackets) and 5mb (in express's body parser), but these are set up in a way that causes a network hiccup during upload to require resending the entire file. It would be a good idea to switch over to multipart uploads, so that a network hiccup won't require resending the entire file, but only the most recently failed part.

humphd:
This is a WIP, since I can't get the full dev environment working locally.

### DIFF
--- a/public/scripts/editor/js/fc/filesystem-sync.js
+++ b/public/scripts/editor/js/fc/filesystem-sync.js
@@ -16,12 +16,14 @@ define(["jquery"], function($) {
   function pushFileChange(url, csrfToken, fs, path) {
     var context = this;
     var options = {
-      contentType: "application/json",
       headers: {
         "X-Csrf-Token": csrfToken
       },
       type: "PUT",
-      url: url
+      url: url,
+      cache: false,
+      contentType: false,
+      processData: false
     };
 
     function send() {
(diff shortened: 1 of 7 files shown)


public/scripts/editor/js/fc/filesystem-sync.js
```diff
@@ -52,11 +54,11 @@ define(["jquery"], function($) {
...
-      options.data = JSON.stringify({
-        path: path,
-        buffer: data,
-        dateUpdated: (new Date()).toISOString()
-      });
+      var formData = options.data = new FormData();
+      formData.append("dataUpdated", (new Date()).toISOString());
+      // Don't worry about actual mime type, just treat as binary
+      var blob = new Blob([data], {type: "application/octet-stream"});
+      formData.append("brambleFile", blob, path);
```
gideonthomas: I wonder if sending the path like this is such a good idea because of [this](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append#Browser_compatibility). It says that append with filename may or may not work with IE, Opera, Safari, etc. It might be safer to just send it as a form field?

PR #729, opened 2015-07-22
Fix #708: remove 'Show in public gallery' toggle from the Publish dialog

### LINKED ISSUE #708: Remove "show in public gallery" toggle from the Publish dialog.
Are we keeping track of this setting at all?

Even if we are, we should remove / hide this UI as it doesn't have any effect on the published product since we don't have a gallery.

humphd:
Whoever reviews this, please also test it, since I can't get publish working locally atm.


gideonthomas:
r+

PR #733, opened 2015-07-24
Support remix

gideonthomas:


routes/remix.js
```diff
@@ -0,0 +1,56 @@
+var request = require("request");
+var utils = require("./utils");
+
+module.exports = function(config) {
+  return function(req, res) {
+    var user = req.user;
+    var publishURL = config.publishURL + "/remix/" + req.params.projectId;
+
+    if(!user) {
+    }
```
sedge: What's this?

gideonthomas: Still need to add logic for non-authenticated users

app.js
```diff
@@ -224,6 +225,11 @@ app.put('/unpublish',
         middleware.validateRequest(["description", "dateUpdated", "public"]),
         routes.unpublish);
 
+app.get('/remix/:projectId',
+        middleware.setUserIfTokenExists,
+        middleware.publishLogin(true),
```
sedge: This looks wrong to me. How can someone remix if they aren't logged into thimble?

sedge: Whoops! You aren't done yet. Mah bad

lib/middleware.js
```diff
@@ -538,6 +539,45 @@ module.exports = function middlewareConstructor() {
           next();
         }
       };
+    },
+
+    /**
+     * Login with publish.webmaker.org for the user
+     */
+    publishLogin: function(allowUnauthUser) {
```
sedge: `allowUnauthUser` -> `allowAnonymous`

app.js
gideonthomas: So if you aren't logged into thimble, you still can remix a project. It will just drop the entire project into the editor but you wont be able to persist it!

lib/middleware.js
```diff
@@ -508,19 +509,20 @@ module.exports = function middlewareConstructor() {
...
-     * has been loaded
+     * Check if if a project has been loaded for the current context
      */
-    isProjectLoaded: function(req, res, next) {
-      if(!req.session.user) {
-        return next(utils.error(401, "Unauthorized request"));
-      }
+    isProjectLoaded: function(allowAnonymous) {
+      return function(req, res, next) {
+        if(!req.session.user && !allowAnonymous) {
```
sedge: This isn't great. We probably want a separate middleware for this piece. In fact, [we have one!](https://github.com/gideonthomas/thimble.webmaker.org/blob/remix/lib/middleware.js#L114) I think you should modify that one to allow anonymous instead of `isProjectLoaded`

lib/middleware.js
```diff
@@ -538,6 +540,45 @@ module.exports = function middlewareConstructor() {
           next();
         }
       };
+    },
+
+    /**
+     * Login with publish.webmaker.org for the user
+     */
+    publishLogin: function(allowAnonymous) {
```
sedge: This should probably be named `getPublishUser`

routes/utils.js
```diff
@@ -1,7 +1,113 @@
...
+
+function persistProjectFiles(config, user, project, data, callback) {
+  var publishURL = config.publishURL + "/files";
+  var files = {};
+
+  function persist(file, callback) {
+    var options = url.parse(publishURL);
+    options.method = "POST";
+    options.headers = { "Authorization": "token " + user.token };
+    var formData = new NodeFormData();
```
sedge: nit: newline above this line

routes/utils.js
```diff
@@ -1,7 +1,113 @@
...
+
+        if(response.statusCode !== 201) {
+          callback({ message: body, status: response.statusCode });
+          return;
+        }
+
+        files[body.path] = {
+          id: body.id,
+          project_id: body.project_id,
+          path: body.path
```
sedge: So we don't need the file contents here?

routes/utils.js
```diff
@@ -1,7 +1,113 @@
...
+        };
+
+        callback();
+      });
+    });
+  }
+
+  if(!user) {
+    callback(null, 200);
```
sedge: This is weirdly hacky.

routes/utils.js
```diff
@@ -65,10 +171,29 @@ function updateCurrentProjectFiles(config, token, session, project, callback) {
   });
 }
 
+function getRemixedProjectFiles(config, projectId, callback) {
+  var publishURL = config.publishURL + "/publishedProjects/" + projectId + "/publishedFiles";
+
+  request.get({ uri: publishURL }, function(err, response, body) {
+    if(err) {
+      console.error("Failed to send request to " + publishURL + " with: ", err);
+      callback(null, 500);
```
sedge: Shouldn't this not be null? It could help to tell a request failure apart from a publish error.

routes/utils.js
gideonthomas: no, because that's what's being sent into the function, so there's no reason for it to go out of the function too

public/scripts/editor/js/fc/load-project-files.js
```diff
@@ -1,58 +1,14 @@
...
-      console.error("[Bramble] Failed to send request to persist the file to the server with: ", err);
-      callback(err);
-    });
-  }
-
   function writeFile(config, path, data, callback) {
     var parent = Path.dirname(path);
 
     function write() {
+      console.log("Parent: ", parent, ", path: ", path);
```
sedge: Kill this

routes/utils.js
gideonthomas: Yeah I did this mainly to avoid the following in the main code:
```js
if(user) {
  ball(function() {
  });
  return;
}

apple(function() {
  ball(function() {
    cat();
  });
});
```
and instead just doing:
```js
apple(function() {
  ball(function() {
    cat();
  });
});
```

routes/utils.js
gideonthomas: Whatever is passed as `err` in the first parameter will be propagated to the client, we don't want that for request errors, hence the console log.

public/scripts/editor/js/fc/load-project-files.js
gideonthomas: Ah good catch

routes/get-current-project.js
```diff
@@ -15,11 +46,15 @@ module.exports = function(config) {
...
       }
 
-      res.type("application/json");
-      res.send({
-        project: project,
-        files: files
-      });
+      if(!user) {
+        files.forEach(function(file) {
+          file.path = Path.join(utils.getProjectRoot(project), file.path);
```
sedge: What would the root be in this case?

gideonthomas: "Untitled Project"

routes/utils.js
sedge: I'd prefer to see the conditional in the control logic to make the different code paths more clear. It would also get rid of the weird stubbing of callbacks

routes/utils.js
sedge: Agreed. That said, I think I'd prefer fewer cases to more. Right now you have:

1. `err` exists
2. an error happened, but `err` is null
3. success.

Shouldn't 1 & 2 be combined somehow? If you passed a generic status it would solve the problem, and you would be able to get rid of the 

```js
if(status === 500)
```
case entirely.

routes/utils.js
gideonthomas: I'm not sure I agree, I think the utility methods should be agnostic to whether a user exists or not! I think the right fix for this is to change **all** the utility methods to be in this form

sedge:
r+

Issue #573, opened 2015-06-17
Investigate ways to help authors add comments and tutorials to their projects.


flukeout:
This issue is now about investigating ways to help people create tutorials and leave more effective comments in their projects because..

* Current tutorial solution clutters the interface and is often ignored
* There's no clear UI for creating a tutorial or attaching it alongside your project
* Code comments can be messy when used extensively throughout a project

This ticket will be a UX / UI brainstorm dumping ground as we discuss and iterate on possible solutions.

flukeout:
Brainstorming "folded" inline comments...
![image](https://cloud.githubusercontent.com/assets/25212/8221093/11554008-150f-11e5-8cf6-3dbefab62b7f.png)

hannahkane:
I like this UX a lot.

Are code hints the same as comments? I thought code hints were, for example, brief explanations of tags automatically pulled in from a data source like MDN, whereas comments are more like...code comments that an author or reviewer adds themselves.

flukeout:
Here's an option for converting inline code comments into fancy, tutorial like comments.

![image](https://cloud.githubusercontent.com/assets/25212/8237129/f5953056-15a1-11e5-9ac0-aa7a121530a1.png)

jgmac1106:
I like this option. I don't think we would need a ton of formatting but this works better than the tutorial frame or the inline commenting overload.

For example here is a make being shared as part of #clmooc https://thimble.webmaker.org/project/155281/remix it is just full of comments being used as a tutorial. 

Being able to navigate through the comments would be really nice and allow learners to actually see the relationship in the code.

flukeout:
Here's a fancier version of the approach above that supports **markdown** and also **multiple highlighted lines* - all of this would be optional, but would offer additional functionality.

![image](https://cloud.githubusercontent.com/assets/25212/8238413/5951f0d6-15aa-11e5-8533-2eea22d427c3.png)

hannahkane:
This looks great. Still confused about code hints v. code comments. Aren't they different things?

flukeout:
Yeah so, in my mind..
* **Code Hints** pop up as I'm typing various HTML or CSS properties and help me auto-complete, this is a feature of the editor, and not something the Project author has control over
* **Comments** are bits of text added by the Project author to explain and annotate parts of the code

hannahkane:
Agreed. So this ticket is about comments, right?

jgmac1106:
[...]

PR #736, opened 2015-07-24
Initial UI for Thimble to use Bramble Tutorial feature, part of #573

humphd:
This is only a start, but it lets us really get a feel for things.  I still need a visual way to add the tutorial.html, but adding one manually (right-click in file tree, name your file `tutorial.html`) will have the right effect.

I keep track of whether or not there is a tutorial, and show/hide the TUTORIAL button accordingly.  I have to fix one bug in Bramble where deleting an active `tutorial.html` file leaves you in tutorial mode--it doesn't cause problems, but it's not right.  I'll file after this.

### DIFF
--- a/public/scripts/editor/js/fc/bramble-ui-bridge.js
+++ b/public/scripts/editor/js/fc/bramble-ui-bridge.js
@@ -165,6 +165,56 @@ define(function(require) {
       bramble.refreshPreview();
     });
 
+
+    // Preview vs. Tutorial preview mode. First check to see if there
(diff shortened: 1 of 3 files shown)


flukeout:
R+ seems sane to me!

gideonthomas:
r+

humphd:
I've done a fix for the "`tutorial.html` file vanishes while in override mode" bug in Bramble here: https://github.com/humphd/brackets/pull/393

PR #779, opened 2015-08-04
Fix #778 - Refreshing the editor on a remixed project causes loss of content

### LINKED ISSUE #778: Refreshing the page on a remixed project reverts back to the "stay calm" project
STR:
* **As an unauthenticated user**, open a published project and remix it. It loads as it should in the editor.
* Now refresh the page and the project retains it's name but the contents are reverted back to the default stay calm project.

gideonthomas:


sedge:
r+

PR #781, opened 2015-08-04
Fix #775 - Make the size of cookies smaller

### LINKED ISSUE #775: Make the size of a thimble cookie smaller
For a temporary fix for https://github.com/mozilla/thimble.webmaker.org/issues/774, lets make the size of the session cookie smaller by doing the following:
* Remove the `project_id` from each file representation and use it from `project.meta.id` instead
* Change the files object from:
```js
files: {
  "/a/b": {
                 path: "/a/b",
                 id: 1
             }
}
```
to
```js
files: [id1, path1, id2, path2...]
```

gideonthomas:


routes/delete-file.js
```diff
@@ -35,7 +35,7 @@ module.exports = function(config) {
         return;
       }
 
-      delete req.session.project.files[path];
+      req.session.project.files.splice(req.session.project.files.indexOf(existingFile.id), 2);
```
humphd: Can you move the `splice` stuff into `utils` too?  Let's hide all this nonsense of packing the files from callers.

routes/utils.js
```diff
@@ -84,11 +84,11 @@ function persistProjectFiles(config, user, project, data, callback) {
           return;
         }
 
-        files[body.path] = {
+        files.push({
           id: body.id,
           project_id: body.project_id,
```
humphd: Should this come out and go in the meta block?

gideonthomas: So this is what gets sent back to the user! This is not being stored in the session

humphd: Got it, thanks for clarifying.

PR #786, opened 2015-08-05
Fix #768 - Use symlinks for project roots

### LINKED ISSUE #768: Use symlinks for project roots
Currently for an unauthenticated user, our project root looks something like `/<project name>`. This makes renaming a hassle because we need some way to tell Bramble that the mounted path has been renamed (which is complicated to implement). Instead, we should have a symlink with the name of a uuid which will link to the real project path. We will mount the symlink in Bramble so that renaming the real project root will not affect anything.

humphd:


public/scripts/editor/js/fc/load-project-files.js
```diff
@@ -58,8 +54,8 @@ define(function(require) {
 
     files.forEach(function(file) {
       // TODO: https://github.com/mozilla/thimble.webmaker.org/issues/603
-      if(!filePathToOpen || Path.extname(filePathToOpen) !== ".html") {
-        filePathToOpen = Path.relative(config.root, file.path);
+      if(!config.filePathToOpen || Path.extname(config.filePathToOpen) !== ".html") {
+        config.filePathToOpen = Path.relative(config.root, file.path);
```
sedge: This makes way more sense than what I did.

sedge:
r+

Issue #754, opened 2015-07-30
Renaming the index.html file in the default project causes some weirdness


flukeout:
**Steps to reproduce**
* Create a new project or open an existing stay calm project
* Rename index.html to something else (in my case, staycalm.html)
* Refresh the preview

What happens

![image](https://cloud.githubusercontent.com/assets/25212/8991181/c9882286-36ab-11e5-84d1-a18f915ba589.png)

**List of Problems**
* Strange filename listed at the top of the Editor window
* The page no longer gets the styles specified in style.css
* In my case, editor is stuck showing the "saving..." message at the top

**Also**
* If you refresh the browser, the index.html file is back and the staycalm.html file previews and renders correctly.

After refreshing...

![image](https://cloud.githubusercontent.com/assets/25212/8991216/1f46468a-36ac-11e5-84d9-0d4dd266d02b.png)

humphd:
Looks to me like the file rename is using an absolute vs. project-relative path, and the HTMLRewriter or CSSRewriter is probably using that path and not finding assets relative to it.

humphd:
For the rename path being too big, I need to do what's in https://github.com/mozilla/thimble.webmaker.org/pull/795/files#diff-9748587150ca9269d951b5b55c578dc8R283 and strip off the project root.

humphd:
@flukeout one thing I should also point out, we haven't written the code yet to persist renames to the publish server (i.e., it will work locally but not save it).  So I'd avoid having people rename files for MVP.

sedge:
This has been fixed AFAIK. @flukeout Can you confirm?

flukeout:
Renamed "index.html" to "blam.html", refreshed, and everything works great.

![image](https://cloud.githubusercontent.com/assets/25212/9364928/a30331d8-4663-11e5-9d91-0ecdb16a3448.png)

PR #795, opened 2015-08-06
Fix #788 - transform paths from symlinked to real project paths

### LINKED ISSUE #788: There's an .mnt folder in my project
image

No idea how it got there!

humphd:
This takes a path like `/.mnt/1234asd;flkj234j/index.html` and transforms it into `/7/projects/12/index.html` so that we can send it to the server.  It depends on https://github.com/humphd/brackets/pull/404, which I'm about to land, but FYI.  Also, I can't test this locally, so I need to test it with someone on appear.

This should also fix the `Strange filename listed at the top of the Editor window` case in #754.

### DIFF
--- a/public/scripts/editor/js/fc/path-utils.js
+++ b/public/scripts/editor/js/fc/path-utils.js
@@ -1,0 +1,83 @@
+define(function() {
+  var Path = Bramble.Filer.Path;
+  var uuid = require("uuid");
+  var fs = Bramble.getFileSystem();
+  var sh = fs.Shell();
+
+  // Bramble loads all projects via symlinks stored in entries named /.mnt/<uuid>
+  var MOUNT_DIR = "/.mnt";
+
+  function createMountPoint(root, callback) {
+    var mountPoint = Path.join(MOUNT_DIR, uuid.v4());
+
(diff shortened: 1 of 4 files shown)


public/scripts/editor/js/fc/path-utils.js
```diff
@@ -0,0 +1,83 @@
...
+  var sh = fs.Shell();
+
+  // Bramble loads all projects via symlinks stored in entries named /.mnt/<uuid>
+  var MOUNT_DIR = "/.mnt";
+
+  function createMountPoint(root, callback) {
+    var mountPoint = Path.join(MOUNT_DIR, uuid.v4());
+
+    sh.mkdirp(MOUNT_DIR, function(err) {
+      if(err && err.code !== "EEXIST") {
```
sedge: Shouldn't this just be `if(err)`? If we're generating a UUID it shouldn't exist.

humphd: This is creating `/.mnt`, which might exist.

humphd:
We're going to revert the symlink stuff, so this is unnecessary.

PR #800, opened 2015-08-06
Fix #680 - Add (MVP level) confirmation before deleting a project

### LINKED ISSUE #680: Add confirmation before deleting a project
Seems too easy to accidentally delete a project. Can we add an "Are you sure?" step?

humphd:


sedge:
r+, nice.

PR #808, opened 2015-08-07
Fix #773 - prevent publishing if a save or sync is in progress

### LINKED ISSUE #773: Prevent publishing if a save or sync is in progress
While saving of files is occurring, or while they're being persisted to the publish server, the user shouldn't be able to attempt a publish.

humphd:


public/scripts/editor/js/fc/publisher.js
```diff
@@ -93,7 +101,7 @@ define(function(require) {
     }
 
     function run() {
-      CloseWarning.enable();
+      SyncState.syncing();
```
gideonthomas: This should be on line 127 instead

gideonthomas:
r+

PR #809, opened 2015-08-07
Fix #805 - Change green in the header to match homepage

### LINKED ISSUE #805: Change green in the header to match homepage
* The editor green bar header background color should be #2D9961
  * The user menu link "active" color should be #268C54
* The add file plus button color should be #2D9961

Need these small changes so that the editor matches the colors from the homepage for the video.

humphd:
@flukeout, I'm not sure if I've done what you want for the 2nd bit about the "active" state on the menu, can you review this please?


flukeout:
R+ great!

PR #811, opened 2015-08-07
Fix 'Uncaught SyntaxError: Unexpected end of input' in /projects/ page by not loading Bramble, main.js

humphd:
We currently load Bramble and the entire code for main.js when we load the projects list.  We don't need all that, and it causes errors.

I need this to get tested locally.


sedge:
r+

Issue #627, opened 2015-06-29
Signing in causes me to lose my work.


flukeout:
Steps to reproduce:
* Go to https://bramble.mofostaging.net/
* Make some edits to the starter project
* Click Sign In
* Complete the Sign In process
* You end up on the **Your Projects** view
* You've lost the changes you made prior to Signing In

You should end up in the editor, with the same content you had before signing in.

**Suggested Flow**
When someone signs in or creates an account, we should save whatever work they had in the editor as a new project and return the user there after they complete the sign up/in process so they can continue editing.

humphd:
cc @gideonthomas, @sedge

flukeout:
So does refreshing the browser while I'm signed in.

gideonthomas:
So, if you do some work in thimble before signing in and then sign in, we don't have any UI/UX to alert the user whether they want to discard existing changes or to create a new project out of those changes, etc. This is part of the anonymous user's workflow which we haven't implemented yet. Hence, for now, you lose all those changes and you'll be shown the "Your Projects" page.

As far as refreshing the browser goes, we had it always show you the "Your Projects" page (knowing that we would have to change it eventually), but I can fix that right now so that it opens up whichever project you previously had!

gideonthomas:
Part of this fix is [here](https://github.com/mozilla/thimble.webmaker.org/pull/640)

flukeout:
Updated the ticket description proposed a solution.

flukeout:
ping @gideonthomas - how is progress on this one?

sedge:
@flukeout I'm going to take this on

sedge:
This is blocked on #570

gideonthomas:
This is now blocked on https://github.com/mozilla/thimble.webmaker.org/issues/756

sedge:
@humphd and I did some thinking on this today. We figure there are two options, both of which hinge on storing anonymous projects in a special folder:

1. When a signed-in user opens thimble, and there are anonymous projects in the folder, open a dialog telling the user that there are unsaved projects. The user would be able to save each project from here, and when they did so, the anonymous project would be deleted after it had been upgraded. This isn't ideal because of the huge amount of UX/UI work that would be needed to get it running. 
2. We make the anonymous user flow stateless. 

Our raw notes for number 2 are on [this etherpad](https://etherpad.mozilla.org/thimbleanon), but here's the summary:

***Unauthenticated project flow***
[...]

PR #815, opened 2015-08-07
Server side fix for #627

gideonthomas:


Pomax:
adding a crosslink comment to #627

PR #823, opened 2015-08-11
Fix #821 - add Google Verification page for bramble.mofostaging.net

### LINKED ISSUE #821: Add a google verification page to bramble.mofostaging.net

@humphd Can you help?

humphd:


humphd:
@flukeout r?

humphd:
@jbuck says don't do it this way, but solve with something more general.

PR #825, opened 2015-08-11
Offload remix logic(copying projects) onto publish.webmaker.org for authenticated users

gideonthomas:
Relies on mozilla/publish.webmaker.org#136 landing


sedge:
This looks about right. r+, but hold off on merging until we get the publish piece in and test it all.

PR #827, opened 2015-08-11
Pass files from publish.wm.org->thimble server->thimble client using tar streams

gideonthomas:
This patch also removes some state tracking on the server-side viz. project root and files


default/index.js
```diff
@@ -32,13 +32,29 @@
...
 var Path = require("path");
 var MemoryStream = require("memorystream");
+var Tar = require("tar");
+
+var BUFFER = "BUFFER";
+var FILE_STREAM = "FILE STREAM";
+var TAR_STREAM = "TAR STREAM";
 
 var defaultProjects = {};
+var _defaultProjectPaths = {};
```
sedge: if you're going to do this here, do it on line 56 as well.

sedge: the prefix that is

routes/get-file-metadata.js
```diff
@@ -0,0 +1,30 @@
...
+
+module.exports = function(config) {
+  return function(req, res) {
+    var project = JSON.parse(JSON.stringify(req.session.project.meta));
+    var user = req.user;
+    var getMetadata = utils.getProjectFileMetadata;
+    var params = [config];
+
+    if(!user && req.session.project.remixId) {
+      getMetadata = utils.getRemixedProjectFileMetadata;
```
sedge: You never invoke this function.

gideonthomas: Now I do

routes/create-or-update-file.js
```diff
@@ -86,6 +76,7 @@ module.exports = function(config) {
 
         response.on('end', function() {
           body = JSON.parse(body);
+          delete body.buffer;
```
sedge: This makes me think that we shouldn't bother sending the buffer back when a new file is created.

sedge: OR updated.

routes/delete-file.js
```diff
@@ -3,29 +3,25 @@ var utils = require("./utils");
...
-    var path = utils.stripProjectRoot(req.session.project.root, req.body.path);
+    var path = req.body.path;
     var user = req.user;
     var project = req.session.project.meta;
-    var existingFile = utils.getFileFromArray(req.session.project.files, path);
+    var fileId = req.params.fileId;
 
-    if(!existingFile) {
+    if(!fileId) {
       res.status(400).send({error: "No file representation found for " + path});
```
sedge: This isn't the correct error message anymore IMO, since it means a piece of data was omitted.

routes/home.js
```diff
@@ -9,7 +9,8 @@ module.exports = function(config) {
     }
     var project = req.session.project && req.session.project.meta;
     var makedetails = encodeURIComponent(JSON.stringify({
```
sedge: How much extra hassle would it be to change this to `projectDetails`?

sedge:
r+

routes/create-or-update-file.js
gideonthomas: Sure that probably makes sense

routes/home.js
gideonthomas: I'm doing it in my next patch coz we've already changed it there

PR #829, opened 2015-08-11
Fix #828 - Rework project loading to use tarballs

### LINKED ISSUE #821: Add a google verification page to bramble.mofostaging.net

### LINKED ISSUE #828: Implement tar streaming for transferring files from server->client
Right now when we send the project files to the client, we send one large JSON object containing all the file buffers which are JSON parsed. This is how it is being received from publish.webmaker.org.
Instead we are teaching publish.wm.org to stream tarballs when requesting a collection of files (all files for a project). We should stream this down to the client as well.

On the client-side, this will also involve some other things besides teaching it to read a tar stream and untar:
* The server will no longer provide the project root in the "makedetails". Instead, the project id will be rendered into the makedetails object. From this, the client needs to figure out the project root! For authenticated users, it will be `/userId/projects/projectId`. In the makedetails, the projectId will be rendered as `id` and the userId will be rendered as `userID`. For an anonymous user, the path will be `/projectTitle` (which will change in https://github.com/mozilla/thimble.webmaker.org/pull/815 later).
* A consequence of the above is that when the client sends a path to the server, it **must** send a path sans the project root. For e.g. to request the server to update `/1/projects/5/index.html`, the client will send the path `/index.html`.

There probably will be a few more changes that a fix for this will need to involve...will update the bug when we come across them!

humphd:
This is a WIP, but sketches out the basics of doing project loading with tarballs.


public/scripts/editor/js/bramble-editor.js
```diff
@@ -58,21 +64,8 @@ define(function(require) {
...
-        return;
-      }
-
-      // Now that fs is setup, tell Bramble which root dir to mount
-      // and which file within that root to open on startup.
-      Bramble.mount(config.root, config.open);
-    }
-
     // Update the Project Title in the UI and allow it to be renamed
+    // TODO: should this be in BrambleUIBradge or Project or a combo?  Seems wrong here.
```
gideonthomas: The problem here is timing. If we do it in `BrambleUIBridge`, which only initializes after populating the fs, we are blocking on a long operation and will see an empty project title at the top until then. Maybe we should put this in `Project`.

public/scripts/editor/js/fc/filesystem-sync.js
```diff
@@ -28,72 +28,120 @@ define(function(require) {
       processData: false
     };
 
-    function send() {
+    path = Project.stripRoot(path);
```
gideonthomas: This should not happen here because we do a `readFile` below which will use this path (which does not exist)

public/scripts/editor/js/fc/filesystem-sync.js
```diff
@@ -28,72 +28,120 @@ define(function(require) {
...
+      request.fail(function(jqXHR, status, err) {
+        error = err;
+        console.error("[Bramble] Failed to send request to delete the file to the server with: ", err);
+      });
+      request.always(finish);
+    }
+
+    Project.getFileID(path, function(err, id) {
+      if(err) {
+        error = err;
```
gideonthomas: we should just pass this to `finish` now without assigning a global

public/scripts/project.js
```diff
@@ -0,0 +1,245 @@
...
+
+  // Installs a tarball (arraybuffer) containing the project's files/folders.
+  function installTarball(tarball, callback) {
+    var untarWorker;
+    var pending = null;
+    var root = getRoot();
+    var sh = new _fs.Shell();
+
+    function extract(path, data, callback) {
+      path = Path.resolve(root, path);
```
gideonthomas: This won't work. `Path.resolve("/1/projects/5", "/index.html")` results in `/index.html`. Instead, we should probably just follow the convention that we are sending `/index.html` and instead just call `addRoot(path)`.

public/scripts/project/project.js
```diff
@@ -0,0 +1,120 @@
...
+  function load(projectDetails, host, authenticated, callback) {
+    _user = projectDetails.userID;
+    _id = projectDetails.id;
+    _title = projectDetails.title;
+    _host = host;
+    _publishUrl = projectDetails.publishUrl;
+    _fs = Bramble.getFileSystem();
+
+    // TODO - what to do about these?
+//    if(!options.authenticated) {
```
gideonthomas: we probably don't care about these

public/scripts/project/remote.js
```diff
@@ -0,0 +1,142 @@
...
+      } else if(data.type === "error") {
+        finish(new Error("[Thimble error]: " + data.msg));
+      }
+    });
+
+    untarWorker.postMessage({file: tarball});
+  }
+
+  function loadTarball(fs, root, host, callback) {
+    // jQuery doesn't seem to support getting the arraybuffer type
```
gideonthomas: That really sucks

public/scripts/project/remote.js
```diff
@@ -0,0 +1,142 @@
...
+    var url = host + "/getFileContents";
```
gideonthomas: I wonder if we need to cache bust this as well as the `/getFileMeta` route

humphd: Won't hurt to do it.  Does it require any changes to server side to strip that at all?

gideonthomas: No since they're just xhrs

public/scripts/project/remote.js
```diff
@@ -0,0 +1,142 @@
...
+  var Buffer = Bramble.Filer.Buffer;
+
+  // Installs a tarball (arraybuffer) containing the project's files/folders.
+  function installTarball(fs, root, tarball, callback) {
+    var untarWorker;
+    var pending = null;
+    var sh = new fs.Shell();
+
+    function extract(path, data, callback) {
+      path = Path.resolve(root, path);
```
gideonthomas: This won't work. `Path.resolve("/1/projects/5", "/index.html")` results in `/index.html`. Instead, we should probably just follow the convention that we are sending `/index.html` and instead just call `addRoot(path)`.

public/scripts/project/remote.js
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  // Places project metadata (project id, file paths + publish ids) as an
+  // extended attribute on on the project root folder.
+  function setMetadata(fs, root, key, data, callback) {
+    // Data is in the following form, simplify it and make it easier
+    // to get file id using a path:
+    // [{ id: 1, path: "/index.html", project_id: 3 }, ... ]
+    var project = { id: data[0].project_id };
+    project.paths = data.map(function(info) {
```
gideonthomas: Umm, we don't want an array here do we? Shouldn't it be `project.paths = {}; data.forEach(function(info) { project.paths[info.path] = info.id; })`?

public/scripts/project/remote.js
```diff
@@ -0,0 +1,142 @@
...
+      }
+
+      // Step 2: download the project's metadata (project + file IDs on publsih) and 
+      // install into an xattrib on the project root.
+      loadMetadata(fs, root, host, key, function(err) {
+        if(err) {
+          return callback(err);
+        }
+
+        /** What to do about picking the file to open???
```
gideonthomas: Probably default to `index.html` if it exists otherwise recursively search for a deep `index.html` if it exists otherwise a random html file?

gideonthomas: I wonder if we should do this in a different bug! For now maybe just use the same logic before?

public/scripts/project/remote.js
```diff
@@ -0,0 +1,150 @@
...
+            files.forEach(function(file) {
+              // TODO: https://github.com/mozilla/thimble.webmaker.org/issues/603
+              if(!filePathToOpen || Path.extname(filePathToOpen) !== ".html") {
+                filePathToOpen = Path.relative(config.root, file.path);
+              }
+        ***/
+
+        callback(null, {
+          root: root,
```
gideonthomas: We no longer need to send the root

public/scripts/editor/js/fc/filesystem-sync.js
```diff
@@ -28,72 +28,115 @@ define(function(require) {
...
-        return;
+        console.error("[Thimble] Failed to read ", path, " with ", err);
+        triggerCallbacks(context._callbacks.afterEach, [err, path]);        
+      }
+
+      if(err) {
+        return onerror(err);
       }
 
       options.data = FileSystemSync.toFormData(path, data);
```
gideonthomas: path here should be stripped of it's project root

public/scripts/editor/js/fc/filesystem-sync.js
```diff
@@ -28,72 +28,115 @@ define(function(require) {
...
-        console.error("[Bramble] Server did not persist ", path, ". Server responded with status ", request.status);
-      }
-    });
-    request.fail(function(jqXHR, status, err) {
-      error = err;
-      console.error("[Bramble] Failed to send request to delete the file to the server with: ", err);
-    });
-    request.always(function() {
+
+    path = Project.stripRoot(path);
```
gideonthomas: This also should not happen here since `getFileId` does the stripping

public/scripts/editor/js/fc/filesystem-sync.js
```diff
@@ -28,72 +28,115 @@ define(function(require) {
...
+    function doDelete(id) {
+      var request = $.ajax({
+        contentType: "application/json",
+        headers: {
+          "X-Csrf-Token": csrfToken
+        },
+        type: "PUT",
+        url: url + "/" + id,
+        data: JSON.stringify({
+          path: path,
```
gideonthomas: I don't think we need the path anymore

public/scripts/editor/js/fc/filesystem-sync.js
gideonthomas: We probably do not need to strip project root at all in the function

PR #830, opened 2015-08-12
Fix #828 - Server and client implementation of Tar streaming files

### LINKED ISSUE #821: Add a google verification page to bramble.mofostaging.net

### LINKED ISSUE #828: Implement tar streaming for transferring files from server->client

gideonthomas:
This is a combination of https://github.com/mozilla/thimble.webmaker.org/pull/829 and https://github.com/mozilla/thimble.webmaker.org/pull/827


gideonthomas:
#828 fix

sedge:
This needs one more commit to be in a mostly working state: https://github.com/sedge/thimble.webmaker.org/commit/4c77961528636941c49850005555ad7ccc323469

Bugs I've noticed:

1. Pathing is still odd when a project opens. Sometimes styling doesn't appear in the live preview. Other times an error pops up saying that an `index.html` file is required for preview (despite it being in the tree)
2. Creating a new project sucks in `tutorial.css` and `tutorial-style-guide.html`. This is a regression, so something wonky has happened with the history.

humphd:
The index.html not having CSS is my mission for tomorrow.  It's a bug in the timing of the live dev server loading and it using raw file system blob urls vs. rewritten instrumented pages.

sedge:
The strange files popping up were artifacts in my indexeddb. Clearing it fixed that issue.

humphd:
...and https://github.com/humphd/brackets/pull/419 should fix 1)

PR #833, opened 2015-08-13
Lock access to project metadata xattr

humphd:
This wraps all read/write access to the project's metadata xattr in a lock, so we can safely parallelize file updates on the server, but queue them on the client.  It will fix https://github.com/mozilla/thimble.webmaker.org/issues/814.  This lock isn't safe across tabs, but I don't think we are going to support opening the same file across tabs?  If we want that, I just need to move the lock into a file vs. ram.


gideonthomas:
Code looks right

PR #840, opened 2015-08-13
Fix #824: Add UI for calling bramble.export()

### LINKED ISSUE #824: Update export icon
I'd love some kind of simple MVP solution that lets me call `bramble.export()`, which will create a `project.zip` file for you of all your files/folders, and download it to the browser.  Can we come up with something simple for MVP?  Maybe in the user's options at the top?

Now that I've done the code for this in Brackets, I think it's kind of a killer feature to support for MVP, since it lets users work totally anonymously and completes the import/export story.

cc @flukeout

humphd:
@flukeout still needs to get me a proper icon (I reused the sign-out one), but this works perfectly.


gideonthomas:
r+

PR #842, opened 2015-08-13
Fix #826 - Project metadata and Remix not showing up

### LINKED ISSUE #826: Project meta data and Remix button not showing up correctly on beta.
This is in Chrome.

screen shot 2015-08-11 at 10 32 39 am

humphd:
I *think* this is right, but I'm not an expert on jquery.  From my research, this seems to be the way they want you to go.

I've also bumped up the z-index on the remix details.

### DIFF
--- a/public/resources/remix/index.js
+++ b/public/resources/remix/index.js
@@ -88,18 +88,19 @@ function getMetadata($) {
 }
 
 function run() {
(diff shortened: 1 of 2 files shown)


gideonthomas:
looks right! r+

PR #850, opened 2015-08-13
Add rename handling, fix lock callbacks for error cases

humphd:
This seems to work like I want, but I see some issues with `Saving...` never going away, and I wonder if what I'm doing in the callbacks for `handleFileRename` are right.

### DIFF
--- a/public/scripts/editor/js/fc/filesystem-sync.js
+++ b/public/scripts/editor/js/fc/filesystem-sync.js
@@ -15,80 +15,67 @@ define(function(require) {
     });
   }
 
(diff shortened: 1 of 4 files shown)


public/scripts/editor/js/fc/filesystem-sync.js
```diff
@@ -15,80 +15,66 @@ define(function(require) {
     });
   }
 
-  function pushFileChange(url, csrfToken, fs, path) {
-    var context = this;
+  function fileChange(host, csrfToken, fs, path, callback) {
```
gideonthomas: where does callback come from?

PR #851, opened 2015-08-13
Bump z-index on remix button in published header

humphd:
The Remix button is sitting under the remix header, due to tag ordering.  This forces it to sit on top.


gideonthomas:
r+

Issue #848, opened 2015-08-13
Staging: Projects not deleting on my projects view


flukeout:
**Convoluted steps to reproduce**
* Go to My Projects list
* Delete a Project
* Immediately after, as the project is deleting, click where it says "Deleting..."
* This time click "Cancel" on the popup
* Sometimes, this causes the project to get stuck "Deleting..."
* Even if completes deleting and removes itself from the page, it's back when you refresh
* Additionally, you can't delete this project now
  * If you do, it's always back again when you refresh 

@humphd Thoughts on what might be happening? Can it be alleviated by not making the "Deleting..." indicator clickable?

humphd:
Things to do: 
* Make the `Deleting...` non clickable while delete is in progress.

flukeout:
Things to do:
* Never show ``Deleting...``` and just delete the project immediately.

humphd:
We also need to stop storing "file to open" in local storage -- it should go in the project metadata, or not be stored at all.

humphd:
I think this is fixed now, please re-file if you find other edge cases.

PR #853, opened 2015-08-13
Part of #848 - don't allow a second click after a delete has started

humphd:
This improves things, I think.


humphd:
@flukeout r?

flukeout:
R+ looks good to me, tested and the deleted project disappears right away which is nicer I think.

PR #854, opened 2015-08-14
Various tweaks to improve caching, compression, and remove unused css

humphd:
Just a few fixes based on things Chrome's auditor found.  There's probably more CSS that we can rip out, but I'm not confident to use tools to do it, since some content is dynamically inserted so tools can't find it.


gideonthomas:
r+

gideonthomas:
Merged in 30855e94dfd2e437273e64bd67cae0aa79a7d854

PR #870, opened 2015-08-17
Editor title bar UI cleanup

flukeout:
Made updates to a bunch of icons, fixing mostly positoning, hover & active states and size.


flukeout:
Will file from different branch.

Issue #868, opened 2015-08-17
UI Cleanup


flukeout:
There are lots of parts of the UI that can use a little polish and TLC. I've listed them all here.

**Tutorial toggle**
* [x] Update preview & tutorial toggle UI

**Refresh Icon**
* [x] Move the refresh icon to the right side
* [x] Update hover & active states
* ~~[ ] Add animation animation after it’s clicked?~~

**Icon Buttons**
* [x] Adjust size and position of all icons, many are too large
* [x] Remove the slow transition to the hover state, makes the editor feel slow
* [x] Make sure they all have hover & active states

**Editor theme colors**
* [ ] Finalize the editor theme colors for both the dark and light themes

**Button Cleanup**
* [ ] Fix the styling, hover and active states of a bunch of buttons.
  * [x] Main Publish button
  * [ ] Buttons in the Publish Dialog
  * [x] Rename Button
  * [x] New Project button on My Projects page

**User Menu**
* [x] Remove the 2px border above the sign out option
* [x] Update the "export" icon with something that connotes a download
* [x] Make sure the hover state background color for menu items in this menu matches the Add File menu
* [x] Change the green background to match the rename UI hover state when the menu is open
* [x] Adjust icon size & positioning

**Publish Menu**
* [x] Button Styling
  * [x] Active & Hover states
  * [x] Rounded corners
* [x] Fix spacing on the description label and textarea
* [ ] Nicer styling on the published project link
* [x] Sort out the 2px borders, they don't look nice.
* [x] Fix padding on the popup heading to match the contents

**Add File Menu**
* [x] Remove green hover effect

**Gear Menu**
* [x] Center + and - icons in the Text Size changer
* [x] Hover & active states for the plus and minus icons
* [x] Fix the styling of the theme toggle
  * [x] Borders are uneven when light theme is on
  * [x] Add cursor:pointer style to the theme toggle
* [x] Make the menu appear instantly instead of fading in
* [x] Center the arrow and the dialog below the gear icon

**Mobile Preview**
* [x] Different styling for the mobile preview toggle
* [x] Different background for the mobile mode (match the grey in the top bar)

**File Tray**
* [x] Expand Tray  and Add File icons need…
  * [x] Hover states
  * [x] Active states
  * [x] Consistent rounded corners
  * ~~[ ] Change the file tree expand / contract UI so that it’s in the same place for easy expanding and collapsing.~~
* [ ] Font size of the filename is too small when renaming

[...]

PR #873, opened 2015-08-17
Preview bar updates

flukeout:
Some minor updates to the Preview Bar
* Moved refresh icon to the right side of the preview bar
* Added different styles to the Preview and Tutorial toggle
  * Selected option is only underlined when there are two options
* White background on the mobile & desktop toggle to match the refresh icon

Part of the effort outlined in #868


sedge:
Way better. r+

PR #874, opened 2015-08-17
Style updates to the green header bar

flukeout:
Changes
* Updated Publish button styling and hover/active states
* Updated rename Save button styling and hover/active states
* Font size in the project rename input field now same as non-editable project name font size
* Removed style transitions from UI elements
* Added hover & active states to the Thimble logo
* Updated position of the rename project pencil icon

Continuation of work in #868


sedge:
r+

PR #875, opened 2015-08-17
Editor options menu updates

flukeout:
Updates to the gear menu
* Centered the Plus and Minus icons inside the green buttons
* Cleaned up Theme toggle styles
* Replaced "+" and "-" in the font size icons with SVG images


sedge:
thankyouthankyouthankyou r+

PR #876, opened 2015-08-17
Add file menu updates

flukeout:
**Changes**
* Removed extra spacing above the Upload... option
* Changed the harsh green hover states with a lighter grey

Continued work from #868


sedge:
Way way better. r+

PR #880, opened 2015-08-18
User dropdown menu UI updates

flukeout:
**Changes**
* Size tweak to the darker hover effect when the menu is open
* Added a 'cloud' icon for the export
* Dark green background color adjustment when the menu is open
* Menu icon size and positioning tweaks
* Reduced width of border above Sign up

Continuation of work in #868


sedge:
Looks great! This would also solve #824. r+

PR #881, opened 2015-08-18
Fix #750 - Allow different projects in different tabs

### LINKED ISSUE #750: Allow multiple thimble instances per browser session
The application is configured in such a way that it can only track one opened project at a time. Opening the application in more than one window/tab in the same browser session would have unexpected and disastrous results.

gideonthomas:


app.js
```diff
@@ -165,88 +165,111 @@ app.get("/login", routes.login);
...
+ * Main routes for thimble.webmaker.org
+ */
+// Entry point for all users
 app.get('/',
+        middleware.setUserIfTokenExists,
+        middleware.setPublishUser,
+        routes.main.root);
+
+// Main route for authenticated users
+app.get('/user/:username/:projectId',
```
sedge: Why do we need `:username` here? Won't we already have it on `req.user`?

app.js
```diff
@@ -165,88 +165,111 @@ app.get("/login", routes.login);
...
+/*
+ * Project operations
+ */
+// Get all projects for a user
 app.get('/projects',
         middleware.checkForAuth,
         middleware.setUserIfTokenExists,
         middleware.setPublishUser,
-        routes.projects);
+        routes.projects.read);
```
sedge: Finding the correct file is becoming sorta difficult with all of the abstraction. For example, this actually lives in `view-user-projects`. Can we make the filename/handler semantics more consistent somehow?

gideonthomas: I just changed how the routes file tree looks to be more like publish.wm.org. See if it makes more sense now

app.js
```diff
@@ -165,88 +165,111 @@ app.get("/login", routes.login);
...
 
-app.get('/getFileContents/:remixId?',
-        middleware.setUserIfTokenExists,
-        routes.getFileContents);
-
-app.get('/getFileMeta/:remixId?',
-        middleware.setUserIfTokenExists,
-        routes.getFileMetadata);
 
+/*
```
sedge: These separations would be great as separate files. What are the chances you could factor them out to clean up this file?

app.js
gideonthomas: This was purely for semantics. I can't do `/:projectId` because no other route would work. I wanted to do `/projects/:projectId` but I couldn't because there already is a route like that (new projects), and I didn't want to do `/user/:projectId` because I don't think it would make sense because it would be a mismatch between class and identifier (user->projectId). So just to prettify the url a bit (what the user will see), I added the username.

app.js
gideonthomas: I can try that

lib/middleware.js
```diff
@@ -597,7 +589,41 @@ module.exports = function middlewareConstructor() {
...
+        req.user.publishId = body.id;
+        next();
+      });
+    },
+
+    /* Sets the project for the current request based on the project
+     * id provided as a parameter. The request's `user` property must
+     * be set by calling the `setUserIfTokenExists` middleware operation.
+     */
+    setProject: function(req, res, next) {
```
sedge: This is poorly named, since it supports three cases: 1. No user, 2. user and no projectId,  3. user and projectId.

`maybeFetchProject` could work maybe?

gideonthomas: I'm not sure that sound's great either! This function basically set's the current context. What I can do is the following:
```js
if(!user || !projectId) {
  req.project = null;
  next();
}
```

sedge:
I'm about two thirds of the way through this review. I'll pick it up tomorrow!

gideonthomas:
@sedge I recommend doing the review based on commits because refactoring the route file structure shows up as a huge change! I basically haven't changed any server logic between 95d1ac9 and 3db78f1

routes/index.js
```diff
@@ -38,110 +15,23 @@ function getPageData(req) {
...
+    init: function(app, middleware) {
+      [
+        require("./auth"),
+        require("./main"),
+        require("./projects"),
+        require("./files"),
+        require("./tutorials")
+      ].forEach(function(module) {
+        module.init(app, middleware, config);
+      });
```
sedge: Nice reuse of this pattern

routes/main/homepage.js
```diff
@@ -0,0 +1,95 @@
...
+var Constants = require("../../constants");
+var utils = require("../utils");
+
+function getProjectMetadata(config, req, callback) {
+  var project = req.project;
+  var remixId = req.params.remixId;
+  var anonymousId = req.params.anonymousId;
+
+  if(anonymousId) {
+    anonymousId = anonymousId.replace("^id-", "");
```
sedge: Why is the anon id prefixed in the first place?

routes/main/index.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+  init: function(app, middleware, config) {
+    // Entry point for all users
+    app.get('/',
+      middleware.setUserIfTokenExists,
+      middleware.setPublishUser,
+      require("./root").bind(app, config));
```
sedge: This is a clever use of `.bind()`. I'm not sure I would have thought of this!

routes/main/root.js
```diff
@@ -0,0 +1,56 @@
...
+        res.sendStatus(500);
+      } else {
+        res.status(status).send({error: err});
+      }
+      return;
+    }
+
+    delete req.session.project.migrate;
+
+    res.redirect(301, "/user/" + user.username + "/" + project.id);
```
sedge: Doesn't the front-end still need the anonymousId of the project in order to know which one to upgrade?

routes/projects/read.js
```diff
@@ -0,0 +1,44 @@
...
+    if(response.statusCode !== 200) {
+      res.status(response.statusCode).send({error: response.body});
+      return;
+    }
+
+    var options = {
+      csrf: req.csrfToken ? req.csrfToken() : null,
+      HTTP_STATIC_URL: "/",
+      username: user.username,
+      projects: response.statusCode === 200 ? JSON.parse(body) : [],
```
sedge: This condition is unnecessary now because of the check on line 23. Just `projects: JSON.parse(body)`

routes/main/root.js
gideonthomas: It will coz it will be rendered into the page! It is temporarily stored in the session and deleted immediately in the next request

routes/files/index.js
```diff
@@ -0,0 +1,31 @@
...
+      middleware.setUserIfTokenExists,
+      read.metadata.bind(app, config));
+
+    // Create or update a file for a project for a user
+    app.put('/projects/:projectId/files/:fileId?',
+      middleware.checkForAuth,
+      middleware.setUserIfTokenExists,
+      middleware.validateRequest(["dateUpdated", "bramblePath"]),
+      middleware.setProject,
+      middleware.fileUpload,
```
sedge: Pull this up

routes/projects/index.js
```diff
@@ -0,0 +1,53 @@
...
+    // Unpublish an existing project for a user
+    app.put('/projects/:projectId/unpublish',
+      middleware.checkForAuth,
+      middleware.setUserIfTokenExists,
+      middleware.setProject,
+      middleware.validateRequest(["description", "dateUpdated", "public"]),
+      require("./unpublish").bind(app, config));
+
+    // Remix an existing project
+    app.get('/projects/:projectId/remix',
```
sedge: Can you change this to `/projects/:publishedId/remix`?

lib/middleware.js
sedge: Sure, let's do that.

sedge:
A few comments and questions. This is a great patch

routes/main/homepage.js
gideonthomas: Sorry, that is a regression! Will remove

PR #882, opened 2015-08-18
More UI adjustments

flukeout:
**Changes**
* Added hover and active states to the increase and decrease font size buttons
* Centered the Add File & Editor Options menu under the icons when they're opened
* Removed the fade in & out effect from the menus opening and closing


sedge:
r+

Issue #512, opened 2015-05-14
Unauthenticated users should be prompted to log in or sign up to Publish their work


hannahkane:
Right now the Save button is disabled for unauthed users, but we give no cues for how to enable it. 
Needs UX design.

flukeout:
In our current design, when a user is Signed Out and hits **Publish** - we have a couple of options:
* Redirect straight to the Sign In/Up page
* Show a modal that explains that you need to Sign Up/In with a link to do so

Neither of these are amazing, though.

Another option is that we can change the wording in the top bar slightly and then highlight the Sign In and Log In actions when someone hits the publish button, like this...

![image](https://cloud.githubusercontent.com/assets/25212/8140754/63f826b8-1113-11e5-8660-a6a9833d60e7.png)

@cade @simonwex Can we use our new Sign In / Sign Up process in a dialog of some sort, or do we have to leave Thimble and hit https://id.webmaker.org/login ?

cade:
> Can we use our new Sign In / Sign Up process in a dialog of some sort

I’m pretty sure this can be solved with a few dozen iframes. Agreed, @cadecairos?

flukeout:
Haha, sorry @cade - If we do end up using a dozen iframes, we'll make sure to credit you!

flukeout:
I've made this a bug because right now the Publish button is totally non-functional when a user is not signed in.

Alternatively, we could just replace the Publish button with text when a user is not signed in, like this.

"**Log In** or **Sign Up** to Save and Publish"

@hannahkane Thoughts?

hannahkane:
I'd prefer the copy change + pulsing, because I think it's important to keep the Publish button in tact for consistency. If replacing the button with text entirely is easier, I definitely think that's fine for MVP, though.

I think it's slightly weird that we say "to save or publish" since "saving" is not an action the user can actively take. Could we just say "Log In or Sign Up to -> [Publish button]"?

gideonthomas:
Fixed in b4a5f434106db3537803d356adacc7b20ebd6597

PR #886, opened 2015-08-18
Updates to the behaviour and styling of the Log In area.

flukeout:
Changes, when not signed in...
* Copy now says "Log In or Sign Up to Save or [Publish]" in a sentence-like structure
* When the Publish button is pressed, the Log In and Sign Up sentence is strobed for emphasis

Continuation of work in #868


sedge:
This will also address #512

sedge:
r+, though @hannahkane seems to want just "Log in or sign up to publish" vs "to save and publish".

PR #888, opened 2015-08-18
Small fix for scroll bug

flukeout:
This gets rid of the always-on scrollbar. Each pane has their own scroll behaviour so this shouldn't be a problem.

To test:
* Load up the editor
* There is no scrollbar along the right side of the screen
* There was before


sedge:
r+! But I'll merge this patch. Let me know when all your other ones have been merged

PR #890, opened 2015-08-19
Add requirejs build step to bundle everything in dist/main.js

humphd:
This isn't quite right yet, but it's close.  I basically don't want to see us load all the JS for Thimble in separate files--it should get built/minified.  I know we have longer-term plans to do a revamped code-structure for the front end, but for MVP, we should just make what we have work better.  I'll pick away at this


humphd:
I have this working in another PR coming in a sec.

PR #898, opened 2015-08-19
Sped up the theme toggle UI element

flukeout:
I changed it so that the transition speed is set with one variable and sped it up to 200ms.


gideonthomas:
r+

PR #901, opened 2015-08-20
Styling updates to the column headers

flukeout:
**Changes**
* Added a bolder version of Open Sans for the column headings (Files, Editor, etc.)
* Added a minimum width to all of the column headings so that the icons and UI doesn't jump to the line below (it gets cut off instead)
* Adjusted the font color for the column headings to compensate for their new boldness


sedge:
r+, looks good.

Issue #896, opened 2015-08-19
Staging: Editor not loading again (same error as before)


flukeout:
Got myself into this state after trying to load the editor...

![image](https://cloud.githubusercontent.com/assets/25212/9370519/c3003b76-4685-11e5-88a3-d6bb104a4a90.png)

This is the ```Uncaught TypeError: Cannot read property 'doc' of undefined``` we've seen before.

I'm able to resolve it by clearing my bramble cookies...

![image](https://cloud.githubusercontent.com/assets/25212/9370538/f9a85794-4685-11e5-880b-976665080eab.png)

Anyway - worried this is a bigger problem that users might run into at some point too.

flukeout:
I had no other tabs with thimble open.

humphd:
Can you please paste a full stack trace in this bug?  I can only see the top frame, and I'd love to see it all.

humphd:
I was able to hit this today.  I haven't been using the app a lot for a week, while on holidays.  I came back this morning, and the app was down (in https://github.com/MozillaFoundation/mofo-devops/issues/155 we have determined that our Heroku tier was sleeping the node app when inactive, and forcing it to start again).  After it restarted I went to an old project.  It tried to load, and eventually did the same thing as @flukeout reported above.  I did a bit more digging and confirmed that the filesystem is OK, and that the expected project files are in fact there.  I suspect this is the session getting corrupted/expiring.

![screenshot 2015-08-24 11 03 03](https://cloud.githubusercontent.com/assets/427398/9443634/173f4cea-4a50-11e5-94bb-f7433ac45a5f.png)

humphd:
We figured this out.  The issue is caused by setting the preview mode to "mobile" and when it loads it tries to change the preview mode and reload the iframe browser, which isn't set up yet.  It's a timing bug, and can be fixed by just avoiding the reload if the browser isn't ready yet.  I'll do a patch.

humphd:
Fixed!

PR #909, opened 2015-08-24
Thimble portion of fix for #896, also requires https://github.com/humphd/brackets/pull/440

humphd:
This stops us from re-setting the theme on startup, when all we need to do is change the UI for the buttons.


gideonthomas:
r+

PR #911, opened 2015-08-24
Change redirect responses to use HTTP 307 vs. HTTP 301

gideonthomas:


lib/middleware.js
```diff
@@ -521,7 +521,7 @@ module.exports = function middlewareConstructor() {
       if(req.session.user) {
         next();
       } else {
-        res.redirect(301, "/anonymous/" + uuid.v1() + qs);
+        res.redirect(307, "/anonymous/" + uuid.v1() + qs);
```
humphd: `v4()` vs. `v1()`

routes/projects/remix.js
```diff
@@ -5,7 +5,7 @@ module.exports = function(config, req, res) {
   var publishedId = req.params.publishedId;
   var user = req.user;
   if(!user) {
-    res.redirect("/anonymous/" + uuid.v1() + "/" + publishedId);
+    res.redirect(307, "/anonymous/" + uuid.v1() + "/" + publishedId);
```
humphd: `v4()` vs. `v1()`

routes/projects/remix.js
```diff
@@ -30,6 +30,6 @@ module.exports = function(config, req, res) {
       return;
     }
 
-    res.redirect(301, "/user/" + user.username + "/" + JSON.parse(body).id);
+    res.redirect(307, "/user/" + user.username + "/" + JSON.parse(body).id);
```
humphd: Are you at all worried about `JSON.parse` throwing?

humphd:
r=me with those things dealt with.

routes/projects/remix.js
gideonthomas: No coz publish is stable enough for us to trust the data it sends back

humphd: I think that's a bad idea, but I won't block this patch on that.

PR #915, opened 2015-08-24
Fix #914 - Remix Url added to the href of the link

### LINKED ISSUE #914: Add the Project Remix URL to the Remix button href attribute instead of triggering via javascript
Right now if you hover over the remix button, you get no information about what the URL is to remix something...

image

The Remix is triggered via Javascript, let's change it so that we update the href attribute of the Remix button and treat it like a normal link.

gideonthomas:


humphd:
r+

PR #918, opened 2015-08-25
Fix #917 - Use most recent metadata for published project to avoid conflicts

### LINKED ISSUE #917: Ensure the most recent metadata for a published project is used in the remix bar
In the editor...

image

In the details bar after publishing...

image

Published project url: https://d157rqmxrxj6ey.cloudfront.net/mozillalearning/62/

gideonthomas:


humphd:
Just add a comment above this line so it's clear why we're using the current value if present, r=me with that.

PR #920, opened 2015-08-25
Kill off CSP violations and reduce CSP warning spam on console

humphd:
This gets rid of all the annoying CSP warnings.  All that's left are warnings about us being in report-only mode, which is fine.


gideonthomas:
r+

Issue #541, opened 2015-05-29
Figure out starting content for editor.


flukeout:
What do we want our starting content to be a blank/new make? Currently, we have a ```p``` tag with a line of content, but I think we can do better. We can divide this into beginner and multi-file modes.

**Beginner**
This is the single page mode without a file browser and has...
* A couple of tags with content, maybe an ```h1``` and a ```p```
* A ```style``` tag in the ```head``` with...
  * Some basic styles for the text elements (font-size, color and font-family)
  * A background color for the ```body```

**Advanced**
Multi-file project with...
* index.html - same markup as beginner mode, but with an additional ```button``` element
  * *style.css* and *script.js* are linked in the ```head``` tag
* style.css - some basic styles (like listed above) for the elements on the index page
* script.js - a very basic code example, like a click handler for the ```button```

This will give people something to tinker with immediately.
@hannahkane @cassiemc what do you guys think about something like this? Have we experimented with different starting content before?

We could add an element of chaos to this too, and pick a different background-color for the ```body``` every time you start a new project.

hannahkane:
My comments are a bit outside the scope of this ticket, but it's difficult to talk about this initial experience without also talking about the experience around it.

Short answer: I think what you've proposed for beginner and advanced modes are fantastic. Like!

Longer answer: 
 * your suggestion implies that we can offer different starting points for these two different modes, perhaps by including the mode in the query string, as discussed earlier, right? That makes me wonder how the user can switch modes once they're in it. Like, if I end up on the Advanced mode, and I'm overwhelmed, can I become aware of Beginner mode and how to get there?
 * For many users, I imagine the right first experience will be to remix. Sometimes, there's nothing worse than being faced with a blank page! (Or even a blank page with a button on it.) Again, this is probably out of scope for this ticket, but maybe we should think about exposing pathways to existing content.

humphd:
Some technical issues to be aware of and some thoughts:
* Because we auto-preview HTML files when they are created, they must be valid HTML, which means we can't ever create a totally blank HTML file.  Luke's point above about having default content fits with this, but just for future discussions regarding clicking "New" or the like.
[...]

PR #925, opened 2015-08-26
Add default HTML, CSS, and JS files, use updated bramble.addNewFile()

humphd:
Watching my user testing videos, it became clear that when adding new files, providing something to start with would be nice vs. a totally blank page.  Depends on https://github.com/humphd/brackets/pull/444

### DIFF
--- a/public/scripts/editor/js/fc/bramble-menus.js
+++ b/public/scripts/editor/js/fc/bramble-menus.js
@@ -84,19 +84,17 @@ define(function(require) {
     // Add File button and popup menu
     var menu = PopupMenu.create("#filetree-pane-nav-add", "#filetree-pane-nav-add-menu");
 
-    function addFileType(type) {
-      menu.close();
-      bramble.addNewFile(type);
-    }
-
-    function downloadFileToFilesystem(location, localPath, callback) {
+    function downloadFileToFilesystem(location, fileOptions, callback) {
       callback = callback || function noop() {};
 
-      return $.get(location)
-      .then(function(data) {
-        bramble.addNewFileWithContents(localPath, data, function(err) {
+      menu.close();
+
+      $.get(location).then(function(data) {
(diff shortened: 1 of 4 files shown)


humphd:
@flukeout, @hannahkane: can I get you two to review the 3 default files here (html.txt, js.txt, css.txt)?  I want to provide *something* when you make a new file vs. just a blank page, and ideally I want it to be a mix of a) helpful; b) to-the-point; c) providing key info for getting more help.

Let me know if you'd change these to better achieve those goals.

public/scripts/editor/js/fc/bramble-menus.js
```diff
@@ -84,19 +84,18 @@ define(function(require) {
...
-      bramble.addNewFile(type);
-    }
-
-    function downloadFileToFilesystem(location, localPath, callback) {
+    function downloadFileToFilesystem(location, fileOptions, callback) {
       callback = callback || function noop() {};
 
+      menu.close();
+
       return $.get(location)
```
gideonthomas: We don't need to return here right?

gideonthomas:
r=me with confirmation about the default html file contents

hannahkane:
We talked about this here: https://github.com/mozilla/thimble.webmaker.org/issues/541

humphd:
Thanks for reminding me of that ticket, @hannahkane.  I've done an update to the default files to make them more like what was discussed in #541.  Here's what it looks like when you add each one:

screenshot 2015-08-26 13 58 02

screenshot 2015-08-26 13 58 23

screenshot 2015-08-26 13 58 44

NOTE: this is not our default content, it's just what you get when you add a new HTML, CSS, or JS file via the `+` menu.

hannahkane:
copy in comments looks good. I think this addresses feedback we've heard from users that beginning coders don't always know how to reference stylesheets, for example.

PR #929, opened 2015-08-26
Fix #541 - create default empty project

### LINKED ISSUE #541: Figure out starting content for editor.

humphd:
screenshot 2015-08-26 17 02 46

I modified what you had in that bug slightly so that it forces HTML5 rendering vs Quirks Mode (doctype) and will render properly in mobile.

@sedge NOTE: we'll need an env update when this lands.

### DIFF
--- a/default/empty-project/index.html
+++ b/default/empty-project/index.html
@@ -4,6 +4,7 @@
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title>Made with Thimble</title>
+  <link rel="stylesheet" href="style.css">
 </head>
 <body>
 
--- a/default/empty-project/style.css
+++ b/default/empty-project/style.css
@@ -1,0 +1,3 @@
+body {
+  font-family: "Open Sans", sans-serif;
+}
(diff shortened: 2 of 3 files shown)


flukeout:
Can you change ``<b>`` to ``<strong>`` then R+

Though - now that I'm thinking about it - that Open Sans won't really work for someone who doesn't have it installed. We can update it later.

PR #933, opened 2015-08-27
Fix #932 - stop popup menu from throwing on resize, missing click events

### LINKED ISSUE #932: this._underlay.remove() throws on resize after a popup menu was open
Load Thimble, open a popup menu (hit the `+` sign), resize the browser window.

```
Uncaught TypeError: Cannot read property 'remove' of null
PopupMenu.close	@	bramble-popupmenu.js:41
n.event.dispatch	@	index.js:3
n.event.add.r.handle	@	index.js:3
```

humphd:
This fixes two things:
* properly deals with unbinding the resize event so it doesn't fire forever after closing the popup
* deals with click events on the popup menu button that were being missed (saw this in user testing video).


sedge:
r+

PR #936, opened 2015-08-27
Fix #903 - projects list should scroll

### LINKED ISSUE #903: Projects page doesn't scroll vertically
As the list of projects grows, we push the `Create new project` button down and down, and eventually it just drops off the page.  We should have overflow scroll so we don't lose it, and maybe consider moving it to the top?

screenshot 2015-08-20 18 28 01

cc @flukeout

humphd:


flukeout:
So the one problem with this is that it introduces the 'dreaded white bar' and 'tiny scroll' on the editor which was really lame...

image

Can you add this style to just the My Projects page?

humphd:
Closing in favour of https://github.com/mozilla/thimble.webmaker.org/pull/938

PR #938, opened 2015-08-27
Made the projects list scroll when there are lots of items.

### LINKED ISSUE #903: Projects page doesn't scroll vertically

flukeout:
Fixes #903

### DIFF
--- a/public/stylesheets/projects-style.css
+++ b/public/stylesheets/projects-style.css
@@ -4,6 +4,15 @@ body {
   overflow: hidden;
 }
 
+.project-list-wrapper {
+  position: fixed;
+  overflow: scroll;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
+}
+
 #container {
   width: 80%;
   margin: 35px auto;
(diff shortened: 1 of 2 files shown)


public/stylesheets/projects-style.css
```diff
@@ -4,6 +4,15 @@ body {
   overflow: hidden;
 }
 
+.project-list-wrapper {
+  position: fixed;
+  overflow: scroll;
```
humphd: I would use `auto` so you get scrollbars showing at least.

humphd:
One suggestion, r=me otherwise.

public/stylesheets/projects-style.css
flukeout: You don't see scrollbars? I do. They only show up if you have a mouse plugged in though, like other pages.

humphd: No, but it does scroll.  So either way, let's get a fix in.

PR #941, opened 2015-08-27
Don't spit out massive error when preloading editor modules on homepage

humphd:
This overrides the default error handling in require and keeps it from throwing when we preload those Bramble modules, see http://requirejs.org/docs/api.html#errors


sedge:
r+

PR #967, opened 2015-08-28
Add Google Analytics to homepage

humphd:
Related to https://github.com/mozilla/thimble.webmaker.org/issues/749 and https://github.com/MozillaFoundation/plan/issues/542#issuecomment-135829652

We still need the env updated for the right GA account.

### DIFF
--- a/views/homepage/index.html
+++ b/views/homepage/index.html
@@ -9,6 +9,14 @@
     <link rel="stylesheet" href="/resources/stylesheets/userbar.css">
     <link rel="stylesheet" href="/homepage/stylesheets/style.css">
     <title>Thimble by Mozilla - An online code editor for learners &amp; educators.</title>
+
+    {% if GA_ACCOUNT %}
+    <script id="google-analytics-js" src="/homepage/scripts/google-analytics.js"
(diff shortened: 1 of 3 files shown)


views/homepage/index.html
```diff
@@ -9,6 +9,14 @@
     <link rel="stylesheet" href="/resources/stylesheets/userbar.css">
     <link rel="stylesheet" href="/homepage/stylesheets/style.css">
     <title>Thimble by Mozilla - An online code editor for learners &amp; educators.</title>
+
+    {% if GA_ACCOUNT %}
```
sedge: You'll need to add this to the options being rendered in 

https://github.com/mozilla/thimble.webmaker.org/blob/bramble/routes/main/homepage.js#L11-L21

humphd: These are defined globally here https://github.com/mozilla/thimble.webmaker.org/blob/bramble/app.js#L125-L126, so should already exist, no?

sedge: Ah, good call.

sedge:
r+

PR #968, opened 2015-08-28
Fix #950 follow-up - Fix unauthorized error when logging in and out more than once

### LINKED ISSUE #950: Homepage Sign Up & Log In links don't work after Signing Out
Made a little screencap, after signing out, I'm trying to click them. Are we doing some kind of JS on these? Can we not hardcode their destinations?

signout

cc @sedge

gideonthomas:


sedge:
r+ w/ travis

PR #972, opened 2015-08-28
Moved the loading spinner into this codebase so it could overlap the …

flukeout:
* Moved the loading into the Thimble codebase so that it could overlap the entire UI
* Added an error state to the spinner in case Brackets fails to load


public/editor/stylesheets/editor.less
```diff
@@ -634,3 +634,80 @@
 .hide {
   display: none;
 }
+
+
+
+/*
```
humphd: I think you should move this to an inline `<style>...</style>` in the editor HTML so that it renders instantly vs. blocking on the network.

humphd:
This needs a rebase, r=me with that.

PR #973, opened 2015-08-28
Warn users with browsers that aren't known to work with Thimble

humphd:
This is a start, we need proper UI from @flukeout.  I've added the `browser-support-warning` div and css to the `bramble.html` view.

### DIFF
--- a/views/editor/bramble.html
+++ b/views/editor/bramble.html
@@ -1,112 +1,125 @@
 {% block bramble %}
-
-
-<style>
-  /*
-  * Initial loading spinner css, inlined so we don't block loading.
-  * Spinner is based on https://github.com/tobiasahlin/SpinKit (MIT)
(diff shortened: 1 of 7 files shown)


views/editor/bramble.html
```diff
@@ -1,112 +1,120 @@
...
 
     <div id="spinner-container">
         <div class="thimble-spinner"></div>
         <div class="loading-message message">Loading Project</div>
         <div class="error-message message">There was an error loading your Project. <br>Please refresh your browser.</div>
     </div>
+
+    {% if browserNotSupported %}
+    <div id="browser-support-warning">
+      Your Browser isn't full supported yet, but we're working on it.
```
sedge: isn't **fully** supported yet

flukeout:
R+

PR #982, opened 2015-08-31
Fix #976 - Allow only one click on buttons that create a new project

### LINKED ISSUE #976: Clicking 'Start a project from scratch' has odd behaviour
On the homepage, when you click "Start a project from scratch", it seems to take some time to work, and makes you want to click it more than once, which makes you feel like it isn't doing anything.  Also, it needs `cursor: pointer` in the CSS so that it looks clickable.  We shouldn't allow multiple clicks on it, as this likely slows things down.

gideonthomas:


gideonthomas:
r+ from @humphd

PR #1006, opened 2015-08-31
Updated the starter project styling - Fixes 1005

### LINKED ISSUE #1005: Fix ugly shadow element placeholders on Homepage when refreshing.
See the weird grey shape below?

shadows

flukeout:
**Changes**
* Removed "Strip" from the Comic Strip title so that the title fits on one line.
* Added a background color behind the project thumbnail images so that
  * It doesn't look as jarring when the images load because there is already a matching background color
  * The weird element that provides the drop shadow is hidden before the thumbnail images load

Fixes #1005

### DIFF
--- a/public/homepage/stylesheets/style.less
+++ b/public/homepage/stylesheets/style.less
@@ -406,15 +406,15 @@ footer {
       }
 
       &.keep-calm .image {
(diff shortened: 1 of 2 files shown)


gideonthomas:
r+

PR #1016, opened 2015-09-01
Editor heading bar updates - Fixes #1010

### LINKED ISSUE #1010: Editor and preview often seem slightly mis-aligned with their headers.
image

flukeout:
**Changes**
* Set the editor header column widths by pixels to avoid percentage rounding problems
* Changed the toolbar background to grey so there aren't flashes of white when things are resizing

Tested in FF and Chrome and performance is fine!

Fixes #1010


gideonthomas:
r+

PR #1017, opened 2015-09-01
Fix #1011 - Catch JSON.parse errors without throwing and crashing the server

### LINKED ISSUE #1011: [crash] thimble.mozilla.org crash SyntaxError: Unexpected token <
@gideonthomas, can you look at this? From Heroku logs on thimble.mozilla.org

```
PUT /projects/276/files/1404 - - ms - - 
Aug 31 18:34:55 bramble-production app/web.1:  undefined:1 
Aug 31 18:34:55 bramble-production app/web.1:  <!DOCTYPE html> 
Aug 31 18:34:55 bramble-production app/web.1:  ^ 
Aug 31 18:34:55 bramble-production app/web.1:  SyntaxError: Unexpected token < 
Aug 31 18:34:55 bramble-production app/web.1:      at Object.parse (native) 
Aug 31 18:34:55 bramble-production app/web.1:      at IncomingMessage.<anonymous> (/app/routes/files/create-update.js:78:21) 
Aug 31 18:34:55 bramble-production app/web.1:      at IncomingMessage.emit (events.js:129:20) 
Aug 31 18:34:55 bramble-production app/web.1:      at _stream_readable.js:908:16 
Aug 31 18:34:55 bramble-production app/web.1:      at process._tickCallback (node.js:355:11) 
Aug 31 18:34:56 bramble-production heroku/web.1:  Process exited with status 1
```

The app also didn't self restart, which would be something nice to have (cc @jbuck).

gideonthomas:


gideonthomas:
@humphd  whenever you get some time, could you r?

routes/files/create-update.js
```diff
@@ -75,7 +75,13 @@ module.exports = function(config, req, res) {
       });
 
       response.on('end', function() {
-        body = JSON.parse(body);
+        try {
+          body = JSON.parse(body);
+        } catch(e) {
+          console.error("Failed to parse response for sending file with ", e.message, "\n at ", e.stack);
+          res.sendStatus(500);
```
humphd: Sometimes `utils.error(500)` and sometimes this, what's the difference?

routes/utils.js
```diff
@@ -6,8 +6,15 @@ var async = require("async");
 var defaultProject = require("../default");
 
 function createProject(config, user, data, callback) {
-  var project = JSON.parse(JSON.stringify(data));
   var createURL = config.publishURL + "/projects";
+  var project;
+  try {
+    project = JSON.parse(JSON.stringify(data));
```
humphd: Do you really need a clone of this?  Why?

routes/utils.js
```diff
@@ -107,7 +120,14 @@ function updateProject(config, user, data, callback) {
     return;
   }
 
-  var project = JSON.parse(JSON.stringify(data));
+  var project;
+  try {
+   project = JSON.parse(JSON.stringify(data));
```
humphd: Why clone?

routes/utils.js
gideonthomas: because of the `delete project.id` below! We don't want that changing it in the caller

humphd:
I'd say this is r+, with a few questions.  I don't like all the `JSON.parse(JSON.stringify(...))` to clone--we shouldn't need to clone this much.  Also, it would be good if we could standardize on the way we send back errors--we seem to have a real mix of things you do.

But all of this is less important than not-crashing, so r=me.

routes/files/create-update.js
gideonthomas: So the `utils.error` thing seems to be specific only to middleware and I think it wraps the error in webmaker stuff! I think we'll need to figure that out when we refactor the server and remove old code (I believe we won't need to do any of that `utils` stuff then).

humphd: Yeah, that's fine.  Leave it.

routes/utils.js
humphd: OK

PR #1034, opened 2015-09-02
Fix regression - Show 'Untitled Project' when creating a new project as a signed in user

gideonthomas:
Seems like I introduced a regression with 224ddda. When we create a new project/remix a project as an authenticated user, we lose the project title. This PR makes it so that we check to see if we can get the project title from the project root, if not, we should use what the server sends us.


humphd:
r=me

PR #1036, opened 2015-09-02
Build one js file for editor

gideonthomas:


public/editor/scripts/main.js
```diff
@@ -53,7 +45,7 @@ function init(BrambleEditor, Project, SSOOverride, ProjectRenameUtility) {
   });
 }
 
-require(["jquery", "bowser"], function($, bowser) {
+require(["jquery", "bowser", "bramble-editor", "project", "sso-override", "fc/project-rename"], function($, bowser, BrambleEditor, Project, SSOOverride, ProjectRenameUtility) {
```
humphd: This might be OK now, but before we had cases where IE was refusing to load `bramble.js`, which mean that `Bramble` wasn't defined globally, so it would fail and not run our checks on startup.  If you do this, you need to change how we access `Bramble`, probably doing a check on `window.Bramble` and then if it's missing, we need to call the generic error provider here.

What about the idea from https://github.com/mozilla/thimble.webmaker.org/issues/1035#issuecomment-137204974 of putting these modules in the `include` of the build?

humphd: Another thing to do might be to add a check above the `require` call on 48:

```js
if(!window.Bramble) {
  // do error UI here before we call require at all
}

require(["jquery", ....
```

gideonthomas:
So, it works with the `findNestedDependencies` fix but not with the `include: [...]` fix. I also do the `if(!window.Bramble)` check before the first require call

humphd:
r=me, let's test it on staging.

PR #1037, opened 2015-09-02
Fix #1031: warn users of failed filesystem operations in private browsing mode

### LINKED ISSUE #1031: Unable to Remix or Start a new Project without Logging-in or Signing-up on Thimble website
If a user does not Log-in or Sign-up on the [Thimble](https://thimble.mozilla.org/) website, he/she cannot Remix any of the projects provided on the Thimble homepage or "Start a project from scratch".

When I tried to remix all those projects without signing-in first, it went to a new empty page where the thimble icon in the middle of the page just kept loading forever, as shown below:

thimble loading forever

However, when I signed-in to the Thimble website and then tried to Remix those projects, it worked.

thimble working

I hope this bug will be resolved soon.

humphd:
This builds on https://github.com/humphd/brackets/pull/455 and causes us to error when the user is in private browsing mode.  We no longer try to auto recover a failed indexeddb, and that's because we can't do that and warn.  I think the errors we're hearing about in the wild are not corruption, but rather private browsing mode in ff.  This should make things better, since they'll get some info instead of a loading screen that never ends.

### DIFF
--- a/public/editor/scripts/editor/js/bramble-editor.js
+++ b/public/editor/scripts/editor/js/bramble-editor.js
@@ -28,7 +28,6 @@ define(function(require) {
       // Start loading the Bramble editor resources
       Bramble.load("#webmaker-bramble",{
         url: options.editorUrl,
-        autoRecoverFileSystem: true,
         hideUntilReady: true
       });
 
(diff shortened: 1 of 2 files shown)


gideonthomas:
r=me

PR #1066, opened 2015-09-09
Fix for the bug where the File tray header is not appearing in Safari

### LINKED ISSUE #1057: Safari - File tray column header is missing
image

flukeout:
Fixes #1057


gideonthomas:
r+

PR #1071, opened 2015-09-09
Fix #1064 - project TAR file gets corrupted, can't be parsed in IE

### LINKED ISSUE #1064: IE - project TAR file gets corrupted, can't be parsed
Digging into the IE load issues a bit, I've isolated it to our initial project loading logic, and specifically, downloading and unpacking the TAR file we send.  Our process can be summarized as follows:

* on the node.js side, we stream a TAR file to the browser

```js
function sendResponse(res, tarStream) {
  res.type("application/x-tar");
  tarStream
  .on("error", function(err) {
    console.error("Failed to stream tar with: ", err);
  })
  .pipe(res);
}
```

* in the browser we request the TAR file as an `ArrayBuffer` like so:

```js
var xhr = new XMLHttpRequest();
xhr.open("GET", url, true);
xhr.responseType = "arraybuffer";
xhr.onload = function() {
  if(this.status !== 200) {
    return callback(new Error("[Thimble error] unable to get tarball, status was:", this.status));
  }

  unpack(this.response, callback);
};
xhr.send();
```

We notice that the sizes of the files we get vary dramatically between IE (381 bytes) and Chrome (3072 bytes).  Based on reading I'm doing, I think we're sending the file in a way that IE doesn't like (maybe http://stackoverflow.com/questions/1241822/why-is-gzip-compression-with-internet-explorer-not-working/24707644#24707644).

humphd:
This switches our content-type for the project tarball to be something IE can handle (i.e., will decompress).  I've also fixed a load ordering bug in our error handling, where jquery wasn't in scope yet.  Finally, I adjusted the checks for browsers we don't support.  These are temporary, and we can remove them once the UI gets updated.

### DIFF
--- a/public/editor/scripts/main.js
+++ b/public/editor/scripts/main.js
@@ -18,7 +18,18 @@ require.config({
   }
 });
 
-(function() {
+require(["jquery", "bowser"], function($, bowser) {
+  // Temporary check while we finish cross-browser work. We are known
+  // to run well in Firefox, Chrome, Opera, but not Safari or IE.
+  if(bowser.msie || bowser.msedge || bowser.safari) {
+    $("#browser-support-warning").removeClass("hide");
+
+    $(".let-me-in").on("click", function(e) {
+      $("#browser-support-warning").fadeOut();
+      return false;
+    });
+  }
+
   function onError(err) {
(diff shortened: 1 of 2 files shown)


gideonthomas:
r+

PR #1083, opened 2015-09-10
Fix #1060 - deal with browsers not supporting String.prototype.endsWith()

### LINKED ISSUE #883: Local environment: Unable to edit projects when Signed In
Steps to reproduce
* Get your local environment up and running
* Sign In
* Open the thimble editor
  * The starting projects loads correctly
* Refresh
* The editor 'hangs' on the brackets loader

This blocks some of the UI work that I want to do that needs me to be authenticated.

### LINKED ISSUE #1060: Safari - Unable to see editor after signing in locally
After dismissing the browser warning, I'm stuck in this spot after signing in and attempting to load a project...

image

@humphd What about you?

humphd:
See https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith#Browser_compatibility.  I also wonder if this will fix #883

### DIFF
--- a/public/editor/scripts/project/remote.js
+++ b/public/editor/scripts/project/remote.js
@@ -110,7 +110,7 @@ define(function(require) {
     }
 
     function upgradeFile(path, next) {
-      if (path.endsWith('/')) {
+      if (path.match(/\/$/)) {
         return next();
       }
 


sedge:
r+

PR #1094, opened 2015-09-11
Fix #1085 - sanitize text input in Thimble

### LINKED ISSUE #1085: Sanitize text input in Thimble
I realized today that we have a few places where we allow users to enter text, and that they could potentially use this to do XSS attacks if we don't sanitize it, specifically:
* project title
* publish description

I think that's all of them.  Let's clean these values so we don't get sec bugs filed against us.

I'm not sure what MoFo's best practices are for this atm.  cc @jbuck for tips.

humphd:
We have 2 places where we deal with text from a user.  In one case it's always located in a `textarea`.  In the second, we let them type in an `input` and then get the content and put it in a `span`.  I've modified the code to do what @jbuck suggests, and use `textContent`.


gideonthomas:
r=me

PR #1116, opened 2015-09-16
Fix documentation to match new db schema

errietta:
from https://github.com/mozilla/id.webmaker.org/issues/390#issuecomment-140748446

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ To publish locally, you'll need to do the following...
           ( 'test',
             'test',
             '["password", "authorization_code"]'::jsonb,
+            '["code", "token"]'::jsonb,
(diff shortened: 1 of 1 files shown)


sedge:
@errietta Can you confirm that this step is successful with the updated schema?

sedge:
Also, workflow nit, can you rename the commit to reference issue #1116 ? It makes it way easier to track changes if we can jump straight to the issue from the commit history.

Something like: `Fixed #1116 - Fix documentation to ...`

errietta:
@sedge confirmed working, also done :smile:

sedge:
Thank you! r+ :+1:

PR #1123, opened 2015-09-16
Fix #1119 - trigger bramble.saveAll() before navigating away for sign-in

### LINKED ISSUE #1119: Trigger "Save All" before logging in
When a user is working on an anonymous project, we should make sure that everything is saved before we allow them to start the login process.  If we don't, it's possible to lose changes that were very recently made in the editor.  We just need to call `bramble.saveAll(callback)` when they do actions (like login) that will redirect them away (not sure if there are any others).

humphd:
This tries to use Bramble, which may or may not be loaded (homepage vs. editor), and if it does exist, it calls `bramble.saveAll()` before allowing the user to navigate away and sign-in/up.


gideonthomas:
r=me

PR #1126, opened 2015-09-17
Closes #712 - add delete option in navbar

### LINKED ISSUE #712: Add "Delete this Project" item to the dropdown menu.
Users should be able to delete a project while they're looking at it in the editor, so let's give them that option in the dropdown.

Let's add it under the "New Project" action.

When a project is deleted in this fashion, let's show the user the My Projects list.

cc @gideonthomas

errietta:


public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
```diff
@@ -66,6 +66,32 @@ define(function(require) {
...
+          "X-Csrf-Token": $("meta[name='csrf-token']").attr("content")
+        },
+        type: "DELETE",
+        url: "/projects/" + projectId
+      });
+      request.done(function() {
+        if(request.status !== 204) {
+          console.error("[Thimble error] sending delete request for project ", projectId, request.status);
+        } else {
+          window.location.href = '/projects';
```
sedge: This needs to have the querystring attached if it exists.

public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
```diff
@@ -66,6 +66,32 @@ define(function(require) {
...
+        }
+      });
+      request.fail(function(jqXHR, status, err) {
+        console.error(err);
```
sedge: Can you add a more descriptive error prefixed with `[Bramble]`? Something like 

```js
console.error('[Bramble] Delete project request failed', err);
```

sedge:
@errietta A couple of things to fix. Ping me when you've pushed up the changes and I'll test.

errietta:
@sedge PING!

public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
```diff
@@ -66,6 +66,33 @@ define(function(require) {
       queryString = queryString === "" ? "?" + cacheBust : queryString + "&" + cacheBust;
       window.location.href = "/projects/new"  + queryString;
     });
+    $('#delete-project-link').click(function() {
+      // TODO: we can do better than this, but let's at least make it harder to lose data.
```
humphd: Move this comment above line 73

public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
```diff
@@ -66,6 +66,33 @@ define(function(require) {
...
+        },
+        type: "DELETE",
+        url: "/projects/" + projectId
+      });
+      request.done(function() {
+        if(request.status !== 204) {
+          console.error("[Thimble error] sending delete request for project ", projectId, request.status);
+        } else {
+          var queryString = window.location.search;
+          window.location.href = '/projects' + queryString;
```
humphd: @flukeout if the user clicks "Delete Project" while in the editor, where should we take them?  To their project list, or to the homepage?

humphd:
@gideonthomas there's a timing issue here we need to consider.  If files are queued up to be saved, and the user clicks delete, we should probably tell the fs sync code to pause its uploads while the delete step happens, so that it doesn't try to upload a file for a project that (suddenly) doesn't exist.  Thoughts?

gideonthomas:
Hmm, I don't think that should be an issue. We don't really "break" if a file upload fails, so if you try to delete a project when file uploads are queued up there are three possible scenarios:
* All file uploads finish before the request for delete is processed - no issue here
* A file upload request is being processed at the same time as a delete request - This should be handled by the DB since it is a concurrency issue (and should be solved as a general transaction concurrency issue) and worst case is that it will send back a 404 for the file upload request (which we just log).
* A project delete request is processed first before the file upload request is processed - The file upload request will return a 404 which we just log.

The only thing to note is that if there are file uploads queued up and a delete request was processed, the `window.href = ...` code might cause some of the file upload requests to show up as "Client Request Interrupted" errors on Heroku.

humphd:
OK, so let's hear from @flukeout, and then is is r=me.

flukeout:
Scoping now.

flukeout:
@errietta The delete action seems to work great, awesome work.

Could you please update the styling in the dropdown, though?
* The Garbage can icon is darker than the rest of the icons
* The "Delete" project label is slightly too far to the left

image

If you format the markup just like the other list items, it should work fine. I noticed you used the markup from the My Projects page, and that's why there is a difference. 

The garbage can icon is already on the server here:
* https://thimble.mozilla.org/img/icon/garbage-can.svg

Thanks!

errietta:
@flukeout  a little help? Changing the delete button to match the other buttons results into 'Delete Project' not fitting.. 
Seems to fit if we chagne the padding from 12 to 11, should we do that or is there a better trick?
eek

flukeout:
@errietta How about setting the min-width of ``ul.dropdown`` to 185px instead of the current 150px. That will make the menu a little wider and things should fit.

errietta:
@flukeout fixed R?

flukeout:
@errietta Checking now!

flukeout:
@errietta 

Here is something I found, if you have a few projects and delete one, you are redirected to the My Projects list, which is great.

If you only have one project and delete it, the editor refreshes and you are dropped into a new project. Can we redirect to the My Projects page instead? I know it's blank, but it's more consistent, and the user can click the "New Project" button from there. Right now if you delete a blank project, it doesn't seem like it worked because the editor just opens another blank project.

errietta:
@flukeout from what I can tell - it's the 'your projects' page redirecting you when you have 0, not the delete button.

flukeout:
In that case - R+ nice work and thank you for making the small changes I requested :+1:

errietta:
:grinning:

Issue #1122, opened 2015-09-16
Design scalable solution for the curated projects list on the homepage


hannahkane:
We'll be adding more projects to the homepage list. The first one is here: https://d157rqmxrxj6ey.cloudfront.net/mozillalearning/10950/

It doesn't have an accompanying teaching kit, so we'll need a design that allows for that. 

Title: Three Things I <3
Instruction: Learn how to create a multi-page website about things you love!

hannahkane:
cc @chadsansing
related issue: https://github.com/mozilla/learning-networks/issues/241

chadsansing:
By kit, do you mean an activity page? I'm thinking of kits as collections of activities. I can make an activity page for this shortly and will need to, anyway, to finish the Back-to-School Write the Web kit. I can deliver that by EOD tomorrow for review.

hannahkane:
By teaching kit I meant the equivalent of what we offer underneath each of the three activities featured on the Thimble homepage now.

chadsansing:
I can do that.

hannahkane:
Ah, that would be great for consistency, I think.

chadsansing:
It's all happening.

![screen shot 2015-09-16 at 2 27 03 pm](https://cloud.githubusercontent.com/assets/4128600/9914276/0edc634e-5c7f-11e5-9f6d-159940223117.png)

hannahkane:
:)

chadsansing:
@hannahkane, see https://d157rqmxrxj6ey.cloudfront.net/mozillalearning/11288/.

I can't get the Navigation WebLit skill button to display the right makerstrap color. Also, please text in Firefox and Chrome - layout seems broken on my machine outside Safari, but I can't tell why.

ldecoursy:
Looks good! A few small edits:

* First learning objective missing a period
* Intro, 2nd paragraph, 2nd sentence – "play a sorting game" (game should be singular)
* This instruction is unclear: "Explain the rules: After you read each notecard, students who love that thing should all go far to your right, while students who don't love that thing should all go far to the left. Students who are neutral can stay in - or go to - the middle." Maybe we should specify that they should re-sort themselves after each card? 
* There's a stray exclamation point after 3 Things I <3 in the 2nd bullet in the Explore section
* 2nd to last item in the Explore section, I might add "Tell students that today we're going to remix the project to MAKE A SIMPLE WEBSITE THAT SHOWS three things that each of us loves.
[...]

PR #1127, opened 2015-09-17
Added 3 new projects to the homepage

flukeout:
Content for new projects is listed in #1122


flukeout:
Please review, but we need to fix up the content in the published projects before merging.

humphd:
This looks great, r+.

One comment that @LauraReynal brought up...on a small screen, you can encounter the home page like this:

screenshot 2015-09-18 11 30 03

Could you add a second button beside "Start a project from scratch" that says "Remix an existing project" or something, and scrolls you down to the list of 6 projects and teaching kits?  This way people will always know that there is another option.

flukeout:
@humphd Yeah, I noticed that feedback too - I tink it's less about scrolling than the main action not being inviting enough. Let's rethink our homepage approach in #1131

PR #1129, opened 2015-09-18
Run grunt without a path in the postinstall script

### LINKED ISSUE #1128: NPM postinstall-script fails on Windows
Saw [this question on StackOverflow](http://stackoverflow.com/questions/32632589/how-to-install-npm-in-thimble/32644512#32644512) today about the `postinstall` script failing.

The reason (I think?) for the failing is that when using `node_modules/.bin/bower install && node_modules/.bin/grunt requirejs:dist`, as the script, `cmd` cds in to the `node_modules/.bin/` folder before running `bower` and then proceeds to trying to run `node_modules/.bin/node_modules/.bin/grunt`. Ie. cmd runs all the subsequential commands with the same path.

Fix is relatively simple, just removing the path from before `grunt` (or at least it works on Windows with `npm@2.11.3`)

    "postinstall": "node_modules/.bin/bower install && grunt requirejs:dist",

petetnt:
Fixes #1128 on Windows but needs testing on other platforms (linux / osx)

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "webmaker-translation-stats": "0.0.1"
   },
   "scripts": {
-    "postinstall": "node_modules/.bin/bower install && node_modules/.bin/grunt requirejs:dist",
+    "postinstall": "node_modules/.bin/bower install && grunt requirejs:dist",
     "test": "node_modules/.bin/grunt",
     "start": "node app.js"
   }


package.json
```diff
@@ -65,7 +65,7 @@
     "webmaker-translation-stats": "0.0.1"
   },
   "scripts": {
-    "postinstall": "node_modules/.bin/bower install && node_modules/.bin/grunt requirejs:dist",
+    "postinstall": "node_modules/.bin/bower install && grunt requirejs:dist",
```
humphd: Could we do the same thing with `bower` here?

petetnt: @humphd then it would run the globally installed `bower` and `grunt`. Like I mentioned on the issue, they are ran on the same scope.

maybe `cd node_modules/.bin/ && bower install && grunt requirejs` would be clearer?

humphd: This is what I wondered, thank you for looking into it, and thank you for this patch.  Nice work.

petetnt: Thanks @humphd :+1:

Issue #817, opened 2015-08-10
User-menu should open when clicked, not hovered.


flukeout:
Currently, the user dropdown opens when hovered. This can be jarring and unexpected, as all of our other menus / dialogs open when something is clicked. It's also too easy to mouse off of it by accident while trying to pick an option.

![image](https://cloud.githubusercontent.com/assets/25212/9176810/54de15b8-3f42-11e5-9219-884399879c73.png)

**New Behaviour**
* User menu opens when the user name or avatar is clicked
* User menu closes when...
  * A click happens outside of it
  * Something in the menu is clicked

flukeout:
Alternatively, we can have a bigger 'invisible' hover area around the menu so you have a bit of leeway before you've 'hovered off' the menu. Imagine a transparent div with like 15px padding around the menu.

humphd:
Fixed by https://github.com/mozilla/thimble.webmaker.org/pull/1141#issuecomment-142452656

PR #1141, opened 2015-09-21
Fix #1138, #817 - user menu should open when clicked vs hovered.

### LINKED ISSUE #1138: Website menu disapears
A very silly bug, but with FF 40.0.3 win Windows 7 and 8 but not with Chrome the menu at the top right after signing in on https://thimble.mozilla.org/ is slightly below the Hi, USERNAME hover menu item such that the menu (Your projects, New projects, sign out) disappears if you try and select it unless you are very fast or have a touch screen.

humphd:
This re-uses our PopupMenu code to have the usermenu work like the rest of our popup menus (esc key closes, clicking outside it closes).

### DIFF
--- a/public/editor/scripts/editor/js/fc/bramble-keyhandler.js
+++ b/public/editor/scripts/editor/js/fc/bramble-keyhandler.js
@@ -11,7 +11,7 @@ define(function(require) {
 
       e.stopPropagation();
       e.preventDefault();
-      fn();
+      fn(e);
     }
 
     if(typeof elem === "function") {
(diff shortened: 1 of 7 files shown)


flukeout:
@humphd The menu on the homepage doesn't open via click (nor via hover).

flukeout:
@humphd Tested again, now with a different error.

STR
* Load up your branch, restart server
* Go to homepage
* Menu works
* Click New Project
* Go to Editor
* Menu doesn't work
* Refresh the Page - menu works

menu

There isn't really anything in the console that seems related or helpful.

humphd:
This is great, thanks for the details @flukeout .  I'll debug tomorrow.

gideonthomas:
I can't reproduce @flukeout's issue

humphd:
I can't either, so let's have @flukeout do this on vidyo so we can see it and figure out his extra steps.

flukeout:
The ole' wait a day until it goes away strategy worked! Can no longer reproduce! R+

PR #1153, opened 2015-09-28
Fixed header styling when it becomes narrow.

### LINKED ISSUE #1082: Fix up the header UI when the window is narrow.
Currently, we have something like this when the window is small.

image

We can do better!

flukeout:
The UI elements no longer wrap to two lines when the header is squished. Also put a max-width on the project name display so it doesn't get out of hand.

responsive

Fixes #1082


humphd:
Tested in a bunch of browsers, works well. r=me

PR #1159, opened 2015-09-29
Closes #715 Prevent empty titles

### LINKED ISSUE #715: I shouldn't be able to rename a project to a blank string.
Currently, I'm able to save the project title even when the input is blank and I end up with this.

image

Let's at least force this basic validation. If someone tries to Save a blank project title, we just don't let them and the field stays editable until they type in some kind of name.

We should probably also determine if we want any restrictions on the project name
* Allow special characters?
* Max / min length?

I'm not too concerned, as long as we don't allow saving blanks.

cc @gideonthomas

errietta:


public/editor/scripts/editor/js/fc/project-rename.js
```diff
@@ -109,6 +114,12 @@ define(function(require) {
     });
   }
 
+  function validate(context) {
```
sedge: I'd change this to `validateTitle`

errietta:
@flukeout added a commit to disable btton when it's empty :smile: but the problem now is that if you click escape while on the input it empties :open_mouth: any idea what's the deal

flukeout:
@errietta Will take a look!

flukeout:
@errietta Yeah, you're right - but it's not a problem with your code! The production version of thimble behaves in the exact same way, I'll file it as a separate issue.

flukeout:
@errietta In testing your patch, I found that I can still save a 0 length title if I use the enter key while editing.

errietta:
Weird it didn't do thst to me. Ill take a look

On Fri, 2 Oct 2015 16:20 Luke Pacholski <notifications@github.com> wrote:

> @errietta <https://github.com/errietta> In testing your patch, I found
> that I can still save a 0 length title if I use the enter key while editing.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mozilla/thimble.webmaker.org/pull/1159#issuecomment-145059616>
> .
>
-- 
Errietta Kostala
<errietta@errietta.me>

errietta:
@flukeout i can't reproduce this, how do you manage it? I deleted the title and pressed enter but it didn't save?

errietta:
@flukeout i fixed it now, R?

flukeout:
@errietta Checking now!

flukeout:
@errietta Tested this in Firefox and Chrome, and it works as expected! Great work! Let's leave it for @gideonthomas for a code review, but looks good from a UX perspective! :cake:

flukeout:
Awesome work @errietta!

Issue #648, opened 2015-07-02
Add UI for changing line wrap setting 


xmatthewx:
It'd be nice to turn word wrap on/off in the settings (alongside text size and theme).

humphd:
We just need UI to trigger this now, the backend function is all in via https://github.com/humphd/brackets/pull/378

humphd:
I'm morphing this to be about adding the UI.  To whoever implements this, we just need to call `bramble.enableWordWrap()`, `bramble.disableWordWrap()`, and `bramble.getWordWrap()` for state info.

flukeout:
Here is some UI, we already have the little toggle built because we were using it in the Publish dialog to toggle the "Show in public gallery" setting.

![image](https://cloud.githubusercontent.com/assets/25212/8964243/57361c24-35d9-11e5-886f-ec3bdf953e5f.png)

humphd:
@flukeout can I get you to do a PR that adds this select widget, and I'll hook up the backend.

flukeout:
Will do.

humphd:
Fix is happening in https://github.com/mozilla/thimble.webmaker.org/pull/1189

flukeout:
Done

PR #1166, opened 2015-10-06
Added a line wrap toggle UI element in the Editor Options menu.

flukeout:
@humphd Here you go, for #648


humphd:
This is great, thank you.  I've used it in https://github.com/mozilla/thimble.webmaker.org/pull/1189, so I'll close this and we can finish it there.

PR #1176, opened 2015-10-09
Enable thimble to be run on Heroku

cadecairos:
This allows the old thimble to run on heroku by adding bower as a dependency (which it is) and by committing the locales folder.

@jbuck R?


jbuck:
@cadecairos r+

PR #1180, opened 2015-10-15
Fix #1092 - store file sync state info in the project root xattrib

### LINKED ISSUE #1092: Store file upload progress in project root metadata attrib to avoid data loss
When we upload/save files to the server, we do so in response to events from Bramble that a file has changed.  If we are uploading 6 files, and after 3 have been uploaded the browser crashes or user navigates away, we'll lose track of the fact that the other 3 files haven't been sent.  We should first write the queue of file paths to send to the server into the project root's metadata, and remove them one by one after we successfully save them to the server.  We should also inspect this queue on startup to see if there are any files that need to be synced.

humphd:
This is untested, and not 100% complete.  But I think I'd like a review, since I didn't write any of the code I'm replacing.  I'm especially not sure about enabling/disabling the publish dialog.  I'm trying to move to using events instead of callbacks for the sync manager.  I can probably remove a bunch of stuff in that file, but I'm not sure yet.

### DIFF
--- a/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
+++ b/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
@@ -4,7 +4,7 @@ define(function(require) {
   var KeyHandler = require("fc/bramble-keyhandler");
   var BrambleMenus = require("fc/bramble-menus");
   var Underlay = require("fc/bramble-underlay");
-  var Path = Bramble.Filer.Path;
+  var FileSystemSync = require("fc/filesystem-sync");
   var Project = require("project");
 
   var _escKeyHandler;
@@ -15,8 +15,7 @@ define(function(require) {
     $(".preview-pane-nav").width(data.secondPaneWidth);
   }
 
-  function init(bramble, options) {
-    var sync = options.sync;
+  function init(bramble) {
     var publisher;
 
     // *******ON LOAD
@@ -42,12 +41,6 @@ define(function(require) {
         $("#editor-pane-nav-options-menu").hide();
         $("#editor-pane-nav-fileview").hide();
         $(".filetree-pane-nav").css("display", "block");
-      }
-    }
-
-    function showFileState() {
-      if(sync) {
-        $("#navbar-save-indicator").removeClass("hide");
       }
     }
 
@@ -230,7 +223,9 @@ define(function(require) {
       // Listen for ESC to close
       _escKeyHandler = new KeyHandler.ESC(hidePublishDialog);
 
-      publisher.fsync.saveAndSyncAll(function(err) {
+      // Force a Save All in the editor so we get the current state of the editor
+      // written to disk before we sync and publish.
+      FileSystemSync.saveAndSyncAll(function(err) {
         if (err) {
           console.log('[Bramble] Error saving and persisting dirty files:', err);
           return;
@@ -250,8 +245,8 @@ define(function(require) {
       $("#navbar-publish-button").click(showPublishDialog);
       $("#publish-button-cancel").click(hidePublishDialog);
 
-      publisher = new Publisher(options);
-      publisher.init();
+      publisher = new Publisher();
+      publisher.init(bramble);
     } else {
       $("#navbar-publish-button").click(showPublishHelper);
     }
@@ -295,23 +290,6 @@ define(function(require) {
     bramble.on("activeEditorChange", function(data) {
       console.log("thimble side", "activeEditorChange", data);
       setNavFilename(data.filename);
-    });
-
-    // File Change Events
-    bramble.on("fileChange", function(filename) {
-      console.log("thimble side", "fileChange", filename);
-      showFileState();
-    });
-
-    bramble.on("fileDelete", function(filename) {
-      console.log("thimble side", "fileDelete", filename);
-      showFileState();
-    });
-
-    bramble.on("fileRename", function(oldFilename, newFilename) {
-      console.log("thimble side", "fileRename", oldFilename, newFilename);
-      setNavFilename(Path.basename(newFilename));
-      showFileState();
     });
 
     $("#spinner-container").fadeOut();
--- a/public/editor/scripts/editor/js/fc/filesystem-sync.js
+++ b/public/editor/scripts/editor/js/fc/filesystem-sync.js
@@ -1,314 +1,84 @@
+/**
+ * The FileSystemSync module connects the Bramble editor's file system change events
+ * to the PathCache and SyncManager, making sure that all changes to local files
+ * get recorded and eventually sent to the server.
+ */
+
 define(function(require) {
   var $ = require("jquery");
   var Project = require("project");
-  var FileSystemSync = {};
+  var SyncManager = require("fc/sync-manager");
+  var SyncState = require("fc/sync-state");
 
-  function hideFileState(remainingFiles) {
-    if(!remainingFiles) {
-      $("#navbar-save-indicator").addClass("hide");
-    }
-  }
+  var syncManager;
+  var brambleInstance;
 
-  function triggerCallbacks(callbacks, args) {
-    callbacks.forEach(function(callback) {
-      callback.apply(null, args);
-    });
-  }
-
-  function fileChange(csrfToken, fs, path, callback) {
-    var options = {
-      headers: {
-        "X-Csrf-Token": csrfToken
-      },
-      type: "PUT",
-      url: Project.getHost() + "/projects/" + Project.getID() + "/files",
-      cache: false,
-      contentType: false,
-      processData: false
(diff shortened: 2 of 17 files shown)


public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -60,7 +60,7 @@ define(function(require) {
     if(publishUrl) {
       this.updateDialog(publishUrl, true);
     }
-
+/** TODO - not sure about this...
     if(publisher.fsync) {
```
gideonthomas: I think this will become:
```js
// earlier as part of the publisher object
bramble.on("fileChange", function() {
  if(publisher.needsUpdate) return;
  if(!Project.getPublishUrl()) return; // We also need to get/set this everytime we publish/unpublish below
  publisher.handlers.unpublishedChangesPrompt();
  Project.publishNeedsUpdate(true, function(err) {
    if(err) {
      console.error("[Thimble] Failed to set the publishNeedsUpdate flag after a file change: ", err);
      return;
    }
    publisher.needsUpdate = true;
  });
});
```

humphd: Why on `fileChange`?

gideonthomas: So that we can detect when to show the "You need to update your published project" dialog, which should be when a file was changed after a publish occurs.
We also need to do this for `fileRename` and `fileDelete` events

public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
```diff
@@ -230,7 +223,7 @@ define(function(require) {
       // Listen for ESC to close
       _escKeyHandler = new KeyHandler.ESC(hidePublishDialog);
 
-      publisher.fsync.saveAndSyncAll(function(err) {
+      FileSystemSync.saveAndSyncAll(function(err) {
```
gideonthomas: We should add a comment above this explaining that this is only done as a precautionary measure to make sure that every time we make the publish dialog appear, we save unsaved files to the fs.

humphd:
@gideonthomas this is ready for you to test and review.

public/editor/scripts/editor/js/fc/filesystem-sync.js
```diff
@@ -1,314 +1,78 @@
...
-    Bramble.once("ready", function(bramble) {
-      fsync.bramble = bramble;
+    syncManager = SyncManager.init(csrfToken);
 
-      bramble.on("fileChange", fsync.handlers.change);
-      bramble.on("fileDelete", fsync.handlers.del);
-      bramble.on("fileRename", fsync.handlers.rename);
+    // Update the UI with a "Saving..." indicator whenever we sync a file
+    syncManager.on("file-sync-start", function() {
+      $("#navbar-save-indicator").removeClass("hide");
```
gideonthomas: Why don't we do this with the `sync-start` and `complete` events instead?

humphd: I debated this, and we could switch to it.  The reason I didn't is that I thought the "Saving..." indicator might show for a long time, and the user might feel like it's taking a long time and got stuck.  Whereas, if it goes on and off quickly, and more frequently, I think it gives the message that things are happening.

public/editor/scripts/editor/js/fc/filesystem-sync.js
```diff
@@ -1,314 +1,78 @@
 define(function(require) {
```
gideonthomas: Can we add a comment block at the top to explain the purpose of this module? Something indicating that this is a high-level API.

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -19,8 +20,7 @@ define(function(require) {
       .on("click", publish);
   }
 
-  function Publisher(options) {
-    this.fsync = options.sync;
+  function Publisher() {
```
gideonthomas: I don't think we need the `dialog.trackSyncChanges` property anymore; we don't use it anywhere.

public/editor/scripts/editor/js/fc/sync-manager.js
```diff
@@ -0,0 +1,376 @@
...
+  var SYNC_OPERATION_DELETE = require("constants").SYNC_OPERATION_DELETE;
+  var Project = require("project");
+  var PathCache = require("PathCache");
+  var logger = require("logger");
+
+  // SyncManager instance
+  var _instance;
+
+  // Timer for how often to empty the sync queue if not explicitly asked to do so
+  var SYNC_TIMEOUT_MS = 10 * 1000;
```
gideonthomas: I think this should be in `constants.js`

public/editor/scripts/editor/js/fc/sync-manager.js
```diff
@@ -0,0 +1,376 @@
...
+  SyncManager.getInstance = function() {
+    return _instance;
+  };
+
+  // Start the autosync interval.
+  SyncManager.prototype.start = function() {
+    var self = this;
+
+    // Setup automatic syncing
+    setInterval(function() {
```
gideonthomas: `setInterval(this.sync.bind(this)), SYNC_TIMEOUT_MS);` ?

public/editor/scripts/editor/js/fc/sync-manager.js
```diff
@@ -0,0 +1,376 @@
...
+      self.sync();
+    }, SYNC_TIMEOUT_MS);
+  };
+
+  SyncManager.prototype.emitProgressEvent = function() {
+    var self = this;
```
gideonthomas: We don't really need self here do we?

public/editor/scripts/editor/js/fc/sync-manager.js
```diff
@@ -0,0 +1,376 @@
...
+      request.done(function() {
+        if(request.status !== 201 && request.status !== 200) {
+          return callback(new Error("[Thimble] unable to persist `" + path + "`. Server responded with status " + request.status));
+        }
+
+        var data = request.responseJSON;
+        Project.setFileID(path, data.id, callback);
+      });
+      request.fail(function(jqXHR, status, err) {
+        console.error("[Thimble] unable to persist the file to the server. Error was:", err);
```
gideonthomas: Should we use `logger` here as well?

public/editor/scripts/editor/js/fc/sync-manager.js
```diff
@@ -0,0 +1,376 @@
...
+
+      // If there are no pending paths to sync, we're done.
+      if(self.pendingCount === 0) {
+        logger("SyncManager", "no pending sync operations, stopping syncing.");
+        self.emitProgressEvent();
+        self.setSyncing(false);
+        return;
+      }
+
+      // Sync first path in the pending list
```
gideonthomas: It won't really be the "first" path will it?

public/editor/scripts/logger.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Simple console logging. To enable logging, use ?logging=1
+ */
+
+define(function() {
+
+  function log(module, data) {
```
gideonthomas: This is very cool!

public/editor/scripts/project/project.js
```diff
@@ -153,19 +180,30 @@ define(function(require) {
...
+        // If we have paths to sync in the SyncQueue, ignore this error and keep going
+        if(!Object.keys(syncQueue.pending).length) {
+          return callback(err);
+        }
+      }
+
+      // If there are cached paths that need to be updated, queue those now
+      if(pathUpdatesCache && pathUpdatesCache.length) {
+        pathUpdatesCache.forEach(function(path) {
+          queueFileUpdate(path);
```
gideonthomas: What about deletes?

public/editor/scripts/editor/js/fc/filesystem-sync.js
gideonthomas: Ok sounds good

gideonthomas:
Filed https://github.com/mozilla/thimble.webmaker.org/issues/1197

gideonthomas:
I cannot seem to rename the project!
I get:
```
Uncaught TypeError: Cannot read property 'enter' of undefined
```
Also, it seems to hit the `/projects/<project_id>/rename` route 4 times for one rename

gideonthomas:
All the actual sync stuff works very nicely! This patch is terrific!

gideonthomas:
A few coding comments and the one rename issue, r=me with that!

humphd:
I fixed the things you pointed out.  I'm not sure what you're hitting with that rename type error--I can't reproduce.  Can you give me some STR, a screencast gif, etc?

PR #1190, opened 2015-10-21
Added Email Us link to the footer.

flukeout:
We don't have this requirement in an issue, but by God, do we need it.


flukeout:
@hannahkane This is how it looks after adding the email link...

image

hannahkane:
+1

PR #1199, opened 2015-10-27
Fix #1198 - Renaming the project triggers multiple requests

### LINKED ISSUE #1198: Renaming the project triggers multiple requests
If you rename the project and click "Save", we get a console error "Uncaught TypeError: Cannot read property 'enter' of undefined" about 5-10 times and the request to `/projects/<project_id>/rename` also happens the same number of times

humphd:
This fixes a few bugs:
* we were adding/removing (and mostly adding) the click handler over and over when the any key handler ran.  Instead of doing this, I just add it once, then ignore it if the button is disabled when it fires
* I've fixed the esc key handling so that it restores the title as it was before editing began. This includes both the 0 length string case, and also all other cases.
* to fix the above, I had to fix a bug in the input field, where it wouldn't set the proper attribute on both an `input` and `span`.

Testing this now only gives me 1 request for both the `enter` case and also the `click` of the Save button.  I also find the esc key does what I expect.

@gideonthomas r?

### DIFF
--- a/public/editor/scripts/editor/js/fc/bramble-input-field.js
+++ b/public/editor/scripts/editor/js/fc/bramble-input-field.js
@@ -93,12 +93,13 @@ define(function(require) {
   // Similar to JQuery's `.val()` method
   InputField.prototype.val = function() {
     // Extract the value from the <input> or <span> depending on input state
+    var method = this.mode === InputField.Modes.STATIC ? "text" : "val";
+
     if(arguments.length === 0) {
-      return this._element[this.mode === InputField.Modes.STATIC ? "text" : "val"]();
+      return this._element[method]();
     }
 
-    // Store the value as harmless text content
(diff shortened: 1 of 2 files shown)


public/editor/scripts/editor/js/fc/bramble-input-field.js
```diff
@@ -93,12 +93,14 @@ define(function(require) {
...
   InputField.prototype.val = function() {
     // Extract the value from the <input> or <span> depending on input state
+    var method = this.mode === InputField.Modes.STATIC ? "text" : "val";
+
     if(arguments.length === 0) {
-      return this._element[this.mode === InputField.Modes.STATIC ? "text" : "val"]();
+      return this._element[method]();
     }
 
     // Store the value as harmless text content
```
gideonthomas: We should remove this comment since it is not "text" anymore

PR #1208, opened 2015-11-02
Fix #1196 - add UI for enabling/disabling automatic JS execution in the preview

### LINKED ISSUE #1196: Prevent this page from creating additional dialogues not present in Chrome
Hi there, 

I sometimes create activities that include javascript alert pop ups. Example at https://thimble.mozilla.org/user/mchesterman/35277

In the previous thimble both FF and Chrome had a box users could tick  "Prevent this page from creating additional dialogues not present in Chrome". 

FF is still working fine. 

fff_thimb

But this message never appears for Chrome which make remixing the activity impossible. 

Thanks for the great tool!

chr_thimb

humphd:
This adds UI to toggle automatic JS execution in the preview on/off.  We've had support for this in Bramble since the beginning, but never bothered to do UI.  I think it's pretty useful when you're doing large JS projects and want to block re-running JS on every refresh.

I'm not sure of the best wording: "Allow JavaScript" vs. "Autorun JavaScript" vs. ...?

NOTE: this is a runtime option vs. something we save in your overall preferences, since it's pretty specific to a particular project, and I think liable to confuse users if we disable it by default on startup.

r? @gideonthomas, @flukeout

### DIFF
--- a/public/editor/scripts/editor/js/fc/bramble-menus.js
+++ b/public/editor/scripts/editor/js/fc/bramble-menus.js
@@ -43,6 +43,24 @@ define(function(require) {
     setWordWrapUI(bramble.getWordWrap());
 
 
+    // Enable/Disable JavaScript in Preview
+    $("#allow-scripts-toggle").click(function() {
+      // Toggle current value
+      var $allowScriptsToggle = $("#allow-scripts-toggle");
+      var toggle = !($allowScriptsToggle.hasClass("switch-enabled"));
+
+      if(toggle) {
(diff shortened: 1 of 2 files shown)


gideonthomas:
@humphd I'm having issues running JS now though:

humphd:
That doesn't look related to my change.  Can you find me on irc so we can debug?

humphd:
Let's do a prod push in the next few days so this stuff gets in and tested ahead of MozFest.

mickfuzz:
Amazing! Great work

PR #1209, opened 2015-11-02
Hide published details bar when on mobile phones without enough room

humphd:
@flukeout, see if this is good enough for now, so we don't crowd out the page content on a mobile phone.

### DIFF
--- a/public/resources/remix/index.js
+++ b/public/resources/remix/index.js
@@ -17,7 +17,8 @@ var detailsBarHtml =
 function customizeScroll($) {
   var bodyEl = $("body");
   var detailsBar = $(".remix-details-bar");
-  var detailsBarHeight = 64;
+  // If the remix details bar is not visible, don't add padding
+  var detailsBarHeight = detailsBar.is(":visible") ? detailsBar.height() : 0;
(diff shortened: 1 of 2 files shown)


flukeout:
@humphd Looking now

flukeout:
@humphd I don't have a good way to test this out, as I'm not able to publish locally at the moment. Were you able to run this on a device?

One comment is that we're adding a ``padding-top: 64px`` to the body element to accomodate the details bar when it's visible...

image

So, if we're going to hide the bar, we should remove that padding too.

humphd:
@flukeout take a look at this last commit, see if this fixes it.

@gideonthomas can you test this locally please?  The goal is not to have the remix stuff show at all on a phone.

flukeout:
@humphd seems good - can we push it to staging for testing?

humphd:
Landed on staging, let's try this out tomorrow.

flukeout:
:+1:

humphd:
It's not quite right.  Here's a follow-up: https://github.com/mozilla/thimble.webmaker.org/pull/1212

humphd:
This has all been backed out now.

PR #1213, opened 2015-11-03
Disable publish route using env param

cadecairos:


cadecairos:
@jbuck does this seem sane?

jbuck:
@cadecairos r+

PR #1214, opened 2015-11-03
Remove frontend/* skeleton, which was never fully implemented.

humphd:
I've torn out the `frontend/*` stuff that we added in the hope of redoing our front end code to use a better set of tooling--it never happened, and it's not likely to happen, so I want it gone in order to not confuse people working on this code.


Gruntfile.js
```diff
@@ -130,91 +111,17 @@ module.exports = function( grunt ) {
           src: [
             "public/editor/**/*.js",
             "public/homepage/**/*.js",
+            "public/resources/remix/index.js",
```
gideonthomas: Should we be jshinting the resources at all? Theoretically, any web resource can be put into this folder (which may have lint errors) so should we bother linting any of them at all?

gideonthomas:
one question, otherwise r+

Gruntfile.js
humphd: I think this one is worth doing, since it's the basis of our injected remix header, and if it has obvious JS issues, it would be nice to get those fixed in development before we ship.  I agree with not having everything in `public/resources/*` being linted--we can add more to this list when we add things we write.

PR #1216, opened 2015-11-04
Show anonymous local projects in My Projects list

humphd:
This is the start of improvements to how we show you your past projects.  I want to make it possible to get back to local anonymous projects you might have, but for which you don't have the uuid and URL saved.

The only thing I have left to do is to figure out how to show this list when you aren't logged in.  I'll either do that in this patch, or do it in a follow-up.  For now, this is all working.

### DIFF
--- a/public/editor/scripts/open-project.js
+++ b/public/editor/scripts/open-project.js
@@ -1,6 +1,9 @@
 require.config({
+  baseUrl: "/editor/scripts/editor/js",
   paths: {
-    "jquery": "/bower/jquery/index"
+    "jquery": "/bower/jquery/index",
+    "logger": "../../logger",
+    "constants": "../../constants"
   },
   shim: {
     "jquery": {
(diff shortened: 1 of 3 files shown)


gideonthomas:
Are we sure we want to allow this? I'm not sure how the classroom stuff works but here's a scenario:
A student works on a project anonymously in a classroom (saves the url somewhere) and leaves. The next class comes in, another student signs in and sees that there is an anonymous project there so they meddle with it/upgrade it/whatever. Previous student comes back for another class, tries to open up the url that he saved and finds that the project is not there anymore.

I'm not sure if we should put effort into exposing anonymous projects because it is a bit of a rabbit hole where we would have to deal with data loss and other issues like the one I mentioned above. I think the whole point of the uuid is to "identify the project" without owning it. If you lost that identity, I think it is fair to say that you lose that project. If we want to allow you to recover from that, we should probably do that at an admin level (there can be a bunch of different solutions for that including automation - something like show a "Recovery" dialog box asking for other information like the name of the project, how many files where in it, etc. and then finding it in Indexeddb based on this info and revealing the  project with the uuid; possibly a warning to remind the user to note down the uuid).

public/editor/scripts/open-project.js
```diff
@@ -9,86 +12,169 @@ require.config({
...
+          return;
+        }
 
-  Array.prototype.forEach.call(projects, function(project) {
-    var projectSelector = "#" + project.getAttribute("id");
-    var lastEdited = project.getAttribute("data-project-date_updated");
+        var table$ = $("#project-list-anonymous");
+        entries.forEach(function(entry) {
+          table$.append(createTableRow(entry));
+          $("#anonymous-project-" + entry).on("click", function() {
```
gideonthomas: where is the prefix `anonymous-project-` being applied in the html?

humphd:
I think your concerns are valid.  I wonder how to balance the fact that we currently just lose your work, by not offering to show you your previous URLs, with not over-sharing projects between multiple users on the same browser (in a school or other shared computer)?

One thought: what if we remembered anonymous project UUIDs in the browser's session storage?  On the home page, in addition to the "Start from scratch..." link we could show you your previous anonymous project link so you could get back to it.

We'll have to discuss with @flukeout and @hannahkane.

public/editor/scripts/open-project.js
humphd: I think I changed this, good catch.

flukeout:
Wanted to follow up on this based on our chats last week. We'd agreed that the primary focus of this feature would be to help you not lose work you were previously working on anonymously. So, if you worked on something, closed your browser, and ended up in the editor again the next day, it might give you an option to restore your last project...

* Hey, last time you were here you worked on some stuff, [click here] to restore it.

Or maybe beside the [Create a New Project] button on the homepage, you could see a link to 'restore your last session' or something better-worded.

This doesn't really address the issues that @gideonthomas mentioned though and also exposes anonymous work to other people at that same computer. If we go down this road, should we make it easy to 'delete' your anonymous work too so that the next person at the computer won't see it?

hannahkane:
Trying to catch up. Do we indicate to the user that their anonymous project will be accessible to anyone using the same login/browser? The fact that we say "Log in or Sign up to Save" implies that the work is *not* being saved anywhere, so users might be unpleasantly surprised.

humphd:
@hannahkane I think you're right, this entire bug is mostly about setting (and keeping) expectations.  We don't really explain the fact that your stuff is saved (locally to *this* browser) when not logged in, vs. saved to the server but not published, vs. published.  I think trying to untangle this is complex, because most people won't understand that their browser can store data for a page which isn't really accessible to them (i.e., it's not saved to their drive).

Here's why I'm making an issue of this, and looking to improve it.  I'm teaching my daughters to code using Thimble.  They don't have accounts, and just work anonymously, so are typical <13 year old users.  When they make something, it's saved on their Chromebooks automatically by us.  **But!** we don't tell you how to get back to it at all, so if you close the tab and go back to Thimble, you ostensibly lose your work.  I watched it happen, so I know it's a real user flow :(  "Dad! Where did all my work go!?"

I just want some way to allow a user to get back to the thing(s) they've made if they weren't logged in.

ryanwarsaw:
@gideonthomas @humphd What's the status on this? Do we want to clean this up and merge?

gideonthomas:
I don't know if this is ever going to be a feature we implement. @humphd what do you think about closing this?

humphd:
Yup, let's toss it.

PR #1244, opened 2015-11-11
Fix #1241 - Stop Thimble (client) from DDoS'ing Thimble/Publish servers

### LINKED ISSUE #1241: Stop Thimble (client) DDoS'ing Thimble (server) and Publish (server)
Spin-off from #1225, specifically related to timeouts and autosave.  Our current autosave works like this:

* Bramble autosaves (locally to filer fs/indexeddb) file changes every 5 seconds, or whenever the user switches files
* Thimble listens for file update/delete/rename events from Bramble, and queues them up
* Every 10s, we have a timer that runs and looks at the queue to see if there are updates that need to go to the server.
* If there are updates in the queue, it tries doing one.  If it fails, it re-queues it and goes on to the next one
* If the user clicks Publish, it drains the queue of pending events to be synced (i.e., it forces the above process to run).

I think this can be fixed in a few ways:
* Don't send updates to the server as often as we do.  I bet we could change the 10s update window to something like 30s.
* Don't run the next operation in the queue right away if the last one failed.  We should implement (exponential backoff with jitter)[http://www.awsarchitectureblog.com/2015/03/backoff.html] so that we aren't slamming the server with update requests over and over.  There's some python code we could use as a basis here: https://github.com/awslabs/aws-arch-backoff-simulator/blob/master/src/backoff_simulator.py
* Let the user know when things are failing a lot if it matters.  For example, when they are waiting to publish, they will care that it's taking forever, but when they are just editing, it doesn't matter so much (all changes are already saved locally).

cc @Pomax, @gideonthomas.

humphd:
This patch greatly improves our performance when the network or server go down, or when the server is under heavy load.  When an update or delete network operation fails, this adds some delay before retrying it.  Previously, we just tried again right away, and if the server was under load, we'd just do that over and over in a tight loop forever, which will kill the server with enough clients all doing it at once.

This also fixes a bug where we incorrectly thought we'd completed a sync if the server went down (i.e., `net::ERR_CONNECTION_REFUSED` vs. a `500` response).  This would mean that we could lose track of files that need to get synced in the event of a crashed server.

This also provides a bit of UI indication to the user when saving fails, letting them know and that it is being retried.

There are a few things I'd like advice on, though:
* are my numbers right?  I'm doing autosync intervals of 15ms, and a failure backoff delay of 1-20s.  For the most part a user is unaware of server syncing, and everything gets saved to the local db almost instantly (i.e., we aren't in a rush to get it on the server).
* I was going to do something fancy with [exponential backoff with jitter](http://www.awsarchitectureblog.com/2015/03/backoff.html), but in the end just chose a random Nms between 1000ms and 20,000ms.  My thinking is that this spreads out the load across 20s time slices with many simultaneous clients.  Is there something smarter I could do?

If you want to test this, just kill the publish (will cause a `500`) or thimble (will cause a connection refused error) servers while it's running or in the middle of a sync. With this patch, the servers can come and go without any data loss or increased load from the client.  Running it with `?logging=1` will give lots of info on the console. 

r? @gideonthomas, and also @Pomax and @jbuck on the higher level thinking + numbers I chose.

### DIFF
--- a/public/editor/scripts/editor/js/fc/backoff.js
+++ b/public/editor/scripts/editor/js/fc/backoff.js
@@ -1,0 +1,27 @@
+/**
+ * Determine a backoff delay (ms) to wait before running the next
+ * network operation. Inspired by https://github.com/awslabs/aws-arch-backoff-simulator/blob/master/src/backoff_simulator.py
+ */
+define(function(require) {
+  var BACKOFF_BASE_MS = require("constants").BACKOFF_BASE_MS;
+  var BACKOFF_MAX_DELAY_MS = require("constants").BACKOFF_MAX_DELAY_MS;
+
+  // Returns a random integer between min (included) and max (included)
+  // Using Math.round() will give you a non-uniform distribution!
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random
+  function _getRandomIntInclusive(min, max) {
+    return Math.floor(Math.random() * (max - min + 1)) + min;
+  }
+
+  function Backoff() {
+    this.iter = 0;
+  }
+
+  Backoff.prototype.next = function() {
+    this.iter++;
+    var v = Math.min(BACKOFF_BASE_MS * Math.pow(2, this.iter), BACKOFF_MAX_DELAY_MS);
+    return _getRandomIntInclusive(0, v);
+  };
+
+  return Backoff;
+});
--- a/public/editor/scripts/editor/js/fc/sync-manager.js
+++ b/public/editor/scripts/editor/js/fc/sync-manager.js
@@ -35,9 +35,12 @@ define(function(require) {
   var EventEmitter = require("EventEmitter");
   var SYNC_OPERATION_UPDATE = require("constants").SYNC_OPERATION_UPDATE;
   var SYNC_OPERATION_DELETE = require("constants").SYNC_OPERATION_DELETE;
-  var SYNC_TIMEOUT_MS = require("constants").SYNC_TIMEOUT_MS;
+  var AUTOSYNC_INTERVAL_MS = require("constants").AUTOSYNC_INTERVAL_MS;
+  var AJAX_DEFAULT_DELAY_MS = require("constants").AJAX_DEFAULT_DELAY_MS;
+  var AJAX_DEFAULT_TIMEOUT_MS = require("constants").AJAX_DEFAULT_TIMEOUT_MS;
   var Project = require("project");
   var PathCache = require("PathCache");
+  var Backoff = require("fc/backoff");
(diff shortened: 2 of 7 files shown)


jbuck:
I'd say that increasing the autosync interval higher (every minute?) would be a good idea, and I think that using exponential back with jitter is probably a better idea than a uniformly distributed value between 1-20s.

public/editor/scripts/editor/js/fc/sync-manager.js
```diff
@@ -247,11 +256,18 @@ define(function(require) {
...
             self.setPendingCount(syncQueue);
 
-            // If there are more files to sync, run the next one
+            // If there are more files to sync, run the next one.
             if(self.getPendingCount() > 0) {
-              self.runNextOperation();
+              self.emitProgressEvent();
+
+              // If the last operation errored, try to apply a backoff delay.
+              delay = ajaxError ? self.backoff.next() : 10;
```
gideonthomas: The 10 should probably be documented (or made a constant)

public/editor/scripts/editor/js/fc/backoff.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * Determine a backoff delay (ms) to wait before running the next
+ * network operation. Inspired by https://github.com/awslabs/aws-arch-backoff-simulator/blob/master/src/backoff_simulator.py
+ */
+define(function(require) {
+  var BACKOFF_BASE_MS = require("constants").BACKOFF_BASE_MS;
+  var BACKOFF_MAX_DELAY_MS = require("constants").BACKOFF_MAX_DELAY_MS;
+
+  // Returns a random integer between min (included) and max (included)
+  // Using Math.round() will give you a non-uniform distribution!
```
gideonthomas: `Math.random()`

public/editor/scripts/editor/js/fc/sync-manager.js
```diff
@@ -247,11 +253,18 @@ define(function(require) {
...
             self.setPendingCount(syncQueue);
 
-            // If there are more files to sync, run the next one
+            // If there are more files to sync, run the next one.
             if(self.getPendingCount() > 0) {
-              self.runNextOperation();
+              self.emitProgressEvent();
+
+              // If the last operation errored, apply a backoff delay.
+              delay = (self.backoff && self.backoff.next()) || 10;
```
gideonthomas: The 10 needs to be documented or made into a constant

gideonthomas:
r+

public/editor/scripts/editor/js/fc/backoff.js
humphd: Copied word-for-word from MDN.  It's saying "don't use Math.round()"

PR #1247, opened 2015-11-13
Fix #975 - Thimble doesn't remember my theme settings

### LINKED ISSUE #975: Thimble doesn't remember my theme settings.
Steps to reproduce
* Switch to dark theme
* Refresh the browser
* Reverts to white

theme-forget

humphd:
This simplifies the code that figures out what we should do on startup, and forces bramble into the right state no matter what.  I can't reproduce the theme problem anymore with this patch.

r? @gideonthomas


gideonthomas:
Uhhh the code looks like it's doing the same thing as before O.o
But if it works, r+

humphd:
It's not quite the same, since this *always* sets the theme.  I think there might have been a timing bug on the brackets side, or in the bramble client api, that sometimes missed it being set.

Issue #1224, opened 2015-11-07
Thimble breaks (timeout) on low-powered computers & sub-3G network speeds


sedge:
On the network side, testing in chrome revealed the threshold for success to be average 3G or greater speeds. We're also seeing older computers failing to load everything in time, causing a timeout.

Here are specs from one computer that failed to load despite adequate network speeds:

**OS:** Windows 7
**RAM:** 2GB
**CPU:** Intel Atom N450 @ 1.66ghz
**GPU:** Inten Graphics Media Accelerator 3150

gideonthomas:
@sedge is this the homepage or the editor too?

sedge:
Editor for sure. @Pomax You were playing with Thimble while travelling right? Did you find the homepage breaking?

humphd:
There's a couple of things we can do here:
* bump our `waitSeconds` value for requirejs, so it doesn't timeout so soon on slow connections.
* the images we load are bigger than they need to be.  We could add some build-time image compression for the images on the homepage.  I tried compressing one of the project PNG files, and it went from 194K to 49K, a 75% size saving.
* evaluate our use of the YouTube video on that page: it costs us ~50% of the total download size to load the embedded player (even if you never watch the video, the player's JS is pretty big, and gets loaded first).

humphd:
I've got PRs up against Brackets and Thimble for the require timeouts portion of this.

PR #1252, opened 2015-11-17
Don't crash when loading over low-bandwidth networks, part of #1224

humphd:
If not instructed to do otherwise, requirejs will timeout module loading after 7s.  Based on similar bugs and fixes in other code, I've set our Thimble timeout to be 2 mins.  I'm going to do a similar fix in Bramble.


gideonthomas:
r+

PR #1253, opened 2015-11-17
Fix #1164 - Add GA tracking to Published projects

### LINKED ISSUE #1164: Add Google Analytics tracking to Published projects.
We don't have any tracking baked into Published projects but I think we should.

I was thinking that we could set up a separate "site" under Google Analytics so we could track these independently from the Thimble homepage and editor.

@adamlofting Thoughts?

We could inject the tracking code along with the details bar and remix button for published projects.

humphd:
This adds GA tracking for published projects via the remix header injection.  Since we load this file without any processing (not templated), I just embedded the tracking id directly in the code.  I've also cleaned-up this file a bit to isolate things from the global scope.

### DIFF
--- a/public/resources/remix/index.js
+++ b/public/resources/remix/index.js
@@ -1,116 +1,139 @@
-// Taken from https://github.com/flukeout/thimble-home
+/**
+ * Thimble Remix and Analytics injection, automatically
+ * added to published projects.
+ */
+(function(document, head){
 
-var detailsBarHtml =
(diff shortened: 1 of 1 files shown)


flukeout:
r+ look good - but I can't test it, haven't been able to publish locally for quite some time.

humphd:
Landed on staging.  @flukeout, can you try loading http://bramble-staging.s3-website-us-east-1.amazonaws.com/humphd/73/ and see if you get GA stats?

PR #1263, opened 2015-11-19
Fix #1210 - Get thimble working on node 4.2.x, remove dead code, modules from old Thimble

### LINKED ISSUE #1210: Get thimble working on node 4.2.x (LTS)
It appears to be failing right now while compiling the Nan package for an unknown dependency.

humphd:
Thimble runs for me on Node LTS (4.2.1) now.  There might be more we can trim, but this is a good start.


env.dist
```diff
@@ -40,39 +40,11 @@ export SESSION_SECRET="dummy secret value"
 # Super Cookie Domain - leave empty for localhost
 # COOKIE_DOMAIN=webmaker.org
 
-# database settings
```
sedge: I've... I've wanted this for so long

package.json
```diff
@@ -10,7 +10,6 @@
     "aws-sdk": "^2.2.4",
```
sedge: Line 6 of this file should be updated to match the new node version

sedge:
r+ with one nit.

humphd:
@sedge I altered `package.json` and `.travis.yml`, if you want to confirm you're OK with what I did.

fyi @gideonthomas, @flukeout--this change will require you to have node.js version 4.2.x or higher.

sedge:
r=me

PR #1271, opened 2015-11-23
Hides the bar markup until the page is fully loaded to avoid a FOUC

### LINKED ISSUE #752: Brief flash of unstyled content on the Details Bar when a published project is viewed.
**What's going on**
When you first open a published Thimble project, the title of the project briefly appears in large text above the project details. Once the Javascript & CSS for the Remix button loads, this content then disappears. It is later visible, as expected, when you hover over the remix Button, like this...

image

**Note** - By "Details Bar", I'm referring to the white box (visible above) when a user hovers over the Remix button.

**Steps to reproduce**
- Load up a published project like http://bramble-staging.s3-website-us-east-1.amazonaws.com/flukeout/77/
- Refresh it over and over and keep an eye on the top part of the page
- Sometimes you see the unstyled markup (like the project title) of the details bar for a bit before it turns into the styled header

**Solutions**
I think what we can do is just apply an inline style like `display: none` to markup that wraps the details bar. Once the JS and CSS loads, we can remove this inline style.

Alternatively, we can inline all of the styles in a `<style>` tag so we don't have to wait for any external resources to load.

I'm not sure if these approaches will work, but they're worth trying out.

flukeout:
What is a FOUC you ask? It's a "Flash of unstyled content", and it is very ugly.

Fixes #752 

@humphd R?

### DIFF
--- a/public/resources/remix/index.js
+++ b/public/resources/remix/index.js
@@ -9,6 +9,7 @@
   function setupBar($){
     var isTouchDevice = 'ontouchstart' in document.documentElement;
     var detailsBar = $(".details-bar");
+    detailsBar.css("display", "block");
 
     if(isTouchDevice) {
       detailsBar.addClass("touch-mode");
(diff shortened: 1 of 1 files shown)


humphd:
Dumping the `style` attribute vs. messing with the `display` style seems a bit unorthodox to me, but I think it's OK here. r=me if you think this approach is clean enough.

flukeout:
@humphd Made it change the display style instead of nuking the style attribute. R?

humphd:
Nice.  r+

PR #1282, opened 2015-12-01
Fix #1276 - implement UX for Bramble preview inspector

### LINKED ISSUE #1276: Design UX for "inspector" feature in Bramble
I've got a [PR up for Bramble](https://github.com/humphd/brackets/pull/506) that adds bi-directional inspecting (hovering in editor shows DOM element in preview; hovering in preview shows associated code).   It needs some thinking around proper UX:

* I think we need a button to turn it on/off, so it doesn't autoscroll your editor just by moving your mouse into the preview.  Maybe modelling it on the Inspector in dev tools
* I've got a "click-to-pin" feature, so if you hover over something, and click, it will stop jumping around to other parts of the code.  I'm not sure if I should still highlight other things, but not scroll you, etc.  Again, the Inspector probably gives us some models to follow.
* I use a flash effect on the background of the lines.  I need some better CSS to apply to the lines.
* I wonder if we should change the mouse pointer when you're in "inspector" mode?

humphd:
This is a WIP patch to enable what's needed on the Thimble side for #1276, and it relies on https://github.com/humphd/brackets/pull/506 for the backend changes to Bramble (make sure you `grunt build-browser` if you pull that).

I can't get the exact line highlighting around a block of lines as desired in the UX designs in #1276, since CodeMirror won't allow it (each line is an individual element, with no option to form a `<div>` around multiple).  I have also *not* implemented scroll-on-highlight in the editor.

I have also not done final icon treatment, pending your choices.

highlight-line-2

Highlights look like this in the dark theme:

screenshot 2015-11-30 21 31 23

And this in the light theme:

screenshot 2015-11-30 21 32 01


flukeout:
@humphd Getting a strange glitch when moving the mouse slowly over an element with the inspector on. Looks like it's removing and reapplying the highlight over and over.

inspector-glitch

flukeout:
Icon does a glitchy flash when you enable the inspector fairly often. I'm guessing this might be because the :active state of the button (when mouse is down) is blue, so when you release the button it turns white again, then blue again when it gets a signal from brackets that the feature is on? That sound about right @humphd ?

icon-flash

flukeout:
The inspector doesn't highlight the ``body`` element when you're not hovering over a child of the body.

flukeout:
When the inspector is initially turned on and used, it shows a "pointer" cursor when hovering over elements. If you turn on the inspector, keep it on and then refresh, it stops using the pointer...

cursor-change

flukeout:
If it's turned on, we need to turn off the inspector (and hide the icon) when someone switches to fullscreen preview mode, otherwise this shows up...

image

flukeout:
In addition, I'd like to make some cosmetic changes, but we can do these in another PR - just noting down here for now:
* When an element is hovered with the inspector, don't fade out the blue highlight
  * Only fade it out if you've selected the element and hovered off
* Change the border to 2px and dashed
* Explore adding a 'brighter' highlight on the dark theme that fades out over time so it's more noticeable immediately since people will be using their peripheral vision to keep track of it.

humphd:
Thanks for the review, @flukeout, you found some good stuff.  I've fixed everything you listed above, some of it here in this PR, some in the Brackets PR.  If you want to re-test with both, let me know where you think this is at.

I agree that those other changes should go in a new bug/PR, as they will require extensive changes to Bracket's highlighting code in LiveDevelopment.

flukeout:
Checking this out now

flukeout:
@humphd This is still happening when I fire up the latest...

inspector-glitch

When you're hovered over a highlighted element and move the mouse, it re-highlights it one every mouse move. Does that happen on your end too?

humphd:
You have the brackets PR fixes too?

On Dec 3, 2015 12:55 PM, Luke Pacholski <notifications@github.com> wrote:

@humphd<https://github.com/humphd> This is still happening when I fire up the latest...

[inspector-glitch]<https://cloud.githubusercontent.com/assets/25212/11543692/f66cd144-98f2-11e5-96d7-6abb165f68ad.gif>

When you're hovered over a highlighted element and move the mouse, it re-highlights it one every mouse move. Does that happen on your end too?

flukeout:
I do yeah, I'll try again - do I have to ``grunt build-browser`` or something like that?

humphd:
Odd, this is fixed for sure, but in brackets

On Dec 3, 2015 1:00 PM, Luke Pacholski <notifications@github.com> wrote:

I do yeah, I'll try again - do I have to grunt build-browser or something like that?

flukeout:
Oh yeah there we go - I think I screwed up getting your latest Brackets changes.

flukeout:
**Checklist **
* [x] Item highlight no longer resets when moving the curos
* [x] Highlighted selection doesn't reset when you mouse onto the editor
* [x] Inspector icon gone and inspector disabled when Fullscreen preview enabled
* [x] No icon flicker when enabling the inspector

Nice work @humphd - tested in FF / Chrome. R+

PR #1291, opened 2015-12-03
Tutorial styling updates

### LINKED ISSUE #1267: Handle overflow in Tutorial's pre.code
Thimble's Tutorial.html includes &lt;pre class="code" &gt; as a preset style. Long lines break the box. Fix with ```.code { overflow:auto; }``` or similar.

flukeout:
**Some fixes**
* Images need a special class ``.border`` now to display in the Polaroid style
  * This makes it so you can inline regular images
  * These images have a max width of 100% now too, so they won't cropped by the edge of the tutorial
* Changed the cat image to a small screenshot of the Thimble site
 * I was hotlinking some random image from some kind of veterinarian org, oops!
* Code that is too long for it's ``<pre class='code'>`` container now scrolls instead of continuing past the bounds

Fixes #1267 
Fixes # 1232


flukeout:
@gideonthomas R?

humphd:
This looks good.  One thing I notice is that there's a vertical alignment issue with the numbered lists:

screenshot 2015-12-02 20 24 44

That may not be new, but FYI, in case you caused it and/or want to fix it here.

r=me otherwise.

flukeout:
Thanks @humphd Updating now.

flukeout:
And merged!

PR #1354, opened 2016-01-25
Localize the homepage

gideonthomas:


gideonthomas:
@alicoding a suggestion on reviewing this - go by the commits, it's probably easier that way

locale/en_US/messages.properties
```diff
@@ -0,0 +1,99 @@
...
+##########################################
+##### Strings visible through the UI #####
+##########################################
+##########################################
+
+##############
+## Homepage ##
+##############
+
+thimbleHeading=Thimble by Mozilla - An online code editor for learners &amp; educators.
```
alicoding: I think you can just do `&` in here...

locale/en_US/messages.properties
```diff
@@ -0,0 +1,99 @@
...
+remixGoal8=Embedding media in webpages
+remixGoal9=Basic coding ideas
+remixGoal10=Editing JavaScript content
+remixGoal11=Creating an interactive button<br /> with JavaScript
+remixGoal12=Searching for openly licensed<br /> images online
+remixGoal13=Editing for specific word choice
+# "Keep Calm Poster"
+keepCalmPoster=Keep Calm Poster
+keepCalmPosterDescription=Change the design of this poster to make it your own.
+keepCalmPosterTeachingKit=Keep Calm {{ gettext("teachingKit") }}
```
alicoding: I don't think I like this format at all :/ I think we can probably cache `{{ gettext(...) }}` in the template and just have the variable here in instead. So, it will just be like `keepCalmPosterTeachingKit=Keep Calm {{teachingKit}}`

package.json
```diff
@@ -39,19 +40,22 @@
...
     "nunjucks": "1.3.4",
+    "properties-parser": "^0.3.1",
+    "q-io": "^1.13.2",
     "request": "2.21.0",
     "serve-favicon": "^2.3.0",
     "tar-stream": "^1.2.1",
     "time-grunt": "^1.2.1",
     "uuid": "^2.0.1",
-    "webmaker-i18n": "0.3.25",
+    "webmaker-i18n": "^0.3.27",
```
alicoding: Let's remove the `^` here

package.json
```diff
@@ -39,19 +40,22 @@
     "multer": "^1.0.1",
     "newrelic": "1.5.5",
     "nunjucks": "1.3.4",
+    "properties-parser": "^0.3.1",
```
alicoding: Probably these too

locale/en_US/messages.properties
gideonthomas: how (or where) would I cache it?

package.json
gideonthomas: you sure we don't want to get patch updates for this?

locale/en_US/messages.properties
gideonthomas: unless you're saying pre-render it and send it in the options object of `res.render('index.html`, options)`?

package.json
alicoding: We might want to use greenkeeper to get updates, so removing the caret is better here.

locale/en_US/messages.properties
alicoding: https://mozilla.github.io/nunjucks/templating.html#set

gideonthomas: ah I see, interesting

alicoding:
R+ with that fix

PR #1369, opened 2016-01-28
Code cleanup - Remove unused variables and dependencies

gideonthomas:
@jbuck r?


app.js
```diff
@@ -28,7 +14,6 @@ var express = require('express'),
 var favicon = require('serve-favicon'),
     cookieSession = require('cookie-session'),
     cookieParser = require('cookie-parser'),
-    logger = require('morgan'),
```
jbuck: Are there any alternative logging mechanisms that you're using? I think it's a pretty bad idea to get rid of *all* console logs.

app.js
```diff
@@ -1,17 +1,3 @@
-process.env.NEW_RELIC_BROWSER_MONITOR_ENABLE = false;
-
-// New Relic Server monitoring support
-var newrelic;
-if ( process.env.NEW_RELIC_ENABLED ) {
```
jbuck: Could you also remove all of the environment variables that aren't used anymore in `env.dist`

public/resources/remix/style.less
```diff
@@ -6,7 +6,7 @@
 
 @font-face {
   font-family: "Fira Medium";
-  src: ~"url(FiraSansOT-Regular.otf)";
```
jbuck: Why is the font-family Fira Medium when you're loading the Regular variant...

jbuck:
@gideonthomas r-

app.js
gideonthomas: the logs should still show up with `console.log` no? we just aren't formatting it in anyway?

public/resources/remix/style.less
gideonthomas: Not sure, @flukeout might be in a better position to answer this

app.js
jbuck: https://github.com/mozilla/thimble.webmaker.org/pull/1369/files#diff-0364f57fbff2fabbe941ed20c328ef1aL100

public/resources/remix/style.less
flukeout: @gideonthomas I'm not exactly certain about why I wrote it this way. It might be that "Medium" as a relative description of font-weight within the UI just maps better to the "Regular" version of the font. "Fira Medium" is just a cosmetic label anyway - so I wouldn't worry about this discrepancy.

jbuck:
@gideonthomas r+

PR #1371, opened 2016-01-29
Fix regression caused by bd6a6f - Add project list to build step

gideonthomas:


jbuck:
@gideonthomas is there any way you could add a Travis CI test that ensures that the build works correctly?

jbuck:
After an in-person discussion we found that we'd need to get in-browser tests running.

@gideonthomas r+

PR #1380, opened 2016-02-03
Client L10N Tools

gideonthomas:
This patch does a couple of things:
* Uses a build script to create separate client files for each locale
* Adds a watch to update the built files when the source code changes

Note: This patch is more of a tool that will be used in the next patch rather than showing any obvious changes in the application.

Note 2: This should be landed after https://github.com/mozilla/thimble.webmaker.org/pull/1377

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -132,8 +132,8 @@ module.exports = function( grunt ) {
     }
   });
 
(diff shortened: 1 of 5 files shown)


alicoding:
Is this patch actually ready for the review?

alicoding:
r+

PR #1383, opened 2016-02-05
Automatically update built localized client files on src change

gideonthomas:


scripts/watch-client.js
```diff
@@ -0,0 +1,58 @@
...
+  error() {
+    console.log.apply(console, arguments ? log.args.call(arguments, "red") : null);
+  },
+  info() {
+    console.log.apply(console, arguments ? log.args.call(arguments, "green") : null);
+  }
+};
+
+function deleteFile(relativeFilePath) {
+  let deleteLocalizedFile = localeDirList.map(localeDir => fs.remove(path.join(localeDir, relativeFilePath)));
```
alicoding: Why assign this now?

alicoding:
r+ with nits

scripts/watch-client.js
gideonthomas: @cadecairos noted that it looked cleaner this way instead of stuffing all the code into one line.

alicoding: Alright

PR #1386, opened 2016-02-08
Setup localized client scripts

gideonthomas:
This patch configures the scripts on the client to find the correct files using requirejs based on the path preceded by the locale and can be found in either the `dist` or `client` directory depending on the environment.

This also adds the localization build process to the `postinstall` script so that it works on heroku.

To test this patch, simply run `npm install` and then run thimble as you normally would.


locales/en-US/messages.properties
```diff
@@ -123,6 +123,7 @@ prjListDeleteProjectBtn=Delete
 ## Shared ##
 ############
 
+locale=en_US
```
alicoding: Why in this file?

public/homepage/scripts/main.js
```diff
@@ -33,10 +33,10 @@ require.config({
 function preloadBramble($) {
   var brambleHost = $("meta[name='bramble-host']").attr("content");
   [
-    brambleHost + "/dist/styles/brackets.min.css",
-    brambleHost + "/dist/bramble.js",
-    brambleHost + "/dist/main.js",
-    brambleHost + "/dist/thirdparty/thirdparty.min.js"
+    brambleHost + "dist/styles/brackets.min.css",
```
alicoding: Is the local enforce here? What happened if no locale present in the url

gideonthomas: this is specifically for brackets which has no locale in the url (until we localize it).

locales/en-US/messages.properties
gideonthomas: coz it needs to be rendered into the requirejs config like out here https://github.com/mozilla/thimble.webmaker.org/pull/1386/files#diff-270f212d36b8a2447fd4f97c986a0b06R4
I was thinking about other ways to do this but I can't come up with one :( (suggestions welcome)
But I think this will be a bit clearer when I add the client strings into this file

locales/en-US/messages.properties
```diff
@@ -4,6 +4,12 @@
 ##########################################
 ##########################################
 
+######################
+## Locale directory ##
+######################
+# This needs to be in every locale file and the value should be the locale with an underscore replacing any occurences of a hyphen
+locale=en_US
```
alicoding: why it needs to be in the locale file? We don't want people to translate this.

alicoding:
r+

PR #1387, opened 2016-02-08
Extract strings for l10n from client scripts

gideonthomas:


gideonthomas:
This is the final step before we add Pontoon to this!

alicoding:
r+

PR #1389, opened 2016-02-09
Enable Pontoon for l10n

gideonthomas:
cc @alicoding, @simonwex

### DIFF
--- a/server/security.js
+++ b/server/security.js
@@ -6,15 +6,23 @@ let csurf = require("csurf");
 
 const defaultTrustedDomains = {
   default: [ "'self'" ],
-  connection: [ "'self'" ],
+  connection: [
+                "'self'",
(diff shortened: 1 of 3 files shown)


server/security.js
```diff
@@ -9,7 +9,11 @@ const defaultTrustedDomains = {
   connection: [ "'self'" ],
   frame: [
           "'self'",
-          "https://docs.google.com"],
+          "https://docs.google.com"
+         ],
+  iframe: [
```
alicoding: I think you want to allow pontoon on `script` part as well. Pretty much you want to make sure Pontoon can do everything like it's an internal server? https://github.com/mozilla/donate.mozilla.org/blob/master/server.js#L282-L293

alicoding:
r+

PR #1392, opened 2016-02-09
Handle best language

alicoding:


alicoding:
travis is failing, but I will let you handle this :P

PR #1401, opened 2016-02-11
Expose language list to templates

gideonthomas:
@flukeout try rebasing against this branch or just use this code.

I think the nunjucks code for this will look something like:
```
{% for langCode, obj in languages %}
   <a href="/{{ langCode }}/whatever">{{ obj.nativeName }}</a>
{% endfor %}
```

Let me know if that works


flukeout:
OK, will give it a try in a bit.

Issue #1402, opened 2016-02-12
Rename this repo to thimble.mozilla.org


simonwex:
Pretty simple ^

simonwex:
Hey @gideonthomas, what effects would doing this rename have on things like how we build and deploy this?

gideonthomas:
Should be fairly simple. Changes that I know need to be made:
* Update `.travis.yml`
* Update `bower.json` and `package.json`
* Update `README.md`

I don't know if anything on heroku needs to be updated - cc @cadecairos

gideonthomas:
Fixed in c46c3fe46e374b22ac802b978f65146b97538b34, 108acfbf11883d75ae4112c6dedf7758d0010989 and 3a0c59c7943f41a0e66acf48bf91ec97c7cdd8d7

PR #1404, opened 2016-02-13
Feature/rename repo

raphamorim:
Ref: #1402


gideonthomas:
nice work @raphamorim :)

PR #1413, opened 2016-02-17
Wrap homepage footer elements in spans for discoverability by Pontoon

gideonthomas:
Pontoon needs to be able to discover these text nodes and so we need to wrap them in container elements...I chose spans.

### DIFF
--- a/views/homepage/index.html
+++ b/views/homepage/index.html
@@ -239,12 +239,12 @@
     <footer>
       <div class="content-wrapper">
         <ul class="social-links">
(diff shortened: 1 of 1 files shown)


Pomax:
that sounds odd - does Pontoon not recognise <a> textcontent? If not, that sounds like a bug that should be fixed in pontoon rather than worked around here (`<li><a><span>` becomes pretty unwieldy and can have unintended sideeffects from span CSS)

gideonthomas:
@Pomax here's what was explained to me on IRC:
"pontoon searches for all text nodes and then makes parent DOM nodes contenteditable if their content matches any of the strings in source file; and since footer element text nodes are preceded with spans (icons), pontoon doesn’t find a match"

gideonthomas:
@Pomax either way, I think we should go with this change now and we can remove it later

alicoding:
r+

PR #1424, opened 2016-02-24
Fix #1388 - localize the injected project remix header bar and script

### LINKED ISSUE #1388: Localize the remix script
The remix script that is injected into published projects needs to be localized!

This is going to be tricky since all we can rely on is the locale info given to us by the browser. Accordingly, we need to make sure we send the right correctly localized file to the user-agent.

I think simple server-side rendering of the localized content into the js file should be good enough to accomplish this. But, we need to make sure that the routes to access this remix script go through the node-webmaker-i18n middleware first. We also need to make sure that it does not go through the i18n-force-locale-in-url middleware.

A suggestion is to create a custom dynamic route that will render the file using nunjucks.

humphd:
I don't yet have a fully functional dev env locally, so I'd appreciate an end-to-end test of this with the publish server.  However, what I have run seems to be working well.

Let me know too if I've done what you want with the string properties.

### DIFF
--- a/server/routes/projects/remix-bar.js
+++ b/server/routes/projects/remix-bar.js
@@ -1,0 +1,29 @@
+"use strict";
+
+let moment = require("moment");
+
+module.exports = (config, req, res) => {
+  let updated;
+  let options;
+  let params = req.query;
+  let lang = (req.localeInfo && req.localeInfo.lang) ? req.localeInfo.lang : "en-US";
+
+  // Get a localized "time ago" string, dealing with bogus date strings without crashing.
+  moment.locale(lang);
+  try {
+    updated = moment(params.updated).fromNow();
+  } catch(e) {
+    updated = "";
+  }
+
+  options = {
+    id: params.id,
+    host: params.host,
+    title: params.title,
+    author: params.author,
+    updated: updated,
+    lang: lang
+  };
+
+  res.render("project-remix-bar.html", options);
+};
--- a/views/project-remix-bar.html
+++ b/views/project-remix-bar.html
@@ -1,0 +1,22 @@
+<!-- Project Remix bar -->
+<div lang="{{ lang }}" class="details-bar collapsed cleanslate" style="display: none !important">
+  <a class="thimble-logo" title="{{ gettext("thimbleByMozilla") }}" href="https://thimble.mozilla.org">
+    <span class="icon"></span>
+  </a>
+
+  <h1 class="remix-project-title">{{ title | safe }}</h1>
+  <div class="remix-project-meta">
+    <span class="remix-project-author">{{ gettext("createdBy") | instantiate }} - {{ updated }}</span>
+  </div>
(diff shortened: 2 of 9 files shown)


humphd:
@gideonthomas r?

server/routes/projects/remix-bar.js
```diff
@@ -0,0 +1,27 @@
+var moment = require("moment");
```
gideonthomas: can you convert this file to ES6? We're trying to slowly convert Thimble towards it. For this file, I think the only change that needs to be made is converting all the `var` to `let`. Oh and adding "use strict" at the beginning.

gideonthomas:
@humphd can you rebase onto `master` once https://github.com/mozilla/thimble.mozilla.org/pull/1426 lands. It will help with the following fixes that I suggest.

server/routes/projects/remix-bar.js
```diff
@@ -0,0 +1,27 @@
...
+
+module.exports = function(config, req, res) {
+  var params = req.params;
```
gideonthomas: This should be `req.query` not `req.params`

gideonthomas:
You'll need to add an `env` variable called `PUBLISHED_PROJECTS_HOSTNAME` and set it to `http://localhost:8001` for now. Then in `server/index.js`, you'll want to change [this line](https://github.com/mozilla/thimble.mozilla.org/blob/master/server/index.js#L107) to:
```js
localize(server, Object.assign(env.get("L10N"), {
   excludeLocaleInUrl: [ "/projects/remix-bar" ]
}));
```

You'll also need to add that env variable to the `server/config.js` `module.exports` and in your route setup (where you do `app.get("/projects/remix-bar", ...)`), you want to make a call to `middleware.enableCORS(the env var)`.

After that I think this works, but I'll test again.

gideonthomas:
@humphd my patch has landed so you can go ahead and rebase and try the fixes I suggested.

humphd:
@gideonthomas I've made the fixes and rebased.  r? again please.

views/project-remix-bar.html
```diff
@@ -0,0 +1,22 @@
...
+  </a>
+
+  <h1 class="remix-project-title">{{ title | safe }}</h1>
+  <div class="remix-project-meta">
+    <span class="remix-project-author">{{ gettext("createdBy") | instantiate }} - {{ updated }}</span>
+  </div>
+  <div class="details-bar-remix-button-wrapper">
+    <a title="{{ gettext("remixWithThimble") }}"
+       class="details-bar-remix-button"
+       href="{{ host }}/projects/{{ id }}/remix">{{ gettext("remix") }}
```
gideonthomas: I would add `lang` to the url too. So `href="{{ host }}/{{ lang }}/projects/{{ id }}/remix"`

gideonthomas:
one small nit, r=me

PR #1426, opened 2016-02-25
Enable CORS api and allow routes to not be localized

gideonthomas:
This also converts the `middleware.js` file to ES6.


alicoding:
LGTM r+

PR #1428, opened 2016-02-29
Fix #1415 - Combine anonymous publish text to enable context based l10n

### LINKED ISSUE #1415: Combine the "Sign in or Create an Account to" and publish button
Since context matters here, we should not allow the translation of the word "Publish" to be used in "Sign in or Create an account to" publishButton.

The publish button markup should be a part of the "Sign in..." string so that translators can localize the word "publish" accordingly with respect to the rest of that phrase. The "Publish" word itself should also have a separate translation for other uses.

Note: An announcement re: this change should be posted in the google group once it has been implemented.

gideonthomas:
@alicoding r?

### DIFF
--- a/locales/en-US/messages.properties
+++ b/locales/en-US/messages.properties
@@ -87,7 +87,7 @@ fileSavingFailedIndicator=Saving failed, retrying...
 help=Help
 deleteProjectLink=Delete Project
 exportProjectLink=Export (.zip)
-loginPromptEditor=<a href="#" title="Sign in to Thimble" id="login-link" class="navbar-login">Sign in</a> or <a href="#" title="Create an account" id="signup-link" class="navbar-login">Create an account</a> to Save and
+loginPromptEditor=<a title="Sign in to Thimble" id="login-link" class="navbar-login">Sign in</a> or <a title="Create an account" id="signup-link" class="navbar-login">Create an account</a> to Save and {{ publishBtnOpenTag }}Publish{{ publishBtnCloseTag }}
 
 # Navigation bar
 navFiles=FILES
--- a/views/editor/userbar.html
+++ b/views/editor/userbar.html
@@ -20,6 +20,7 @@
     </div>
 
     <div id="navbar-right" class="container-flex">
+      {% set publishBtnTitle = gettext("publishBtnTitle") %}
 
       <div class="nav-chunk">
         <div title="{{ gettext("helpIconTitle") }}" id="navbar-help" class="nav-link">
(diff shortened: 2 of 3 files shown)


locales/en-US/messages.properties
```diff
@@ -87,7 +87,7 @@ fileSavingFailedIndicator=Saving failed, retrying...
 help=Help
 deleteProjectLink=Delete Project
 exportProjectLink=Export (.zip)
-loginPromptEditor=<a href="#" title="Sign in to Thimble" id="login-link" class="navbar-login">Sign in</a> or <a href="#" title="Create an account" id="signup-link" class="navbar-login">Create an account</a> to Save and
+loginPromptEditor=<a title="Sign in to Thimble" id="login-link" class="navbar-login">Sign in</a> or <a title="Create an account" id="signup-link" class="navbar-login">Create an account</a> to Save and <publishBtn>Publish</publishBtn>
```
alicoding: What is that <publishBtn>?

views/editor/userbar.html
```diff
@@ -93,19 +94,28 @@
             </div>
           </div>
         {% else %}
+          {% set publishBtnOpenTag = '<span title="{{ publishBtnTitle }}" id="navbar-publish-button"><img src="/img/icon/publish.svg" class="icon-publish">' %}
+          {% set publishBtnCloseTag = '</span>' %}
           <div id="navbar-anonymous">
-            {{ gettext("loginPromptEditor") | safe }}
+            {{
+              gettext("loginPromptEditor") |
```
alicoding: Can't you just use `instantiate` instead of doing a replace?

locales/en-US/messages.properties
gideonthomas: Because apparently the publish button context is different when it is part of the "Sign in or Create an account to Save and" string, I had to add the publish button markup to this string. But because it was a lot of markup, and it will be hard for localizers to figure it out, I use `<publishBtn>` tags instead which I replace with the actual markup with nunjucks

views/editor/userbar.html
gideonthomas: see: https://github.com/mozilla/thimble.mozilla.org/pull/1428#discussion_r54361355

gideonthomas: oh I see what you're saying...I'll change it to that.

alicoding: This is weird, but okay...

alicoding:
r+ or tag me again if you want me to look at it.

gideonthomas:
@alicoding I changed it so that it doesn't use `replace` and instead just `instantiate`. Would you be able to take a second look?

alicoding:
if you've tested and this is working then r+

PR #1439, opened 2016-03-07
Add script to trigger build on Brackets for new strings

gideonthomas:
This step will run every time a Travis build occurs and will check to see if the commit being built is a string change to the Brackets' string files. If it is, it triggers a rebuild on Brackets (which would pull the string changes and update the repo - writing code for that now).

One thing I'm not so sure of is the API token that I use in the travis request. I generated that locally using `travis login && travis token` according to https://docs.travis-ci.com/user/triggering-builds but I don't know if that is only for a session (would need to be regenerated every time) or should work for this case.

### DIFF
--- a/scripts/localize-brackets.sh
+++ b/scripts/localize-brackets.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+set -e # exit with nonzero exit code if anything fails
+
+SKIP="Skipping localization on Brackets:"
+
+# request body for the Travis API call to rebuild the latest commit in Brackets
+body='{
+  "request": {
+    "branch": "master",
+    "config": {
+      "env": {
+        "global": ["UPDATE_STRINGS=true"]
+      },
+      "script": "bash ./scripts/pull-new-strings.sh"
+    }
+  }
+}'
+
+# Exit the script if the build is for a pull request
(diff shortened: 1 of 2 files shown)


humphd:
I'm going to get @alicoding to review this one.

scripts/localize-brackets.sh
```diff
@@ -0,0 +1,56 @@
...
+
+      # request body for the Travis API call to rebuild the latest commit in Brackets
+      body="{ \
+        \"request\": { \
+          \"branch\": \"master\", \
+          \"config\": { \
+            \"env\": { \
+              \"global\": [\"UPDATE_STRINGS=true\", \"CHANGED_LOCALES=\"$CHANGED_LOCALES\"\"] \
+            }, \
+            \"script\": \"bash ./scripts/pull-new-strings.sh\" \
```
jbuck: Where does the source for this script live?

gideonthomas: It *will* live in brackets, I'm writing that now.

jbuck:
This PR seems fine to me, but I don't think it can be reviewed or landed until the Brackets pull-new-strings script is in the Brackets repo

gideonthomas:
cool thanks @jbuck! I'll ping you when I finish writing that!

PR #1440, opened 2016-03-07
Add Brackets' strings

gideonthomas:
This adds the strings that were meant to be in the Brackets repo so that Pontoon can access it.

### DIFF
--- a/locales/en-US/editor.properties
+++ b/locales/en-US/editor.properties
@@ -1,0 +1,162 @@
+############
+## ERRORS ##
+############
+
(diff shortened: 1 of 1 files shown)


humphd:
r=me

alicoding:
yup, r+

gideonthomas:
Thanks for the r guys! I'm gonna wait until we land all the nuts and bolts for this to be relevant so that people on Pontoon don't see a bunch of strings and start localizing it without it ever reaching brackets

PR #1450, opened 2016-03-14
Fix other CSP errors

gideonthomas:
Fixes a bunch of CSP errors that pop up so that the console is a bit cleaner


alicoding:
LGTM

PR #1466, opened 2016-04-12
Added UI for toggling the Auto-Refresh feature

### LINKED ISSUE #1417: add a way to turn off auto-previewing
If you have a large document, the auto-preview reloading actually starts to interfere with your typing in files, contantly interrupting the input stream, leading to far more typing errors than is otherwise the case, as well as contant unnecessary preview reset.

A feature similar to JSbin's "autorun with JS: [x]" checkbox would be very much appreciated.

flukeout:
**Changes**
* Moved the Refresh icon to the left side of the Preview bar 
* Added an animation when the Refresh icon is clicked
* Added a checkbox next to it that toggles the new auto-update feature in Brackets

Should look like this...
reload

@gideonthomas R?

Fixes #1417


locales/fr/messages.properties
```diff
@@ -0,0 +1,223 @@
+##########################################
```
humphd: Why is this whole file in this PR?

humphd:
One stray file got committed, otherwise looks good.

I'll let @gideonthomas have the final say. r=me

locales/fr/messages.properties
flukeout: Oh oops - had no idea what this was for.

gideonthomas: Ha I'm pretty sure this was because we were working with this file manually mid-way through the l10n process.
You can git checkout this file!

public/editor/stylesheets/editor.less
```diff
@@ -587,6 +587,125 @@ body {
   }
 }
 
+#preview-pane-nav-refresh {
```
gideonthomas: Empty rule?

locales/en-US/messages.properties
```diff
@@ -197,6 +197,8 @@ navAllowJSTitle=Toggle running JavaScript code in the preview
 navPreviewTitle=View the Preview
 navViewTutorialTitle=View the Project Tutorial
 navDOMElementInspectorIconTitle=Toggle Element Inspector
+navToggleAutoUpdateTitle=Toggle Preview Auto-refresh
+navToggleAutoUpdateLabel=Auto
```
gideonthomas: Let's move this specific label here: https://github.com/mozilla/thimble.mozilla.org/pull/1466/files#diff-aef074d5302a863440aa549e7a2fd981R107

gideonthomas: Also, I wonder if "Auto" is the right word for this. Might be good to get feedback from non-devs about this once it lands.

gideonthomas:
r+ with a few nits.

locales/en-US/messages.properties
flukeout: I'm not sure what you mean @gideonthomas - move it where?

locales/fr/messages.properties
flukeout: Not sure what that means - should I delete it in my branch and re commit the PR?

PR #1476, opened 2016-04-27
Corrected a typo, added a missing word

Lehmax:


ryanwarsaw:
@gideonthomas I'm assuming we can't merge this because localization stuff has to come through Pontoon correct?

Pomax:
Yes, please have a look at https://github.com/mozilla/thimble.mozilla.org/wiki/Localization on how we do Localization for Thimble. Translation is done through Pontoon, and translates are synced automatically from pontoon to this repository, so we cannot take PRs for this: they will get automatically overwritten with the "wrong" string later because Pontoon won't have the fixed version

PR #1479, opened 2016-05-04
Fix #1474 - Disable Fullscreen Button is not working

### LINKED ISSUE #1474: Disable Fullscreen Button is not working
Hi, I'm not sure if this is my browser issue but I cannot return to the code after I have enabled the full screen.
thimble workspace one
thimble workspace two

humphd:
I tried playing with the sizing, and doubling things fixes it.  The problem is the extra width of the newly added `auto` checkbox for refresh.

screenshot 2016-05-04 11 30 05

r? @flukeout, @gideonthomas

### DIFF
--- a/public/editor/stylesheets/editor.less
+++ b/public/editor/stylesheets/editor.less
@@ -956,8 +956,12 @@ body {
     top: 0;
   }
 
-  .filetree-pane-nav, .editor-pane-nav, .nav  {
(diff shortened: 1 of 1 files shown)


gideonthomas:
uh r+ from me

flukeout:
We also need to adjust the width of the wrapper element, the one with the darker border that is getting cut off in your screenshot. 188px will do it...

image

Otherwise, good!

flukeout:
Yep, working as expected now!

PR #1520, opened 2016-07-21
Support concurrent server processes using Throng

### LINKED ISSUE #1519: Implement Node Clustering
We're running this app on a 2X dyno, which gives it 1GiB of memory to play with, but the app only uses a single node process, resulting in it using a fraction of what's available.

We should implement the [throng module](https://www.npmjs.com/package/throng), which we use in several other places to boost concurrency.

[Here's an example of how we're using it in donate](https://github.com/mozilla/donate.mozilla.org/blob/a5c5d045914ae7bace5fdf55751bb682473bffb8/start.js)

Essentially, we export the server object from `server.js` and teach throng how to start it, enabling concurrent server processes.

cadecairos:
Fixes #1519 

Now you can specify `WEB_CONCURRENCY` to get multiple server processes! hooray!

@gideonthomas R??


app.js
```diff
@@ -1 +1,23 @@
-require("./server");
+'use strict';
+
+const throng = require('throng');
+const thimble = require('./server');
+const workers = process.env.WEB_CONCURRENCY || 1;
```
gideonthomas: can we add these new env vars to the `env.dist` file?

app.js
```diff
@@ -1 +1,23 @@
...
+    server.close(function() {
+      process.exit(0);
+    });
+  };
+
+  process.on('SIGINT', shutdown);
+  process.on('SIGTERM', shutdown);
+};
+
+throng(workers, start);
```
gideonthomas: new line

gideonthomas:
r+ with nits

Issue #1506, opened 2016-07-08
Design a generic error page


gideonthomas:
Honestly, this as a temporary solution is fine so that we can stop the template errors that are crashing the server like:
```
» 8 Jul 2016 18:12:36.543 124 <190>1 2016-07-08T22:12:36.238948+00:00 host app web.1 - Template render error: (/app/views/error.html) [Line 4, Column 19] /
» 8 Jul 2016 18:12:36.562 285 <158>1 2016-07-08T22:12:36.247322+00:00 host heroku router - at=info method=GET path="/favicon.ico" host=id-webmaker-org-production.herokuapp.com request_id=0ad6ac10-a940-47f6-afa7-4d128f0269c5 fwd="1.132.96.104,54.240.152.71" dyno=web.1 connect=1ms service=86ms status=200 bytes=1159 /
» 8 Jul 2016 18:12:36.626 121 <190>1 2016-07-08T22:12:36.239027+00:00 host app web.1 - Error: Unable to call `gettext`, which is undefined or falsey /
» 8 Jul 2016 18:12:36.626 139 <190>1 2016-07-08T22:12:36.239035+00:00 host app web.1 - at Object.exports.prettifyError (/app/node_modules/nunjucks/src/lib.js:34:15)
```

This happens when we try to serve the Webmaker 500 error which for some reason is erroring out because of node-webmaker-i18n. In reality, we shouldn't show the Webmaker page and show a custom "Something went wrong" page (I don't know the UI for this but it should probably include instructions to contact us).

I'm making this a P1 (feel free to change that) since I'm seeing a whole lot of these failures in the logs.

cc @hannahkane and @flukeout

flukeout:
We currently have a 404-type page for missing projects...

![image](https://cloud.githubusercontent.com/assets/25212/16742506/c332f1e6-475c-11e6-8d5a-55f5a806b736.png)

We could use it as a base for making an error page. I can work on a design.

We also need a straight up 404 page on the thimble.mozilla.org domain, currently we server this...

![image](https://cloud.githubusercontent.com/assets/25212/16742546/f086ae12-475c-11e6-9384-2cc8408d9789.png)

flukeout:
Looks like we already have an issue for a better 404 page here - #1130

Pomax:
![deadlink](https://cloud.githubusercontent.com/assets/177243/16742975/03bd976e-475f-11e6-85ab-474e82d182ce.png)

[...]

PR #1521, opened 2016-07-25
Better error logging

### LINKED ISSUE #1511: Better error messages
We need better logging on the server side that logs errors that occur along with providing some context so that we can debug these errors for clients.

gideonthomas:
Fix #1511

This patch implements better error logging practices as well as makes use of the work done in #1506 to display better error messages to the user.

This patch is divided into 4 distinct commits outlined as follows:

1. A quick optimization fix
2. The initial foundation for enabling logging which includes:
  * Server request logging with morgan
  * A logging module that can be used to log custom error messages
1. Adds the error handlers to the server to catch 404s and other errors, send them to the client and also log them
4. Calls the error handler implemented above for all error points in every route's logic.
5. JShint error fixes


server/index.js
```diff
@@ -13,13 +13,14 @@ let templatize = require("./templatize");
...
-let errorhandling = require("./lib/errorhandling");
+const HttpError = require("./lib/http-error.js");
 let middleware = require("./lib/middleware")();
 let routes = require("./routes")();
-let getFileList = require("./lib/utils").getFileList;
+const Utils = require("./lib/utils");
 
 let server = express();
-let isDevelopment = env.get("NODE_ENV") === "development";
+const environment = env.get("NODE_ENV");
```
cadecairos: It feels strange to use `const` all of a sudden, while values around it could technically also be `const`.

gideonthomas: around it?

cadecairos: yes, for example:

`let server = express()`

is server ever reassigned? probably not, so why isn't it const? it seems the current style in this file is using let for variable declaration.

gideonthomas: should I use that? Yeah it's pretty inconsistent right now. The idea is to do a full es6 re-write once node 6 is lts (october); so this is just to make that patch a bit smaller.

cadecairos: I'm not sure if making the code style inconsistent now is better than changing a few more lets to consts later on.

gideonthomas: ok, I'll make that change

cadecairos:
"500: Failed to get the list of your projects! The request to the publish server to login the user could not be completed. "

This first and last half of this message don't seem related?

gideonthomas:
@cadecairos updated the PR with the following:
* The message now shown to the user will be something like "Service Unavailable (500): Failed to create the project!" for e.g. which is cleaner and allows for fewer l10n strings
* To do the above, I had to change it so that the error message for the request is set at the middleware level. That is to say, an error message to use is already set for a "request to create a file" regardless of how many sub requests is needed to do that.
* I followed whatever convention is followed in the file, i.e. if `var` is used use `var`, if `let` is used use `let`, if it is a new file use `const`.

Based on this, re-r?

gideonthomas:
@cadecairos re-ping for review

cadecairos:
R+

PR #1522, opened 2016-07-27
Fix #1517 - Update security headers

### LINKED ISSUE #1517: Update server's security headers
Hey there!  April here from the Mozilla infosec team.  Just wanted to help you out a little bit, getting your server's headers inline with our security guidelines.  :)

> X-Frame-Options: ALLOW-FROM https://pontoon.mozilla.org

ALLOW-FROM is only supported by Firefox, and has been deprecated in place of this:

> Content-Security-Policy: frame-ancestors https://pontoon.mozilla.org

This works in Firefox, Chrome, and Safari and hopefully Edge someday.  To protect users of older browser from possible clickjackings, I would set it like this:

Content-Security-Policy: frame-ancestors https://pontoon.mozilla.org
X-Frame-Options: DENY

> Strict-Transport-Security: max-age=86400

Please set this to a year:

> Strict-Transport-Security: max-age=31536000

I see you're well on your way to getting CSP implemented, with Content-Security-Policy-Report-Only, which is absolutely fantastic.  Thanks so much!

gideonthomas:
* Update the CSP headers to use "frame-ancestors"
* Update HSTS header to use 1 year
* Remove CSP Report Only


gideonthomas:
@cadecairos re-ping for review

server/index.js
```diff
@@ -82,9 +82,9 @@ secure.xss()
 .csrf()
 .xframe()
 .csp({
-  frame: [ editorHost, "https://www.youtube.com/embed/JecFOjD9I3k" ],
-  script: [ editorHost ],
-  connection: [ editorHost ]
+  frameSrc: [ editorHost, "https://www.youtube.com/embed/JecFOjD9I3k" ],
```
cadecairos: Shouldn't this be "childSrc"

gideonthomas: Good catch, it should actually be both for browser compatibility

gideonthomas: Fixed!

cadecairos: Could I bother you once again to make sure tha if this youtube link needs to be updated ever, we only have to update it in one place?

gideonthomas: I'm not sure where the right place would be to put it. Env? Seems kinda dirty though

cadecairos: a variable would be fine, It's not going to change often enough to warrant an env var, but only having to change it in one place is a good idea.

gideonthomas:
@cadecairos re-r?

cadecairos:
R-

Issue #1480, opened 2016-05-04
Space missing between button icon & label


flukeout:
Not sure where it went to. Might be due to changes made when setting localization and a space might have gotten removed. Should control the spacing via a margin and not a space anyway...

![image](https://cloud.githubusercontent.com/assets/25212/15021637/ceac2110-11dc-11e6-99a0-e7f79547313e.png)

Joe-Para:
Just noticed that the space between text and icon is only missing when anonymous(which i fixed in above commit), but when logged in the space exists but because of some spaces in the text. (see: http://puu.sh/s3F20/01b88f2b81.png)

Its not in localization .properties files...

gideonthomas:
This has been merged in with ef12c7b680292e542acad8d5584cabc4e8ff97cf, thanks @Joe-Para

PR #1573, opened 2016-11-02
Fix spacing between publish icon and publish text

Joe-Para:
Added margin to the right of the icon to space it from the text.
See Issue #1480 

First commit, Woo!

### DIFF
--- a/public/resources/stylesheets/userbar.less
+++ b/public/resources/stylesheets/userbar.less
@@ -186,6 +186,7 @@ body {
       vertical-align: middle;
       position: relative;
       top: -2px;
+      margin-right: 4px;
     }
 
     &:hover {
--- a/views/editor/userbar.html
+++ b/views/editor/userbar.html
@@ -107,9 +107,7 @@
 
     {% if username %}
       <div id="navbar-publish" class="nav-chunk">
-        <div title="{{ publishBtnTitle }}" id="navbar-publish-button">
-          <img src="/img/icon/publish.svg" class="icon-publish">
-          {{ gettext("publishBtn") }}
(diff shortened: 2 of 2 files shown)


gideonthomas:
Thanks for the contribution @Joe-Para! This patch looks good.

It's interesting that those spaces are added in the text when you are logged in. It doesn't seem to be a localization issue since the properties files don't have those spaces. My guess is that `gettext` is preserving spaces that are present in the html.

Would you be able to add one more change to this patch - combine these three lines: https://github.com/mozilla/thimble.mozilla.org/blob/master/views/editor/userbar.html#L110-L113 into a single line so that it looks like `<div ...>{{ gettext }}</div>` i.e. no spaces in between them? Would like to test and see if that works.

Joe-Para:
I didn't think about the whitespace in the html causing issues, how interesting.

I can add these changed, but i have two questions:

Do you want me to test before i commit?
Can I just add another commit to the PR?

gideonthomas:
you can try testing locally, sure. And yes you can add another commit to the PR, when we merge it in, it will however show up as a single squashed commit.

I will also test it when you make the change so don't worry too much about whether this experiment worked or not.

Joe-Para:
Too easy...

Question: To get local testing running I'd have to set the entire website up locally by basically following the instructions there on home page, correct?

I appreciate the help and opportunity :)

gideonthomas:
@Joe-Para yeah since this patch is on Thimble and requires you to login to test the UI, you would have to set up all the services as specified on the homepage.

We are more than happy to accept contributions from contributors interested in this project and very grateful for their effort.

gideonthomas:
Ok weirdly, it seems to work now, no spaces are added to the beginning of the text. Thank you very much for the patch @Joe-Para, merging in.

PR #1580, opened 2016-11-08
Enable bramble-watch-index.html extension

humphd:
This turns on the bramble-watch-index.html extension (depends on Bramble changes in https://github.com/mozilla/brackets/pull/564), and adds the necessary locale strings for Pontoon.

r? @gideonthomas

### DIFF
--- a/public/editor/scripts/editor/js/bramble-editor.js
+++ b/public/editor/scripts/editor/js/bramble-editor.js
@@ -14,7 +14,13 @@ define(function(require) {
       // Start loading the Bramble editor resources
       Bramble.load("#webmaker-bramble",{
         url: options.editorUrl,
-        hideUntilReady: true
+        hideUntilReady: true,
+        extensions: {
+          enable: [
+            // Watch and warn users on rename or delete of index.html
+            'bramble-watch-index.html'
+          ]
+        }
       });
 
       // Start loading the project files
--- a/locales/en-US/editor.properties
+++ b/locales/en-US/editor.properties
@@ -189,3 +189,8 @@ ERROR_MOVING_FILE_DIALOG_HEADER=Move Error
 UNEXPECTED_ERROR_MOVING_FILE=An unexpected error occurred when attempting to move {0} to {1}
 # {0} is the name of the file/folder being moved and {1} is the name of the folder it is being moved to
 ERROR_MOVING_FILE_SAME_NAME=A file or folder with the name {0} already exists in {1}. Consider renaming either one to continue.
+
+# extensions/extra/bramble-watch-index.html
+
+INDEX_HTML_FILE_MISSING_DIALOG_TITLE=Warning - Missing index.html file
+INDEX_HTML_FILE_MISSING_DIALOG_MESSAGE=Projects without an index.html page will appear broken, for example: google.com/ actually refers to google.com/index.html. Add an index.html file or rename one of your current HTML files.


public/editor/scripts/editor/js/bramble-editor.js
```diff
@@ -14,7 +14,13 @@ define(function(require) {
       // Start loading the Bramble editor resources
       Bramble.load("#webmaker-bramble",{
         url: options.editorUrl,
-        hideUntilReady: true
+        hideUntilReady: true,
+        extensions: {
+          enable: [
+            // Watch and warn users on rename or delete of index.html
+            'bramble-watch-index.html'
```
gideonthomas: interestingly, this breaks localization.

I get: `Invalid language tag: en-us?enableextensions=bramble-watch-index.html`

gideonthomas: The error comes from Brackets!

I think the main problem is that we allow you to enable extensions and determine locale using the Bramble API (kinda like what you've done above) which means we can do query param parsing on the client and pass those to Brackets via the Bramble API. However, in Brackets, in a bunch of places, we read directly from the query string that has been passed in through the bramble url and we take advantage of this [here](https://github.com/mozilla/thimble.mozilla.org/blob/master/server/routes/main/editor.js#L82) to inject those query params that were passed into the thimble url directly into the bramble url, and our client API can't handle both (injecting options via the query params AND passing options via the API)

humphd: How is the locale being passed down on the top-level URL?  Aren't we doing that here if you request it? https://github.com/mozilla/brackets/blob/master/src/bramble/client/main.js#L344-L346  Maybe we need to set that flag so locale gets included?

gideonthomas: This is interesting! Since we're not passing the `useLocationSearch` value in, how is localization working right now? I'm gonna check

humphd: I think we should chat at some point to figure out the right fix for this, unless you know what to do and want to do a patch.  I'm a bit confused.

gideonthomas: sounds good! I'm kinda confused myself! L10N is working for Brackets but based on the code you link to above, it shouldn't. So, I'm gonna have to test this locally and see which parts of the code are hit.

gideonthomas: Oh wow! I know what's happening! This is where stuff goes wrong: https://github.com/mozilla/brackets/blob/master/src/bramble/client/main.js#L358

In that line of code, `options.url` is something like `localhost:8000?locale=en-US` because that's what we pass from Thimble as the bramble url. Then, we construct the search params in the API and end up with `?enableExtensions=...&locale=en-US` if a locale was also passed in through the API or if not, `?enableExtensions=...`. So, when we construct the iframe url in the code I linked, we end up with `localhost:8000?locale=en-US?enableExtensions=...` or `localhost:8000?locale=en-US?enableExtensions=...&locale=en-US` which is where the problem lies.

humphd: So it sounds like we need a patch to the Bramble client code (i.e., where you're pointing here) to take `options.url` and parse it, then merge any search params we find there with whatever options we want to pass down (i.e., whatever we got on `options`).  Want me to do that?

gideonthomas: sure!

humphd:
Did @Pomax recently fix the URL localization thing?  I thought I saw a PR go by.

gideonthomas:
I believe so

Pomax:
I did, https://github.com/mozilla/thimble.mozilla.org/pull/1605

Pomax:
@humph @gideon is this PR still intended for landing?

humphd:
It likely needs some work.  We had a bug (and maybe still have) where we couldn't enable extensions in the iframe via our API.

humphd:
I'm not going to do this, I think we can solve it in better ways.

Issue #1526, opened 2016-07-28
Intercept "ctrl-r" / "cmd-r" in instances where work might be lost.


Pomax:
I've run into multiple instances where I think I'm in a browser and so instinctively hit "reload page" as keyboard shortcut to see the updated changes. Getting a "are you sure you want to leave this page?" using a `window.onreload` might be a good idea

aayushsanghavi:
I tried setting up the project on my machine but I couldn't. Although I've written some javascript code in a separate file that resolves this issue (It throws an alert on pressing ctrl r before reloading the page). Can I just add the file and make a PR and then maybe one of the contributors could integrate it.

gideonthomas:
@aayushsanghavi apologies for the delayed response.

Sure, feel free to open up a PR with your code.

We are also happy to help you with the problems you are having setting up Thimble on your local machine. Just let us know where (or in which step) you are experiencing issues and what operating system you are running.

aayushsanghavi:
Thanks @gideonthomas for your help but I don't remember where I was stuck earlier. So I guess I'll just add a new JavaScript file to the repository. All that has to be done is to call include that file in one of the main code files. Kindly review the pull request I make.

flukeout:
Based on the work by @aayushsanghavi in #1602 we've come to realize that in order for this to be useful, we'd have to detect when you are actually risking losing work by refreshing. The most obvious case for this is when the project you have hasn't fully synced with the remote DB.

If we can detect that case, then it might be useful to show an alert when the user tries to leave the editor. We can make that the focus of this issue. Can we reliably detect that state?

humphd:
We deal with that case here:

https://github.com/mozilla/thimble.mozilla.org/blob/fa3dde088dc016a562520e62a9df1f5558ba8f70/public/editor/scripts/editor/js/fc/sync-state.js

Likely the real fix is to figure out why we're not doing what you expect there, since we already have `beforeunload` handlers to stop the page from closing.

Pomax:
I'm going to close this for now, as it's become a bit of an undefined issue, and I'll have to run into this again before I can properly explain the problem again.

ryanwarsaw:
@Pomax Are you okay with me re-opening this? It's definitely a race condition issue, not as much of an issue for those on strong & stable internet connections, but still an issue nonetheless.

Pomax:
[...]

PR #1602, opened 2017-01-10
Intercepting ctrl + r

aayushsanghavi:
#1526 
This file contains code for intercepting ctrl + r before reloading the page and displaying an alert. It is dependent on jQuery so jQuery will have to be included somewhere.


aayushsanghavi:
@gideonthomas what do you think about the code?

Pomax:
Hi @aayushsanghavi - rather than using jQuery, browsers support this "out of the box" so it will be better to hook into that functionality instead. Using this, for instance:
```
window.onbeforeunload = function() {
  return "Data will be lost if you leave, are you sure?";
};
```
will pop up a warning "the page you're on needs you to confirm leaving: Data will be lost if you leave, are you sure?" (the first part is browser-specific, the second part is the message you specified).

aayushsanghavi:
@Pomax thanks for the tip. I was unaware of this. I'll make the required changes soon.
{I'm going to have to start reading MDN now. I need to know all this.}

aayushsanghavi:
@Pomax I used beforeunload but it isn't displaying the message I'm returning in the function. It displays the default warning instead.
Please review the change.

humphd:
@aayushsanghavi browsers won't use your string, they always use the default.  They do this for security reasons.  It's annoying.  All you can do is return a string, but what it contains is redundant, and will be ignored.

Pomax:
@aayushsanghavi  looking good! If you make sure it's using two-space indentation rather than tabs, we can merge this in =)

scripts/reload.js
```diff
@@ -0,0 +1,5 @@
+window.addEventListener("beforeunload", function (e) {
+	var confirmationMessage = "Are you sure?";
```
Pomax: thimble uses 2-space indentation rather than "real tabs"

aayushsanghavi: Yeah, I'll make that change.

aayushsanghavi:
@Pomax I changed the tabs to a 2 spaced indentation.

Pomax:
@aayushsanghavi  nice. Where is this file getting loaded in now, though? I don't see code changes in this PR that effect it also getting loaded?

aayushsanghavi:
@Pomax yes, the file isn't being loaded anywhere. Could you tell me where to load it? I couldn't set up thimble so I thought that maybe you or any other contributor could do it or tell me where to do it.

Pomax:
Ah, we recently improved the whole "getting thimble to run", have a look at the updated readme and you should be able to get things working now. (If not, that's also valuable information that'll hopefully let us update the process to work for you, too).

The main client bundle entry point is https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/main.js, so the easiest way to ensure this works is to rewrite it to a small require.js module, which mostly involves wrapping our code as:
```
define(function(require) {
  return function() {
    // your beforeundload code here
  };
});
```

and then add it at https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/main.js#L82 as an additional requirement, with `init` given an extra argument at https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/main.js#L55, and a line inside the init function that enables the functionality:

```
function init(BrambleEditor, Project, SSOOverride, ProjectRenameUtility, enableReloadProtection) {
  ...
  enableReloadProtection();
  ...
}
```

aayushsanghavi:
@Pomax I made some changes but I'm sure I must've messed up somewhere. Please do have a look and let me know.

public/editor/scripts/require.js
```diff
@@ -0,0 +1,16 @@
+function unloadProtection(e) {
```
Pomax: please keep this file named after what it does, not "require.js". `require.js` is the bundling software, this should probably be called "reload-protection.js" instead

public/editor/scripts/main.js
```diff
@@ -79,5 +82,5 @@ require(["jquery", "bowser"], function($, bowser) {
     });
   }
 
-  require(["bramble-editor", "project/project", "sso-override", "fc/project-rename"], init);
+  require(["bramble-editor", "project/project", "sso-override", "fc/project-rename", "reloadProtection"], init);
```
Pomax: this name is the part that matches against what you defined in  public/editor/scripts/main.js  above

aayushsanghavi: I defined it as reloadProtection in main.js above and used the same name below. Should it be unloadProtection? Or could you tell me what exactly to change?

public/editor/scripts/main.js
```diff
@@ -52,12 +53,14 @@ require(["jquery", "bowser"], function($, bowser) {
 
   Bramble.once("error", onError);
 
-  function init(BrambleEditor, Project, SSOOverride, ProjectRenameUtility) {
+  function init(BrambleEditor, Project, SSOOverride, ProjectRenameUtility, unloadProtection) {
```
gideonthomas: @aayushsanghavi I think what @Pomax meant (I might be wrong), is that the naming should be consistent throughout this file.

Short explanation: it shouldn't be `unloadProtection`, it should be `reloadProtection` here

Longer explanation:
The parameter that is passed here in this function (what is currently called `unloadProtection`) might not be the right name to use. What you are actually getting in this parameter, is the defined module itself (and from looking at the `reload-protection.js` file, it looks like it's an object) and not the function you have defined in that module (`unloadProtection`). So, since the module is an object, and you named your module `reload-protection.js`, it makes more sense for this function parameter to be called `reloadProtection`. `unloadProtection` is an internal method in that module that you use in the module file; but in this file, we don't need to worry about that because it doesn't even have access to the `unloadProtection` function, only to the object that you returned.

aayushsanghavi: Okay, I think I get what you mean. Thanks @gideonthomas

public/editor/scripts/main.js
```diff
@@ -52,12 +53,14 @@ require(["jquery", "bowser"], function($, bowser) {
...
   Bramble.once("error", onError);
 
-  function init(BrambleEditor, Project, SSOOverride, ProjectRenameUtility) {
+  function init(BrambleEditor, Project, SSOOverride, ProjectRenameUtility, reloadProtection) {
     var thimbleScript = document.getElementById("thimble-script");
     var appUrl = thimbleScript.getAttribute("data-app-url");
     var projectDetails = thimbleScript.getAttribute("data-project-details");
     var editorUrl = thimbleScript.getAttribute("data-editor-url");
 
+    unloadProtection.enable();
```
gideonthomas: this still needs to be changed

gideonthomas:
@flukeout when you're back, I'd like you to try this patch out. I'm a bit on the fence about the UX around this because this not only shows the dialog when you hit Ctrl+R, it also shows you the dialog when you attempt to close the tab or navigate away from the webpage. Since it says, "Changes you made might not have been saved", I feel like it might be misleading since it's not always true.

flukeout:
Hey team, will try out this code sometime today and chime in.

flukeout:
I'm not sure about this change. The only time I'd like to see this dialog is if I'm about to lose work because my project hasn't been fully saved to the cloud yet. Otherwise, it's just an annoyance.

I agree with @Pomax about refreshing accidentally being annoying. Is there a way to intercept the refresh action only, and not all attempts to leave or close the tab? If so, then I could see us just refreshing the Preview (and spinning the refresh arrow on that view) when we see a refresh attempt.

That said, I can't think of an app or website that overrides that basic browser behaviour, so I'm a little hesitant to do that. I'll think about it some more. @Pomax thoughts?

Pomax:
I know plenty of sites that prevent you from leaving if you have unsaved content (Facebook being the clearest example), but I think that if your project has unsaved content, it should *always* ask whether you meant to leave the page, regardless of whether it's through an accidental reload, or by accidentally hitting "back", or by accidentally hitting cmd-L followed by an enter. All navigation is "easily" done by accident (a backspace key with the focus on the page? Hurray, you've just gone back in your history!) so offering protection across the board feels appropriate.

I do like the idea of making sure it only pops up when there is unsaved content, though. If the data has been saved, then reloading or coming back to the back through the history is just an incovenience in timing, not in data loss... @flukeout @gideonthomas is there a way to tell whether a project has unsaved changes so this code can do something like an `if (!project.pending) return false;` shortcut?

flukeout:
@Pomax Just to be clear, I fully support trying to prevent data loss, wether through refreshing, leaving the page or closing the tab. The original issue you filed doesn't data loss as the primary concern - so I'm glad we're on the same page. 

Also curious about what Gideon says, though I believe we have access to all of that info.

gideonthomas:
lol @flukeout!

I think it depends on what "saving" means in this context. 

Saving to IndexedDB happens pretty frequently (every 200ms I believe). so, even if that content is not on the server yet, as long as you are using the same machine (which I would assume on a Cmd+R) and the same url, you won't have any data loss. Currently, we don't have code to track if there are any unsaved (w.r.t. IndexedDB) changes but I don't think it will be too complicated to add that to the Bramble API. 

Saving to the server happens every 1 min. We can listen for events (we already have code for this) for when it begins/ends. So, in this case, we do know when we have unsaved (to the server) changes.

flukeout:
@Pomax have you ever lost work accidentally? I never remember having lost any work. This _might_ be more of an issue in classroom settings where you might not necessarily use the same computer next time you use Thimble.

Pomax:
both - at the time I filed this I think I was working in thimble on the train and the flaky internet made ctrl-r really dangerous.

aayushsanghavi:
@Pomax @flukeout @gideonthomas if there is anything to be changed in the code or if you need any input from my side, do let me know.

flukeout:
Sorry to keep you waiting on this issue @aayushsanghavi - as you can tell we don't have full agreement on what the best course of actions is. We cannot use the current set of changes you made as they as they are too restrictive. We will meet and discuss this feature today and update the issue with next steps. Thanks for your patience.

flukeout:
@aayushsanghavi As it turns out, I think issue is a bit more complicated than when we first thought, so I'm going to close this PR and leave some more details comments in the original issue.

PR #1605, opened 2017-01-11
fix locale redirection due to joining as filename

Pomax:
URL redirects were being concatenated as if they were filenames, rather than being URL strings. These are two completely unrelated things, and the assumption broke the code on all OS that don't use `/` as standard path separator. This patch ensures that the URL is written correctly.

### DIFF
--- a/server/localize.js
+++ b/server/localize.js
@@ -35,10 +35,12 @@ module.exports = function localize(server, options) {
     let bestLanguage = bestlang(langPrefs, webmakerI18N.getSupportLanguages(), "en-US");
     let localizedUrl;
 
+    let orgUrl = req.originalUrl;
+
     if(knownLocales.indexOf(urlLocale) !== -1) {
-      localizedUrl = req.originalUrl.replace(urlLocale, bestLanguage);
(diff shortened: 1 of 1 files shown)


humphd:
This is great.  I'm sure we have a bug on this too, which we should close when this lands.

server/localize.js
```diff
@@ -35,10 +35,12 @@ module.exports = function localize(server, options) {
...
     let localizedUrl;
 
+    let orgUrl = req.originalUrl;
+
     if(knownLocales.indexOf(urlLocale) !== -1) {
-      localizedUrl = req.originalUrl.replace(urlLocale, bestLanguage);
+      localizedUrl = orgUrl.replace(urlLocale, bestLanguage);
     } else {
-      localizedUrl = path.join("/", bestLanguage, req.originalUrl);
+      localizedUrl = "/" + bestLanguage + "/" + (orgUrl === "/" ? "" : orgUrl);
```
gideonthomas: I don't understand this conditional wouldn't:
`"/" + bestLanguage + orgUrl`
do the same thing?

Pomax: not when `orgURL` is `/` - that makes the localizedUrl the string `//`

PR #1663, opened 2017-01-24
Fixed indentation in index.html page of new projects

Rajat-dhyani:


humphd:
@flukeout see if you're happy with these changes.

flukeout:
Looks good to me, good to merge. 👍

public/resources/default-files/html.txt
```diff
@@ -1,15 +1,12 @@
-<!DOCTYPE html>
```
ryanwarsaw: This line was removed entirely, it should probably get added back in.

flukeout: Oh wow, thanks for catching that! Yes, let's keep that in.

Rajat-dhyani: @ryanwarsaw @flukeout  
so sorry sir...i have fixed this now

flukeout:
Thank you very much @Rajat-dhyani 👍 🌮

Rajat-dhyani:
More then happy to contribute sir 👍 :) @flukeout

public/resources/default-files/html.txt
```diff
@@ -1,15 +1,13 @@
...
-  <p>Make something <strong>amazing</strong> with the web!</p>
-
-</body>
-</html>
\ No newline at end of file
+  <head>
+    <title>Made with Thimble</title>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link rel="stylesheet" href="style.css">
```
ryanwarsaw: @flukeout It doesn't appear these changes have landed on staging yet, but this specific line wasn't present on the original in comparison to the diff, but when I create a new project I see this link added in. Are we going to have two links to style.css now?

flukeout: Oh weird, you're right - I'm not sure why it doesn't appear in the original. To me it looks like we'll just be adding the style.css file one, on line 6. Am I totally misreading something?

flukeout: Is this the final version we're rolling with?

image

ryanwarsaw: I'm still seeing the old version on staging, I was waiting for cloud front cache to invalidate to see if it'd automatically insert it twice or what the deal was, I think we'll have to dig into this and figure out what's the deal.

ryanwarsaw:
@flukeout This never seems to have invalidated so either staging isn't working properly, or there is a source file somewhere else in Thimble that creates the template.

gideonthomas:
Thanks @ryanwarsaw (and nice catch)!
@flukeout so I think this fixes the wrong file.

I found out that the file that was edited in this PR seems to be the file you get when you click the "Add New HTML File" button from the "+" dropdown menu. This seems to be distinct from the  default project (with the default index.html file) that you get when you load Thimble. Not sure why they are separate (I assume we might have wanted different copy on both).

One clear distinction between the two is the missing `link` tag to `style.css` in the file edited here. It should be that way since we don't want to link the the same stylesheet.

Instead of reverting this patch, a follow up patch needs to be done where the `link` tag from this file is removed and the indentation is also fixed in https://github.com/mozilla/thimble.mozilla.org/blob/master/default/empty-project/index.html.

flukeout:
Yeah @ryanwarsaw - nice work here. I'm guilty of not actually having tested the patch, it just looked right! Thanks. I'll make sure to fully test next time.

Rajat-dhyani:
hello sir @ryanwarsaw  @gideonthomas @flukeout ,
i was out of town last week so could not reply...i too noticed the issues and sorry for not working it out myself.. @flukeout thanks for pointing out the file related to the issue and @gideonthomas will fix the issue realted. Sorry again for being absent

PR #1680, opened 2017-01-28
Fix #1534 - Publish button in some circumstances wouldn't trigger on click

### LINKED ISSUE #1534: "publish" button frequently does nothing unless I slow-click down and back up
If I just quickly click the "publish" button for my project while moving my mouse around because I'm still busy, quite often nothing happens and I end up having to position the mouse over the button, and click without moving the mouse at all.

It feels like the wrong event is being tapped into in order to determine whether a user clicked the publish button, where it listens for "click" (which can get frustrated by a click-drag when a user is clicking and moving the mouse at the same time) when it should be listening for "mousedown" instead.

cgsingh:
Originally, you were not able to trigger the 'Publish Dialog' window when moving the mouse and clicking the 'Publish' button (click-drag). You could also not trigger the window when clicking the very top edge of the button.

The change I made:
publishmousedown

The 'Publish' button uses the mousedown event where it'll only show the dialog window if it is a left-click. You can move the mouse as fast as you want, click-drag, and the dialog will trigger as it is registered as a mousedown event and a left-click. This also seems to solve the top-edge issue where clicks would not register:
publishtopedge

### DIFF
--- a/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
+++ b/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
@@ -330,7 +330,11 @@ define(function(require) {
 
     if (Project.getUser()) {
       //Publish button
-      $("#navbar-publish-button").click(showPublishDialog);
+      $("#navbar-publish-button").mousedown(function(e) {
+        if (e.which === 1) {
+          showPublishDialog();
+        }
+      });
       $("#publish-button-cancel").click(hidePublishDialog);
 
       publisher = new Publisher();


flukeout:
@cgsingh Does mousedown work on touch devices? Can you test by emulating a touch device in developer tools?

Pomax:
If it doesn't it should be fairly easy to add a secondary "touchstart" event. That said, do we currently support touch devices (that aren't desktop touch-enabled monitors) @flukeout @humphd ?

cgsingh:
@flukeout It seems to work when I test it using Chrome Device mode in Developer tools.

flukeout:
Hey @cgsingh thanks for checking. I thought about it a bit more and realized why it was bugging me. Almost not button-like UI element works immediately when you click down on it, they all work on the click release. This gives the user a split second to change their mind by mousing off of the element they're interacting with it.

It may seem small, but I think it's important, and matches how the rest of our UI functions. I don't think we should fix this bug in a way that changes how the button works.

Here's an interesting article about this very thing, maybe we could use the solution there:
https://css-tricks.com/video-screencasts/123-if-it-moves-when-you-click-make-something-stick/

It involves adding the ":after" pseudo-element to the link that is slightly bigger than the element itself.


Just for reference, here's me clicking on the comment button below but mousing off...
button

cgsingh:
@flukeout I've got it working. I did some testing on here: [http://codepen.io/cgsingh/pen/JEpBgL](url). And I also edited the userbar.css using Google Chrome's developer tools. I'm having an irritating problem where I've edited the userbar.less file but when I use vagrant up, it refuses to update the code to reflect my changes. I also had to completely remove my local clone of my repo because of a strange bug with Git, and I think my original branch is messed up. I have my work on a new branch so I'll be closing this request and making a new one.

PR #1683, opened 2017-01-31
[WIP]Updated installation instructions

simon66:
While installing Thimble, I ran into some installation problems. I didn't know if Bramble was required or not (Since it only mentions needing it if you plan to install it locally).
I've updated the instructions. Hopefully is more clear :)
I've removed the Bramble installation part since the Bramble repository has a really good, in-depth installation instruction. 
Is there anything you would like me to add/remove?

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,14 +25,13 @@ Thimble interacts with the Publish API (source managed in [publish.webmaker.org]
 
 For authentication and user management, Thimble uses Webmaker OAuth which consists of the Webmaker ID System (source managed in [id.webmaker.org](htps://github.com/mozilla/id.webmaker.org)) and the Webmaker Login API (source managed in [login.webmaker.org](https://github.com/mozilla/login.webmaker.org)).
 
-All three services along with Thimble are bundled together using Git subtrees and can either be run together using Vagrant, or [manually](#1b-manual-installation).
+All three services along with Thimble are bundled together using Git subtrees and can either be run together using Vagrant, or [manually](#manual-installation).
 
-# 1) Installing Thimble and Services
-##1a) Automated Installation (Preferred Method)
-**Note:** If you aren't able to properly run virtualization software on your machine (for e.g. some versions of Windows only allow one virtualization client to run at a time and if that isn't VirtualBox, you can't run the required VirtualBox as well. This is often a problem if you have docker installed on Windows) or are trying to host Thimble on your own, refer to the [Manual Installation](#1b-manual-installation) instructions instead.
+## Automated Installation (Preferred Method)
+**Note:** If you aren't able to properly run virtualization software on your machine (for e.g. some versions of Windows only allow one virtualization client to run at a time and if that isn't VirtualBox, you can't run the required VirtualBox as well. This is often a problem if you have docker installed on Windows) or are trying to host Thimble on your own, refer to the [Manual Installation](#manual-installation) instructions instead.
 
-### Pre-Requisites for Automated Installation
-In order for Thimble to be installed correctly, the following dependencies needs to be installed in order:
+### Prerequisites for Automated Installation
+In order for Thimble to be installed correctly, the following dependencies need to be installed in order:
 
 - Node.js (version 4.6 or later) [[download](https://nodejs.org/en/download/)]
 - [Brackets (Bramble)](#installing-brackets-bramble)
@@ -56,7 +55,7 @@ For the first time, to start all dependent services and Thimble, simply run:
 vagrant up
 ```
 
-Depending on your internet connection speed, this process can take a while (Since it needs to download all dependencies not listed above).
+Depending on your internet connection speed, this process can take a while (since it needs to download all dependencies not listed above).
 Once you see `Express server listening on http://localhost:3500`, you are ready to start using Thimble on [http://localhost:3500](http://localhost:3500)!
 Any changes made to the Thimble source code on your system will automatically be reflected on [http://localhost:3500](http://localhost:3500).
 
@@ -70,16 +69,16 @@ vagrant reload --provision
 
 This will take less time to setup compared to the `vagrant up` command.
 
-##1b) Manual Installation
-You can also setup Thimble and it's needed components outside Vagrant and Virtualbox. This might be needed if you want to:
+## Manual Installation
+You can also setup Thimble and its needed components outside Vagrant and Virtualbox. This might be needed if you want to:
 - Host your own instance of Thimble
 - Cannot run virtualization software on your computer
 
-### Pre-Requisites for Manual Installation
-In order for Thimble to be installed correctly, the following dependencies needs to be installed:
+### Prerequisites for Manual Installation
+In order for Thimble to be installed correctly, the following dependencies need to be installed:
 
 - Node.js 4.x or above (see note below)
-  - * **Note:** The login.webmaker.org dependency needs a node version of 4.x only while all the other dependencies work with a node version of 4.x and above. We suggest installing [NVM](https://github.com/creationix/nvm) to allow the use of multiple versions of node.
+  - **Note:** The login.webmaker.org dependency needs a node version of 4.x only while all the other dependencies work with a node version of 4.x and above. We suggest installing [NVM](https://github.com/creationix/nvm) to allow the use of multiple versions of node.
 - [Brackets (Bramble)](#installing-brackets-bramble)
 - [Webmaker ID server](#idwebmakerorg)
 - [Webmaker Publishing Server](#publishwebmakerorg)
@@ -97,7 +96,6 @@ Please note: On Windows, use ``copy`` instead of ``cp``
 * Run ``cp env.dist .env`` to create an environment file
 * Run ``npm install`` to install dependencies
 * Run ``npm start`` to start the server
-* Once everything is ready and running, Thimble will be available at [http://localhost:3500/](http://localhost:3500/)
 
 ####id.webmaker.org
 * Clone https://github.com/mozilla/id.webmaker.org
@@ -122,9 +120,10 @@ Please note: On Windows, use ``copy`` instead of ``cp``
 * Clone https://github.com/mozilla/publish.webmaker.org
 * Run ``npm install`` to install dependencies
 * Run ``npm run env``
-* Run ``npm install knex -g`` to install knex
 * Run ``npm run knex`` to seed the publish database created earlier
 * Run ``npm start`` to run the server
+
+Once everything is ready and running, Thimble will be available at [http://localhost:3500/](http://localhost:3500/)
 
 ### Getting Ready to Publish
 To publish locally, you'll need to do the following...
@@ -134,7 +133,7 @@ To publish locally, you'll need to do the following...
 * Run ``createdb webmaker_oauth_test`` to create a test database
 * In your id.webmaker.org folder
   * Run ``node scripts/create-tables.js``
-  * Edit ``scripts/test-data.sql`` and replace it's contents with:
+  * Edit ``scripts/test-data.sql`` and replace its contents with:
 
       ```sql
         INSERT INTO clients VALUES
@@ -156,21 +155,6 @@ To publish locally, you'll need an account.
 * When you've created your account, click ``Set permanent password instead``
   * This lets you authenticate your account without needing email
 * Go back to Thimble and Log In with your new account
-
-# 2) Running the parts
-If you installed Thimble via the **Automated Installation** method, you need to run the following commands:
-
-* Navigate to Thimble and run: ``vagrant reload --provision``
-* Navigate to Bramble and run: ``npm start``
-
-If you installed Thimble via the **Manual Installation** method, you need to run the following commands:
-
-* Thimble ``npm start``
-* Bramble ``npm start``
-* PostgreSQL Database ``postgres -D /usr/local/var/postgres``
-* Webmaker ID server ``npm start``
-* Webmaker Login Server ``npm start``
-* Webmaker Publishing Server and Published Projects Server ``npm start``
 
 
 It's that simple! You are now ready to start using Thimble to its full potential!
--- a/README.md
+++ b/README.md
@@ -19,42 +19,45 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
 
 Thimble requires a modern web browser, and we recommend using Mozilla Firefox or Google Chrome.
 
+#Setup/Installation
+
 Thimble interacts with the Publish API (source managed in [publish.webmaker.org](https://github.com/mozilla/publish.webmaker.org)) to store users, projects, files and other content as well as publish user projects.
 
 For authentication and user management, Thimble uses Webmaker OAuth which consists of the Webmaker ID System (source managed in [id.webmaker.org](htps://github.com/mozilla/id.webmaker.org)) and the Webmaker Login API (source managed in [login.webmaker.org](https://github.com/mozilla/login.webmaker.org)).
 
-All three services along with Thimble are bundled together using Git subtrees and run together using Vagrant.
+All three services along with Thimble are bundled together using Git subtrees and can either be run together using Vagrant, or [manually](#1b-manual-installation).
 
-Setup/Installation
---------
+## 1) Installing Thimble and Services
+### 1a) Automated Installation (Preferred Method)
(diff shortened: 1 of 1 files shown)


Pomax:
Hi @simon66, thanks for the PR, but I'd like to suggest going in the opposite direction =)

While it may seem a good idea to "remove extraenous information" from a README, for open source projects it is usually actually far better to have as much information in that file. In this case, that means that rather than pruning information about Bramble, these instructions should be preserved, but moved into a section where that information makes sense. To that end, we should make it clear that there are two ways to install things:

1. Vagrant, a one-stop way to install everything, and
2. When vagrant won't work for you, the proper full list of which projects need to be installed in parallel, which includes thimble, bramble, id, login, and publish.

Rather than taking everything except for Vagrant out of the README.md, it will be far more valuable to split the README instructions up into two sections: an initial part "using vagrant" that explains things in terms of just using Vagrant, and then a subsequent more elaborate "manual installation" part that explains how to install the components manually, should you need to.

And you might need to go through the history of the README.md a bit because when the vagrant solution landed about a week ago, we already removed a little too much information that we should really restore around the manual installation process.

simon66:
Thanks for the feedback @Pomax! I actually like the idea of the "manual installation" part. I noticed that it was there before, but it has since been removed.
I'll update/add the missing portion :)

simon66:
Hi @Pomax. I've added the requested section. I went back to the previous versions of the README.md file and added them in.
Let me know if theres anything that needs to be changed

README.md
```diff
@@ -19,53 +19,105 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
...
+All three services along with Thimble are bundled together using Git subtrees and run together using Vagrant.
+
+Setup/Installation
+--------
 
-### Requirements
+## 1) Pre-Requisites
+In order for Thimble to be installed correctly, the following dependencies needs to be installed:
 
+- Brackets (Bramble) [[download](https://github.com/mozilla/brackets)]
```
gideonthomas: let's remove the "download" link and instead link it to the "Installing Brackets (Bramble)" section

README.md
```diff
@@ -19,53 +19,105 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
...
 - Node.js (version 4.6 or later) [[download](https://nodejs.org/en/download/)]
 - Virtualbox (version 5.1 or later) [[download](https://www.virtualbox.org/wiki/Downloads)]
 - Vagrant (version 1.9 or later) [[download](https://www.vagrantup.com/downloads.html)]
-  - __Note:__ On Windows machines, you may need to restart your computer after installing Vagrant for it to be fully usable
+  - __Note:__ On Windows machines, you may need to restart your computer after installing Vagrant for it to be fully usable.
 
-The setup of Thimble can be divided into two distinct sections:
+### Installing Brackets (Bramble)
+* Fork and clone https://github.com/mozilla/brackets
```
gideonthomas: I honestly think that the instructions in https://github.com/mozilla/thimble.mozilla.org/pull/1683/files#diff-04c6e90faac2675aa89e2176d2eec7d8L38 (and below) to install Brackets are a bit clearer so I would undo this change. (I still need to update the Wiki with the correct instructions for manual install)

README.md
```diff
@@ -91,4 +143,4 @@ Contact Us
 
 We're a friendly group, so feel free to chat with us in the "Thimble" channel on Mozilla Chat running on [Mattermost](https://about.mattermost.com). To access Mozilla Chat head to [this link]( http://chat.mozillafoundation.org). Note that you will need to create an account first.
 
-You can also download a mobile or desktop client for Mattermost [here](https://about.mattermost.com/download/#mattermostApps).
+You can also download a mobile or desktop client for Mattermost [here](https://about.mattermost.com/download/#mattermostApps).
```
gideonthomas: What change was made here?

README.md
```diff
@@ -19,53 +19,105 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
...
-For authentication and user management, Thimble uses Webmaker OAuth which consists of the Webmaker ID System (source managed in [id.webmaker.org](htps://github.com/mozilla/id.webmaker.org)) and the Webmaker Login API (source managed in [login.webmaker.org](https://github.com/mozilla/login.webmaker.org)).
+```
+vagrant reload --provision
+```
 
-All three services along with Thimble are bundled together using Git subtrees and run together using Vagrant.
+This will take less time to setup compared to the `vagrant up` command.
 
-The first step is to fork and clone Thimble and navigate to the cloned directory in a terminal shell.
+### Manual Installation (Optional)
```
gideonthomas: I don't think we should add all the instructions here in this readme since it's becoming a bit too long. I would instead link to the Wiki page with all this information: https://github.com/mozilla/thimble.mozilla.org/wiki/Setting-Up-Thimble-without-Virtualization

Pomax: @gideonthomas Then we disagree. Remember that this README is also for tomorrow's user, including people well version in development, who can't use vagrant due to Hyper-V requirements on windows. They should not have to hunt for instructions, and having a "more detailed section" at the bottom of the immediate project documentation is quite common for larger projects (Which Thimble definitely falls under). I would very much advocate we keep the manual installation instructions immediately available rather than locked away in a "you need to go back online" (the README.md downloads with thimble as part of a `git clone` operation - the wiki most definitely does not)

gideonthomas: Good point, I don't think I considered the case for the Hyper-V pitfall with Windows. Alright, I think it's fine to leave these in here for now specifically for that case until we figure out a solution that fixes the virtualization problem on Windows entirely.

I would however make some changes:
- Below "Automated Installation" add the following copy `**Note:** If you aren't able to properly run virtualization software on your machine (for e.g. some versions of Windows only allow one virtualization client to run at a time and if that isn't VirtualBox, you can't run the required VirtualBox as well. This is often a problem if you have docker installed on Windows) or are trying to host Thimble on your own, refer to the [Manual Installation](#manual-installation) instructions instead.`
- Remove "Optional"
- Stick to the instructions from the Wiki here: https://github.com/mozilla/thimble.mozilla.org/pull/1683/files#diff-04c6e90faac2675aa89e2176d2eec7d8L38 since they are a bit clearer and more accurate

simon66:
Thanks for the feedback :)
Going to make these changes asap!

README.md
simon66: I actually don't know hahah.
Nothing got changed in this part really. Not too sure why github thinks I changed it haha.

README.md
simon66: I'm a bit confused here. So you want me to revert just the Brackets installation section?

humphd: Ideally you want to clone Brackets with `git clone --recursive ...` to avoid having to also init/update the submodules.

README.md
```diff
@@ -19,53 +19,105 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
...
+* Make sure you're on the ``bramble`` branch
+* Run ``git submodule update --init`` to install submodules
+* Run ``npm install`` to install dependencies
+* Run `npm run build` to build Bramble
+* Run ``npm start`` to get a static server running on [http://localhost:8000/src](http://localhost:8000/src). You can try the demo version at [http://localhost:8000/src/hosted.html](http://localhost:8000/src/hosted.html)
 
-### Editor
+## 2) Thimble Installation
+### Automated Installation (Preferred Method)
+After forking and cloning Thimble, installing it is quite simple. All we need to do is run the following command inside the cloned directory and let vagrant do the heavy lifting for us:
```
humphd: Vagrant (capital 'V').

humphd:
@simon66 thanks for jumping in to fix docs.  This is great initiative.

README.md
gideonthomas: Alright, not a problem!

README.md
gideonthomas: @simon66 yeah I think the instructions previously for the Brackets installation process (which I linked above and is highlighted in red) might be a bit clearer. That being said, the line about `npm start` can probably be what you have here. Also, I would love to hear your thoughts on which of the two instructions are clearer and why.

gideonthomas:
@ryanwarsaw I think the changes you commented on were changes made based on the manual steps that were in the README before. In [this comment](https://github.com/mozilla/thimble.mozilla.org/pull/1683#discussion_r98940626) I recommended that @simon66 use the instructions from the Wiki instead which should probably address all your comments (if not, please let me know so that I can update them).

ryanwarsaw:
@gideonthomas Just wanted to make sure these issues were documented 👍

simon66:
I just added/changed the README based on the comments. Hopefully I didn't miss anything. Please let me know if I did :)

Here are my thoughts on the new README:
I think the new file is quite long (And maybe a bit confusing). It would probably be a good idea to put the Manual Installation portion on a new README and link it to the current one.
What do you think?

README.md
```diff
@@ -19,53 +19,161 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
...
-- Virtualbox (version 5.1 or later) [[download](https://www.virtualbox.org/wiki/Downloads)]
-- Vagrant (version 1.9 or later) [[download](https://www.vagrantup.com/downloads.html)]
-  - __Note:__ On Windows machines, you may need to restart your computer after installing Vagrant for it to be fully usable
+For authentication and user management, Thimble uses Webmaker OAuth which consists of the Webmaker ID System (source managed in [id.webmaker.org](htps://github.com/mozilla/id.webmaker.org)) and the Webmaker Login API (source managed in [login.webmaker.org](https://github.com/mozilla/login.webmaker.org)).
 
-The setup of Thimble can be divided into two distinct sections:
+All three services along with Thimble are bundled together using Git subtrees and can either be run together using Vagrant, or [manually](#1b-manual-installation).
 
-### Editor
+# 1) Installing Thimble and Services
```
gideonthomas: General comment about headings - let's remove the numbering from all headings since the heading level generally indicates the list level as well

gideonthomas: This heading specifically - let's remove it since it doesn't really add any context that's different than what comes before it. Also, leave all the other heading levels as they are (for e.g. the heading below, don't change it from ## to # to compensate for removing this heading)

README.md
```diff
@@ -19,53 +19,161 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
...
+##1a) Automated Installation (Preferred Method)
+**Note:** If you aren't able to properly run virtualization software on your machine (for e.g. some versions of Windows only allow one virtualization client to run at a time and if that isn't VirtualBox, you can't run the required VirtualBox as well. This is often a problem if you have docker installed on Windows) or are trying to host Thimble on your own, refer to the [Manual Installation](#1b-manual-installation) instructions instead.
 
-To fully use Thimble locally, you need to first setup Brackets locally first. This can be done by following the steps outlined below:
+### Pre-Requisites for Automated Installation
```
gideonthomas: Pre-requisites should be Prerequisites

README.md
```diff
@@ -19,53 +19,161 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
...
+In order for Thimble to be installed correctly, the following dependencies needs to be installed in order:
```
gideonthomas: "following dependencies needs" should be changed to "following dependencies need"

README.md
```diff
@@ -19,53 +19,161 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
...
 
 For the first time, to start all dependent services and Thimble, simply run:
+
 ```
 vagrant up
 ```
-This process can take a while depending on your internet connection speed as it needs to download all dependencies. Once you see a log that says `Express server listening on http://localhost:3500`, you can access Thimble on [http://localhost:3500](http://localhost:3500).
 
-You can now make changes to the Thimble source code on your system and they should be automatically reflected on [http://localhost:3500](http://localhost:3500).
+Depending on your internet connection speed, this process can take a while (Since it needs to download all dependencies not listed above).
```
gideonthomas: lowercase "s" in Since

README.md
```diff
@@ -19,53 +19,161 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
...
-To restart Thimble again, run:
 ```
 vagrant reload --provision
 ```
-This will take a much shorter time to setup compared to the `vagrant up` command.
+
+This will take less time to setup compared to the `vagrant up` command.
+
+##1b) Manual Installation
+You can also setup Thimble and it's needed components outside Vagrant and Virtualbox. This might be needed if you want to:
```
gideonthomas: "it's" should be "its"

README.md
```diff
@@ -19,53 +19,161 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
...
+- Host your own instance of Thimble
+- Cannot run virtualization software on your computer
+
+### Pre-Requisites for Manual Installation
```
gideonthomas: Prerequisites

README.md
```diff
@@ -19,53 +19,161 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
...
+In order for Thimble to be installed correctly, the following dependencies needs to be installed:
```
gideonthomas: "needs" should be "need"

README.md
```diff
@@ -19,53 +19,161 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
...
+
+- Node.js 4.x or above (see note below)
+  - * **Note:** The login.webmaker.org dependency needs a node version of 4.x only while all the other dependencies work with a node version of 4.x and above. We suggest installing [NVM](https://github.com/creationix/nvm) to allow the use of multiple versions of node.
```
gideonthomas: there's an extra asterisk here

README.md
```diff
@@ -19,53 +19,161 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
...
+
+### Manually Installing the Parts
+Please note: On Windows, use ``copy`` instead of ``cp``
+
+####Thimble
+* Fork and clone https://github.com/mozilla/thimble.mozilla.org
+* Run ``cp env.dist .env`` to create an environment file
+* Run ``npm install`` to install dependencies
+* Run ``npm start`` to start the server
+* Once everything is ready and running, Thimble will be available at [http://localhost:3500/](http://localhost:3500/)
```
gideonthomas: let's move this line after the `publish.webmaker.org` section as a general statement.

README.md
```diff
@@ -19,53 +19,161 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
...
+  * If this already exists, run ``rm -rf /usr/local/var/postgres`` to remove it
+* Run ``postgres -D /usr/local/var/postgres`` to start the PostgreSQL server
+* Run ``createdb publish`` to create the Publish database
+
+####publish.webmaker.org
+* These steps assume you've followed the PostgreSQL steps above, including creating the publish database.
+* Clone https://github.com/mozilla/publish.webmaker.org
+* Run ``npm install`` to install dependencies
+* Run ``npm run env``
+* Run ``npm install knex -g`` to install knex
```
gideonthomas: Let's remove this since this is actually not required

README.md
```diff
@@ -19,53 +19,161 @@ how it works in [this blog post](http://blog.humphd.org/thimble-and-bramble/).
...
+####Sign In
+
+To publish locally, you'll need an account.
+* Go to [http://localhost:3000/account](http://localhost:3000/account)
+* Click ``Join Webmaker`` and complete the process, you can use a fake email
+* When you've created your account, click ``Set permanent password instead``
+  * This lets you authenticate your account without needing email
+* Go back to Thimble and Log In with your new account
+
+# 2) Running the parts
```
gideonthomas: I would delete this whole section because it seems redundant once you make the changes I suggest above

Pomax:
Like I said before - long README.md files are perfectly fine for complex projects, and with five services all running at the same time Thimble absolutely qualifies as one. Cloning the repo only gives you the README.md, not any wikis, so keeping all the information a dev needs to get things up and running without needing to load external webpages is a good thing.

Documentation length is never the issue: clarity is. If the REAMDE.md makes sense to someone trying to find the information they need, then length is irrelevant. If, however, there is so much text in such a disorganized fashion that people can't find what they need to know, then even a short README.md can be useless.

simon66:
@gideonthomas Done 👍 
Let me know if there are any other changes that needs to be made :)

gideonthomas:
@simon66 thank you for taking initiative on this and patiently going through the multiple revisions of this document. I think it looks good to merge now and we will follow up with the manual install steps.

simon66:
Thanks everyone for the positive feedback :) Glad I was able to help.
The install instruction is much better now 👍

Pomax:
@gideonthomas of course the best role model behaviour here would be to then say "and we've filed the follow-up issue, it's #......, you're welcome to help us out on that one of course" =)

PR #1687, opened 2017-02-01
Fix #1534 - Publish button clicking

### LINKED ISSUE #1534: "publish" button frequently does nothing unless I slow-click down and back up

cgsingh:
publishbtntest

**:after** Pseudo-elements were added to userbar.less. This solved the top-edge issue. I followed https://css-tricks.com/video-screencasts/123-if-it-moves-when-you-click-make-something-stick/ as a guide.

You can hopefully see what I did here: http://codepen.io/cgsingh/pen/JEpBgL.

### DIFF
--- a/public/resources/stylesheets/userbar.less
+++ b/public/resources/stylesheets/userbar.less
@@ -201,6 +201,22 @@ body {
       top: 2px;
       border: none;
     }
+
+    &:after {
+      content: "";
+      position: absolute;
+      top: 0;
+      left: 0;
+      right: 0;
+      bottom: -2px;
+    }
+
+    &:active:after {
+      top: -3px;
+      left: 0;
+      right: 0;
+      bottom: 1px;
+    }
   }
 
   #navbar-login {


flukeout:
Setting up my environment and reviewin'

public/resources/stylesheets/userbar.less
```diff
@@ -201,6 +201,22 @@ body {
       top: 2px;
       border: none;
     }
+
+    &:after {
+      content: "";
+      position: absolute;
+      top: 0px;
```
flukeout: Please drop the "px" from any "0" value - it's not needed.

flukeout:
@cgsingh - hey this is great, just tested and it works fine both when signed in and signed out. Solid work. If you could fix the small nitpick regarding the "0px" values, then we can merge this.

humphd:
@cgsingh nice work on this.

cgsingh:
@flukeout Thanks! I haven't entirely worked with CSS before so I had no idea the conventions and such, though I see it wasn't consistent with the rest of the file now. It should be changed now.

gideonthomas:
@cgsingh thanks for the contribution 👍 

And thanks for the review @flukeout!

PR #1692, opened 2017-02-01
Fix #1691 - Handle extra slashes in url construction during localization

### LINKED ISSUE #1691: Trouble with Dev enviornment
I've got the brackets repo running on :8000 and the vagrant up version of Thimble running too.

I can load the homepage and the editor, but when I click Sign In or Create an Account, I get an Oops page...

image

There's a double slash in the URL when I click these links, might that be the issue?

``http://localhost:3500/en-US//callback?code=22ed3d7f600...``

Is there an alternate process for setting up an account locally so that I can test publishing, or is this supposed to work.

cc @gideonthomas

gideonthomas:
Safe way to handle slashes while forming the url during localization

### DIFF
--- a/server/localize.js
+++ b/server/localize.js
@@ -41,7 +41,7 @@ module.exports = function localize(server, options) {
       localizedUrl = orgUrl.replace(urlLocale, bestLanguage);
     } else {
       // This regex makes sure that orgUrl always begins with `/` and inserts one if it doesn't
-      localizedUrl = "/" + bestLanguage + orgUrl.replace(/^\/?/, "/");
+      localizedUrl = "/" + bestLanguage + orgUrl.replace(/^\/*/, "/");
     }
 
     res.redirect(307, localizedUrl);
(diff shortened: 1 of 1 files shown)


gideonthomas:
@flukeout or @simon66 can you try testing this out?

simon66:
Hi @gideonthomas,
Its working for me :+1:

flukeout:
Same here @gideonthomas

server/localize.js
```diff
@@ -40,7 +40,8 @@ module.exports = function localize(server, options) {
     if(knownLocales.indexOf(urlLocale) !== -1) {
       localizedUrl = orgUrl.replace(urlLocale, bestLanguage);
     } else {
-      localizedUrl = "/" + bestLanguage + "/" + (orgUrl === "/" ? "" : orgUrl);
+      // This regex makes sure that orgUrl always begins with `/` and inserts one if it doesn't
+      localizedUrl = "/" + bestLanguage + orgUrl.replace(/^\/?/, "/");
```
Pomax: possible nit: I don't know if line 38 is normalized or not. If you request `//en-US` for instance we want to end up with a single `/` so if it's nor normalized we probably want `^\/*` rather than `^\/?`

humphd:
@Pomax...literally reading my mind as I type

server/localize.js
gideonthomas: thanks! Will make the change

PR #1697, opened 2017-02-03
Fix #1635 - Fixed logged in user dropdown menu easier to click

### LINKED ISSUE #1635: Make items in user dropdown menu easier to click.
The items in the dropdown menus only work if you click the text of the item, and not the entire length and height of the highlighted item when you hover.

This is because each item is a...

```
<li>
  <a>Item Label</a>
</li>
```

...and the padding around the item is applied to the ``<li>`` which isn't clickable.

image

To fix these we either...

* Add the click listener to the ``<li>`` item, or..
* Change the CSS styling so that the ``<a>`` item gets the padding and then makes up the entire clickable area.

raygervais:
This resolve #1635  the issue found [here](https://github.com/mozilla/thimble.mozilla.org/issues/1635) which made the clickable padding of the logged in user menu not register. To remedy, I increased the padding of the link itself and decreased the padding of the `<li>` element which surrounded it.

### DIFF
--- a/public/resources/stylesheets/userbar.less
+++ b/public/resources/stylesheets/userbar.less
@@ -476,7 +476,6 @@ a.navbar-login:hover {
       margin: 0;
 
       li {
-        display: block;
         padding: 0;
 
         &:last-child {
--- a/public/resources/stylesheets/userbar.less
+++ b/public/resources/stylesheets/userbar.less
@@ -477,7 +477,7 @@ a.navbar-login:hover {
 
       li {
         display: block;
-        padding: 0px;
+        padding: 0;
 
         &:last-child {
           border-top: 1px solid #e9e9e9;
--- a/public/resources/stylesheets/userbar.less
+++ b/public/resources/stylesheets/userbar.less
@@ -477,7 +477,7 @@ a.navbar-login:hover {
 
       li {
         display: block;
-        padding: 12px;
+        padding: 0px;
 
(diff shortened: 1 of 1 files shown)


public/resources/stylesheets/userbar.less
```diff
@@ -477,7 +477,7 @@ a.navbar-login:hover {
 
       li {
         display: block;
-        padding: 12px;
+        padding: 0px;
```
gideonthomas: off-hand, let's change this to `0` and remove the `px`

raygervais: I've fixed the resulting code as per your requested change. If there's anything else let me know

flukeout:
I'll test it out today.

public/resources/stylesheets/userbar.less
```diff
@@ -477,7 +477,7 @@ a.navbar-login:hover {
 
       li {
         display: block;
```
flukeout: Please remove the ``display: block`` declaration. ``<li>`` items are by definition block-level elements, so we can drop this.

raygervais:
@flukeout, I've fixed your requested change. If you need anything else, let me know.

flukeout:
Tested & works - nice work @raygervais

raygervais:
Thank you @flukeout, I look forward to contributing more to Thimble.

PR #1698, opened 2017-02-03
Fix #1001: Changed text for deleteProjectConfirmText

### LINKED ISSUE #1001: The delete button is scary, because I don't know what it'll do
Creating a project creates a new thimble project, editing files autosaves my project, publishing a project takes what I've written and puts it on the web.

However, deleting does not seem to be this granular, so I'm afraid to press the delete button. Is it going to delete my project _and_ my web-published data? Will I be asked which of the two, or both, I want deleted?

The delete button is scary...

badrmodoukh:
I managed to change the text for the confirm delete project button

### DIFF
--- a/locales/en-US/messages.properties
+++ b/locales/en-US/messages.properties
@@ -136,7 +136,7 @@ newProjectLink=New Project
 signOutLink=Sign out
 publishBtn=Publish
 untitledProject=Untitled Project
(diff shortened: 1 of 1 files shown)


gideonthomas:
@badrmodoukh this works great! Thank for the contribution!

PR #1704, opened 2017-02-07
added url to help page

evantsyferov:


views/editor/bramble.html
```diff
@@ -121,6 +121,7 @@
         <div class="error-message message">
             <p>{{ gettext("errorLoadingProject") | safe }}</p>
             <p>{{ gettext("errorLoadingProjectSuggestion") }}</p>
+	    <a href="https://support.mozilla.org/en-US/products/webmaker/thimble"> get help here </a>
```
humphd: You need to add 1 extra space in front of the `<a href...` text so that it aligns with lines 122, 123.

Also, notice in lines 122 and 123 the use of the `{{...}}` syntax, which is due to the fact that these strings are localized (i.e., translated).  You need to add your string to the same file where `errorLoadingProject` and `errorLoadingProjectSuggestion` are defined (i.e., in the `en-US` locale file), and also add a new key for your string, maybe `errorLoadingProjectHelp`.

PR #1713, opened 2017-02-08
Update login and publish subtrees

gideonthomas:
This patch takes in the new commits from login.webmaker.org and publish.webmaker.org


humphd:
This looks sane to me.  Can I ask that you add a `README.md` to `services/` to explain that those are subtrees, and what steps a dev needs to take when updating things?  It would be good to document that somewhere other than your head :)

PR #1714, opened 2017-02-08
[WIP] Update node to LTS

gideonthomas:
WIP

Depends on https://github.com/mozilla/thimble.mozilla.org/pull/1713 landing first.

This updates node to use 6.9.x which is going to be LTS for more than a year. This is something we need to do for most of our apps.

Currently blocked because of bcrypt not finding its bindings.
It also kinda sucks that we have to run `npm install` twice for each service because of a bug in `npm` - https://github.com/npm/npm/issues/10776#issuecomment-209638410

### DIFF
--- a/scripts/setup-services.sh
+++ b/scripts/setup-services.sh
@@ -7,7 +7,7 @@ DB_PASSWORD="thimble"
 
 # --- Install OS level dependencies ---
 apt-get update -y
-curl -sL https://deb.nodesource.com/setup_4.x | sh
+curl -sL https://deb.nodesource.com/setup_6.x | sh
 apt-get install -y nodejs
 # --- Postgres ---
 echo "deb http://apt.postgresql.org/pub/repos/apt/ trusty-pgdg main" > /etc/apt/sources.list.d/pgdg.list
@@ -83,7 +83,8 @@ cd services
 echo "Setting up login.webmaker.org"
 cd login.webmaker.org
 cp env.sample .env
-sudo npm install --no-bin-links --loglevel=error # sudo needed for bcrypt permissions
+# Refer to https://github.com/npm/npm/issues/10776#issuecomment-209638410 on why we run npm install twice
+sudo npm install --no-bin-links --loglevel=error || sudo npm install --no-bin-links --loglevel=error # sudo needed for bcrypt permissions
 cd ..
 # ---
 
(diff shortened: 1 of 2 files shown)


Pomax:
It kind of looks like there's way more in this PR than the title implies.

gideonthomas:
It's a rebase issue. It has commits that are already in master but whose commit sha changed. I'll rebase and it should fix it.

Pomax:
ah, gotcha

Pomax:
#1713 landed - is this unblocked @gideonthomas?

gideonthomas:
I think it's still blocked on this: https://github.com/mozilla/login.webmaker.org/pull/387

ryanwarsaw:
@gideonthomas You should be good to go now.

scripts/setup-services.sh
```diff
@@ -83,15 +83,17 @@ cd services
 echo "Setting up login.webmaker.org"
 cd login.webmaker.org
 cp env.sample .env
-sudo npm install --no-bin-links --loglevel=error # sudo needed for bcrypt permissions
+# Refer to https://github.com/npm/npm/issues/10776#issuecomment-209638410 on why we run npm install twice
+sudo npm install --no-bin-links --loglevel=error || sudo npm install --no-bin-links --loglevel=error # sudo needed for bcrypt permissions
```
Pomax: I thought @ryanwarsaw switched us over to a js implementation of bcrypt. If so, we shouldn't need the sudo here

Pomax: Also, it would be good to check whether we still need to issue these commands twice, since we're on Node 6 now.

ryanwarsaw: @Pomax +1, although its worth noting we have other dependencies that still do require manual compilation (although usually pre-built binaries are served) so if that's the reasoning behind this I'd leave it in for good measure.

Pomax: given that we're vagranted: let's take out as much as we can. If it works in vagrant, that's a good reason to accept it.

gideonthomas: @Pomax the "double npm install needed" is still an issue with node 6 unfortunately. Although, I think instead of two, I will move it to a looped npm install.

Pomax: dang

flukeout:
@gideonthomas to change to a loop and then we can land land land.

PR #1717, opened 2017-02-09
Fix #719 - Renaming already published project brings up 'Update publi…

### LINKED ISSUE #719: Renaming an already published project doesn't bring up the Update published project option, but it should.
I just renamed my project from "boings" to "joings" yet the project list shows that the project was edited 33 minutes ago. Let's count a project rename as an edit.

image

lkisac:
This addresses issue #719.

After renaming and saving an already published project, the user is able to see the "Update published version" button.

### DIFF
--- a/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
+++ b/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
@@ -340,7 +340,7 @@ define(function(require) {
       publisher.init(bramble);
 
       // Initialize the project name UI
-      ProjectRenameUtility.init(appUrl, csrfToken, publisher);
+      ProjectRenameUtility.init(appUrl, csrfToken, publisher, bramble);
     } else {
       $("#navbar-publish-button").click(showPublishHelper);
     }
--- a/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
+++ b/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
@@ -1,6 +1,7 @@
 define(function(require) {
   var $ = require("jquery");
   var Publisher = require("fc/publisher");
+  var ProjectRenameUtility = require("fc/project-rename");
   var KeyHandler = require("fc/bramble-keyhandler");
   var BrambleMenus = require("fc/bramble-menus");
   var Underlay = require("fc/bramble-underlay");
@@ -16,7 +17,7 @@ define(function(require) {
     $(".preview-pane-nav").width(data.secondPaneWidth);
   }
 
-  function init(bramble, csrfToken, appUrl, ProjectRenameUtility) {
+  function init(bramble, csrfToken, appUrl) {
     var publisher;
     var locale = $("html")[0].lang;
 
@@ -340,7 +341,7 @@ define(function(require) {
       publisher.init(bramble);
 
       // Initialize the project name UI
-      ProjectRenameUtility.init(appUrl, csrfToken, publisher, bramble);
+      ProjectRenameUtility.init(appUrl, csrfToken, publisher);
     } else {
       $("#navbar-publish-button").click(showPublishHelper);
     }
--- a/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
+++ b/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
@@ -16,7 +16,7 @@ define(function(require) {
     $(".preview-pane-nav").width(data.secondPaneWidth);
   }
 
-  function init(bramble) {
+  function init(bramble, csrfToken, appUrl, ProjectRenameUtility) {
     var publisher;
     var locale = $("html")[0].lang;
 
@@ -338,6 +338,9 @@ define(function(require) {
 
       publisher = new Publisher();
       publisher.init(bramble);
+
+      // Initialize the project name UI
+      ProjectRenameUtility.init(appUrl, csrfToken, publisher);
     } else {
       $("#navbar-publish-button").click(showPublishHelper);
     }
--- a/public/editor/scripts/editor/js/fc/project-rename.js
+++ b/public/editor/scripts/editor/js/fc/project-rename.js
@@ -1,5 +1,6 @@
 define(function(require) {
   var $ = require("jquery");
+  var Publisher = require("fc/publisher");
   var InputField = require("fc/bramble-input-field");
   var KeyHandler = require("fc/bramble-keyhandler");
   var Project = require("project/project");
@@ -132,15 +133,20 @@ define(function(require) {
         }
         editingComplete(context);
         analytics.event("ProjectRenamed");
+
+        var publisher = new Publisher();
+        publisher.init(bramble);
+        publisher.showUnpublishedChangesPrompt();
       });
     });
   }
 
-  function ProjectRenameUtility(appUrl, csrfToken) {
+  function ProjectRenameUtility(appUrl, csrfToken, publisher) {
     var context = this;
 
     this.appUrl = appUrl;
     this.csrfToken = csrfToken;
+    this.publisher = publisher;
     this.container = $("#navbar-project-title");
     this.saveButton = $("#project-rename-save");
     this.renameButton = $("#navbar-rename-project");
@@ -156,8 +162,8 @@ define(function(require) {
     });
   }
 
-  function init(appUrl, csrfToken) {
-    return new ProjectRenameUtility(appUrl, csrfToken);
+  function init(appUrl, csrfToken, publisher) {
+    return new ProjectRenameUtility(appUrl, csrfToken, publisher);
   }
 
   return {
--- a/public/editor/scripts/editor/js/fc/project-rename.js
+++ b/public/editor/scripts/editor/js/fc/project-rename.js
@@ -30,7 +30,7 @@ define(function(require) {
       return false;
     }
 
-    saveButton.text("Save");
+    saveButton.text("{{ renameProjectSaveBtn }}");
     saveButton[isSave ? "hide" : "show"]();
     context.renameButton[isSave ? "show" : "hide"]();
     titleBar[isSave ? "save" : "edit"]();
@@ -118,7 +118,7 @@ define(function(require) {
   }
 
   function save(context) {
-    context.saveButton.text("Saving...");
+    context.saveButton.text("{{ renameProjectSavingIndicator }}");
 
     persist.call(context, context.titleBar.val(), function(err) {
       if(err) {
--- a/public/editor/scripts/editor/js/fc/project-rename.js
+++ b/public/editor/scripts/editor/js/fc/project-rename.js
@@ -135,18 +135,19 @@ define(function(require) {
         analytics.event("ProjectRenamed");
 
         var publisher = new Publisher();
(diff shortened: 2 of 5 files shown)


public/editor/scripts/editor/js/fc/project-rename.js
```diff
@@ -132,15 +133,20 @@ define(function(require) {
         }
         editingComplete(context);
         analytics.event("ProjectRenamed");
+
+        var publisher = new Publisher();
+        publisher.init(bramble);
```
humphd: This is failing on Travis:

```
   public/editor/scripts/editor/js/fc/project-rename.js

    138 |        publisher.init(bramble);
                                ^ 'bramble' is not defined.
```

The `bramble` object will have to get passed in here somehow.

lkisac:
I noticed something else while testing.  If the user leaves the project name unchanged and hits the "Save" button, the "Update published version" button will still show.  Which of these two options would be more suitable for handling this?

1. Hide save button if project name text box value matches previous one.  Show if it's different.
2. Do not show "Publish updated version" button if user hits save button and project name remains the same

Please let me know any other suggestions and if this should be included in the pull request.

public/editor/scripts/editor/js/fc/project-rename.js
lkisac: Thanks @humphd.  I fixed with latest commit

public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
```diff
@@ -338,6 +338,9 @@ define(function(require) {
 
       publisher = new Publisher();
       publisher.init(bramble);
+
+      // Initialize the project name UI
+      ProjectRenameUtility.init(appUrl, csrfToken, publisher, bramble);
```
gideonthomas: I think maybe we should maybe have the order be: `bramble, appUrl, csrfToken, publisher`

public/editor/scripts/editor/js/fc/project-rename.js
```diff
@@ -132,15 +133,21 @@ define(function(require) {
         }
         editingComplete(context);
         analytics.event("ProjectRenamed");
+
+        var publisher = new Publisher();
```
gideonthomas: You shouldn't create a new instance of the publisher here. You already have access to a publisher instance in `this` (or out here it's `context`) so you should use that instead.

public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
gideonthomas: Actually, you shouldn't need to pass in `bramble` at all.

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -62,29 +62,10 @@ define(function(require) {
...
-          return;
-        }
-        publisher.needsUpdate = true;
-      });
-    }
-    bramble.on("fileChange", handleFileEvent);
-    bramble.on("fileDelete", handleFileEvent);
-    bramble.on("fileRename", handleFileEvent);
-    bramble.on("folderRename", handleFileEvent);
+    bramble.on("fileChange", publisher.showUnpublishedChangesPrompt);
```
gideonthomas: you'll need to bind this to the publisher instance: `publisher.showUnpublishedChangesPrompt.bind(publisher)`

### Explanation
In javascript, when you pass a function (`showUnpublishedChangesPrompt`) that is bound to an instance (i.e. attached to the prototype of a constructor) (in this case `publisher`) as a parameter when calling another function, it will no longer be bound to that instance (`showUnpublishedChangesPrompt` will not be bound to `publisher`) and so, when the function (`showUnpublishedChangesPrompt`) is called, inside it, `this` will be `undefined`. The javascript `bind` function forces the parameter that you pass in to be bound to a particular instance. You can read a little bit more about it [here](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_objects/Function/bind)

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -103,6 +84,28 @@ define(function(require) {
     });
   };
 
+  // When there are file events triggered by the editor (or project is renamed), update the publish/republish status
+  Publisher.prototype.showUnpublishedChangesPrompt = function() {
```
gideonthomas: I would also add an optional `callback` parameter here.

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -103,6 +84,28 @@ define(function(require) {
...
+    var publisher = this;
+
```
gideonthomas: To make sure that the callback parameter passed in is truly optional, let's set a default value for it here.
`callback = callback || function() {}`

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -103,6 +84,28 @@ define(function(require) {
...
+    if(publisher.needsUpdate) {
+      return;
```
gideonthomas: call the callback here

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -103,6 +84,28 @@ define(function(require) {
...
+    }
+
+    if(!Project.getPublishUrl()) {
+      return;
```
gideonthomas: call the callback here

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -103,6 +84,28 @@ define(function(require) {
...
+    }
+
+    publisher.handlers.unpublishedChangesPrompt();
+    Project.publishNeedsUpdate(true, function(err) {
+      if(err) {
+        console.error("[Thimble] Failed to set the publishNeedsUpdate flag after a file change: ", err);
+        return;
```
gideonthomas: call the callback here (and pass err into it)

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -103,6 +84,28 @@ define(function(require) {
...
+      }
+      publisher.needsUpdate = true;
```
gideonthomas: call the callback after this line too

public/editor/scripts/editor/js/fc/publisher.js
lkisac: Thank you for your feedback @gideonthomas   I pushed a new commit with your recommended changes.  Please let me know if I missed anything or can improve anymore

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -51,7 +51,8 @@ define(function(require) {
     publisher.handlers = {
       publish: publisher.publish.bind(publisher),
       unpublish: publisher.unpublish.bind(publisher),
-      unpublishedChangesPrompt: unpublishedChangesPrompt.bind(publisher)
+      unpublishedChangesPrompt: unpublishedChangesPrompt.bind(publisher),
+      showUnpublishedChangesPrompt : publisher.showUnpublishedChangesPrompt.bind(publisher)
```
gideonthomas: you don't need this change here since we don't use it anywhere

public/editor/scripts/editor/js/fc/publisher.js
gideonthomas: I think I might not have been super clear here, I apologize for that! What I meant was this line (and subsequent ones) should be:
```js
bramble.on("fileChange", publisher.showUnpublishedChangesPrompt.bind(publisher));
bramble.on("fileDelete", publisher.showUnpublishedChangesPrompt.bind(publisher));
...
```

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -103,6 +85,31 @@ define(function(require) {
...
   };
 
+  // When there are file events triggered by the editor (or project is renamed), update the publish/republish status
+  Publisher.prototype.showUnpublishedChangesPrompt = function(callback) {
+    var publisher = this;
+
+    callback = callback || function() {};
+
+    if(publisher.needsUpdate) {
+      callback();
```
gideonthomas: you'll need to return here so that nothing in this function is run after the callback is called. Change this to:
```js
callback();
return;
```

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -103,6 +85,31 @@ define(function(require) {
...
+    }
+
+    if(!Project.getPublishUrl()) {
+      callback();
```
gideonthomas: same as above

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -103,6 +85,31 @@ define(function(require) {
...
+    }
+
+    publisher.handlers.unpublishedChangesPrompt();
+    Project.publishNeedsUpdate(true, function(err) {
+      if(err) {
+        console.error("[Thimble] Failed to set the publishNeedsUpdate flag after a file change: ", err);
+        callback(err);
```
gideonthomas: same here:
```js
callback(err);
return;
```

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -103,6 +85,31 @@ define(function(require) {
...
+      }
+      publisher.needsUpdate = true;
```
gideonthomas: add a blank line before this line

public/editor/scripts/main.js
```diff
@@ -52,7 +52,7 @@ require(["jquery", "bowser"], function($, bowser) {
 
   Bramble.once("error", onError);
 
-  function init(BrambleEditor, Project, SSOOverride, ProjectRenameUtility) {
+  function init(BrambleEditor, Project, SSOOverride) {
```
gideonthomas: now that you've removed `ProjectRenameUtility` here, you should also remove `fc/project-rename` from the array on line 79 since we don't use it anymore.

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -51,7 +51,7 @@ define(function(require) {
     publisher.handlers = {
       publish: publisher.publish.bind(publisher),
       unpublish: publisher.unpublish.bind(publisher),
-      unpublishedChangesPrompt: unpublishedChangesPrompt.bind(publisher)
+      unpublishedChangesPrompt: unpublishedChangesPrompt.bind(publisher),
```
gideonthomas: remove the comma at the end

gideonthomas:
Darn, looks like I was partly wrong. This patch does involve `publish.webmaker.org` to be fixed as well. I filed https://github.com/mozilla/publish.webmaker.org/issues/240 so that when you click "Update published project" on a rename, the published project's name is actually updated.

Regardless, this patch itself looks right and needs to land anyway. Thank you very much for the contribution @lkisac!

lkisac:
@gideonthomas Glad to contribute to this project.  Thanks for the help.  I will keep an eye out on the recent [#240](https://github.com/mozilla/publish.webmaker.org/issues/240) issue you filed from this patch - might be something I'd also be interested in working on

Issue #1608, opened 2017-01-12
Thimble and WebVR: fix remix overlay issue with WebVR toggle


humphd:
We have a bug in published projects, in both orientations on mobile, where the Remix button blocks the VR button

![](https://i.imgur.com/1xrHqpP.png)

gideonthomas:
Possible solutions from previous discussions: https://github.com/mozilla/thimble.mozilla.org/issues/1515

humphd:
NOTE: the code that actually injects this remix button, and places is it is in https://github.com/mozilla/thimble.mozilla.org/blob/master/public/resources/remix/index.js and https://github.com/mozilla/thimble.mozilla.org/blob/master/public/resources/remix/style.less

ZenanZha:
I got where the problem is! I thing I think I could fix it right away. Thank god.

ZenanZha:
I think we could just move the remix button a little bit up(from bottom: 6px to 26px), so both button could works at the same time.
In this way, the remix button will be more friendly for a longer smart phone(like, iphone 7) user to press.

gideonthomas:
@flukeout what do you think of this solution?

flukeout:
Work has moved to #1724

Rajat-dhyani:
i followed  [#1724](https://github.com/mozilla/thimble.mozilla.org/pull/1724) but i guess @ZenanZha  is not really following up... Can i take over the work if no body is working on it ?

ZenanZha:
@Rajat-dhyani  Sorry, man. I just spent some time on study CSS on my own. I'm about to  finish my work before Friday.

Rajat-dhyani:
okay no problem.. @ZenanZha i will find some other issue to work upon.... thanx for replying anyway 👍

flukeout:
Closed via work in #1724 - thanks again @ZenanZha

PR #1724, opened 2017-02-10
The solution of Issue1608 - Remix button position in A-frame project

ZenanZha:
Fix a bug that in the published VR website, the remix button will cover the VR button on the left bottom corner.[#1608](https://github.com/mozilla/thimble.mozilla.org/issues/1608)


ZenanZha:
thimble_firstbug

flukeout:
Hi @ZenanZha - thanks for submitting. I don't want to change the position for ALL projects, but it seems like there might be a way to just target ``a-frame`` projects. Seems like the body gets a special class..

image

Can you write some styles that just apply to A-Frame projects based on this selector? Then maybe you can move the Thimble button to the bottom left?

ZenanZha:
@flukeout Thank you for your suggestion. I thought about it and I fully agree with this idea. I will try to fix the bug on this way.

humphd:
@ZenanZha this needs follow-up, can you finish this PR please?  If you're unclear what @flukeout is asking above, maybe ask some questions here.  Thanks.

flukeout:
Ping @ZenanZha

ZenanZha:
@flukeout What about this one? It's works under "a-body" class for AFrame projects only.

ZenanZha:
Sorry for not squash my commit messages. Can someone tell me how I can rebase everything after I push my branch up?
@humphd

flukeout:
@ZenanZha testing out your changes now. I also have not been able to squash commit messages on my local environment via the usual ``git rebase master -i`` - but I don't think it matters because Github can now squash them automatically when we merge.

public/resources/remix/style.less
```diff
@@ -272,3 +272,21 @@
...
+    
+  .details-bar-remix-button-wrapper {
+    margin-right: auto !important;
+    margin-left: 46px !important;
+  }
+
+  .close-details-bar {
+    left: 10px !important;
+  }
+}
```
flukeout: Please add a line break after the last line of code here.

flukeout:
Hey @ZenanZha - one minor fix requested here and we're good to go. Just add a line-break at the end of the last line. Thanks!

public/resources/remix/style.less
ZenanZha: I made the change but still have a question: Is this a good parctice that adding a line break at the end of the file? If it is, I would be really carefully with the fix that I'm doing right now. Thank you.

ZenanZha: @flukeout

public/resources/remix/style.less
```diff
@@ -272,3 +272,22 @@
     }
   }
 }
+/* move the remix button to the left side for AFrame projects only*/
+.a-body .details-bar.touch-mode {
+  right: auto !important;
+  left: 6px !important;
+    
+  .thimble-button {
+    left: 0px !important;
```
humphd: @ZenanZha for next time, a tip: when doing `0` in CSS don't put any units, that is: `0` (correct) vs. `0px` (incorrect).  It's not a huge deal, and not worth blocking this any further, but fyi.

public/resources/remix/style.less
humphd: For the benefit of others, having a `\n` newline at the end of the file vs. just having it end is done so that you avoid issues when files are concatenated and joined together at build time: you don't want the end of one file and the start of another to get interpreted as a single line in some cases.  Always having a line break on your last line is how we fix this.

flukeout:
Ok, thanks on your work on this @ZenanZha - glad we finally got it landed! 🎺

PR #1726, opened 2017-02-10
Fix #1627: Added ability to open a project in a new tab

### LINKED ISSUE #1627: Your Projects List > Add ability to open a project in a new tab.
Currently, the project opens inside the same browser tab as the Your Projects page, which is annoying if you just want to quickly check what it is.

We could either...
* Figure out while command+click doesn't open a new tab
* Add a separate "open in new tab" action

omytryniuk:
Now users should be able to open a project in a new tab.

pic1

In addition, since we use a regular link now, I see no point in keeping this [listener](https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/project-list.js#L37-L39).  Let me know please whether I am right.

### DIFF
--- a/public/editor/stylesheets/projects-style.less
+++ b/public/editor/stylesheets/projects-style.less
@@ -59,8 +59,8 @@ td.project-title {
   padding-left: 8px;
 }
 
-.project-title a{
-  color: #000000;
+.project-title a {
+  color: #000;
   text-decoration: none;
 }
 
--- a/public/editor/stylesheets/projects-style.less
+++ b/public/editor/stylesheets/projects-style.less
@@ -60,7 +60,7 @@ td.project-title {
 }
 
 .project-title a{
-  color:black;
+  color: #000000;
   text-decoration: none;
 }
 
--- a/public/editor/stylesheets/projects-style.less
+++ b/public/editor/stylesheets/projects-style.less
@@ -58,6 +58,12 @@ td.button-fill {
 td.project-title {
   padding-left: 8px;
 }
+
+.project-title a{
+  color:black;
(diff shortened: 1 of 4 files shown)


public/editor/stylesheets/projects-style.less
```diff
@@ -58,6 +58,12 @@ td.button-fill {
 td.project-title {
   padding-left: 8px;
 }
+
+.project-title a{
```
gideonthomas: space after `a`

public/editor/stylesheets/projects-style.less
```diff
@@ -58,6 +58,12 @@ td.button-fill {
...
+  color: #000000;
```
gideonthomas: let's change this to `#000`

views/editor/projects.html
```diff
@@ -27,8 +27,10 @@
         {% endfor %}
         >
           <td class="project-title">
-            {{ project.title }}
-            <span class="project-information"></span>
+            <a target="_self" href="/{{locale}}/user/{{username}}/{{project.id}}/{{queryString}}">
```
gideonthomas: `queryString` here is not passed in by default. You need to explicitly pass it in here: https://github.com/mozilla/thimble.mozilla.org/blob/master/server/routes/projects/read.js#L76 only then can you use it in this link

omytryniuk: You are right! I am really sorry about that @gideonthomas. I have made changes and it should work now.

gideonthomas:
Nice work @omytryniuk! Thanks for the contribution :)

PR #1727, opened 2017-02-10
fixes to the filesystem-sync.js, issue-1667

Taguiar:
I fixed the issue-1667. I had to change the position between the two lines since in some browsers the "Saving..." spam was not showing whenever you are logged in and had some changes to one of the files. 
I relate this problem to how some browsers handle HTML code differently and thus making changes like this necessary.


gideonthomas:
hmm, @humphd what do you think about this fix?

public/editor/scripts/editor/js/fc/filesystem-sync.js
```diff
@@ -36,8 +36,9 @@ define(function(require) {
 
     // Update the UI with a "Saving..." indicator whenever we sync a file
     syncManager.on("file-sync-start", function() {
+		$("#navbar-save-indicator").removeClass("hide");
```
humphd: Your indenting is too extreme here, it should line-up with what's on line 40.

humphd:
@gideonthomas I think if this fixes it, it's fine with me.  I'm surprised it matters, but cross-browser stuff could be involved.  It shouldn't hurt to do it in this order.

gideonthomas:
Ok, let's take this change, @Taguiar, after you fix the indent issue, I can merge this in. We can reopen the issue if it still happens.

public/editor/scripts/editor/js/fc/filesystem-sync.js
Pomax: as this is jQuery, we don't even need two lines here. This should do the trick too:
```
$("#navbar-save-indicator").removeClass("hide").text("{{ fileSavingIndicator }}");
```
However it is puzzling that this ordering matters, and suggests that maybe this fixes a symptom, not the underlying problem. I would caution landing this before we know *why* it works. And once we do know why, this code should get a little code comment to explain the ordering and warn about changing that ordering.

Taguiar: so I should create a new pull request or I just close and reopen this one??

Taguiar: Ok, I will make some comments for future references!

flukeout: You can just make the change and push the changes to your branch. Your PR will update automatically.

gideonthomas:
@Taguiar how are you doing with this PR?

Taguiar:
@gideonthomas  I am pretty much done just wanted to run on other computers to see if I got it for good!

Pomax:
@Taguiar what would you like to do here, that's a lot closing and opening =)

public/editor/scripts/editor/js/fc/filesystem-sync.js
```diff
@@ -36,8 +36,8 @@ define(function(require) {
 
     // Update the UI with a "Saving..." indicator whenever we sync a file
     syncManager.on("file-sync-start", function() {
+	  $("#navbar-save-indicator").removeClass("hide");
       $("#navbar-save-indicator").text("{{ fileSavingIndicator }}");
```
Taguiar: don't know why my editor is messing up with my indentation! it looks perfect om my machine, but whenever I add it pushes line 39 back!

Pomax: tabs. You have a tab *and* spaces on this line, so do a regexp replace to turn `\t` into `    ` (four spaces) in this file.

Taguiar: Sorry the ignorance, but how should I do that?

Pomax: which code editor do you use? most let you say which indentation rules to use. At the very least they should has some kind of find/replace that lets you search for tabs (typically `\t`) and then replace it another string like `    ` (four spaces)

Pomax:
Thanks for the work on this @Taguiar, and hopefully indentation won't give any more problems in the future! Most editors will let you "force" which indentation style to use, so for a next file edit it might be worth making sure that's set to the right thing so that even if you do accidentally start using tabs vs spaces, the editor will just convert them to the right thing on the fly for you.

PR #1730, opened 2017-02-11
fix Issue1665: adding link to help page on loading project page

evantsyferov:


views/editor/bramble.html
```diff
@@ -117,10 +117,13 @@
 
     <div id="spinner-container">
         <div class="thimble-spinner"></div>
-        <div class="loading-message message">{{ gettext("loadingProject") }}</div>
+        <div class="loading-message message">{{ gettext("loadingProject") }}
```
gideonthomas: Put the `{{ gettext("loadingProject") }}` on a new line

views/editor/bramble.html
```diff
@@ -117,10 +117,13 @@
...
+	<p>{{ gettext("loading problem") }} <button type="button"><a href="https://support.mozilla.org/products/webmaker/thimble">{{ gettext("get help here") }}</button></a></p>
```
gideonthomas: The indent on this should be 4 spaces.

gideonthomas: Let's keep it simple and remove the button, let's just make it a simple anchor and we can style it if needed.

gideonthomas: `loading problem` is probably not the best name for the key (esp. since it has a space in it). Maybe something like `loadingProjectSupportSuggestion`? You would also need an entry for it somewhere [here](https://github.com/mozilla/thimble.mozilla.org/blob/master/locales/en-US/messages.properties#L75-L80). Add an entry with whatever text you think might be appropriate for now and we can suggest changes. Also, I think you should enter the entire contents of the paragraph as one localization string i.e. something like `loadingProjectSupportSuggestion=Blah blah blah <a href="...">Thimble help</a>.`

gideonthomas: Since there will be html in the localized string, your `gettext` query will also change a bit. You will need to write something like `{{ gettext("whateverKeyYouDecide") | safe }}`

views/editor/bramble.html
```diff
@@ -117,10 +117,13 @@
...
+	</div>
         <div class="error-message message">
             <p>{{ gettext("errorLoadingProject") | safe }}</p>
             <p>{{ gettext("errorLoadingProjectSuggestion") }}</p>
+
```
gideonthomas: undo this change.

views/editor/bramble.html
evantsyferov: I put the link on a button because the purple color of the link made the text somewhat hard to read due to the dark green background. When every I made changes to color of the link the loading project page would not show for me.

gideonthomas: @evantsyferov try removing the button for now. I can help you figure out styling for this after that and @flukeout can help out as well

flukeout:
Hi @evantsyferov could you please update the message copy...

```
If it doesn't load, <a title="Thimble help page" href="https://support.mozilla.org/en-Us/products/webmaker/thimble">get help here</a>.
```

I don't want to make it sound like you can get help making it load faster in case you think it's slow. Also, I'm in the process of adding a help page specific to this problem to the SUMO site. I think it should be ready tomorrow. I'll add the link to this issue when it's ready. Thanks!

@gideonthomas Here's the content I'll be adding to SUMO, are there any others steps we can recommend? https://github.com/mozilla/thimble.mozilla.org/wiki/Project-not-Loading-Help

gideonthomas:
@flukeout nope, what you have seems to be right - I would however presuppose "Reset your IndexedDB" with a possibility of some data loss.

Pomax:
@evantsyferov as a general "open source contribution" point of process, could you update this pull request title so that it's descriptive of what it's solving, with the "fixes #..." part as part of the PR comment? This is generally a good way to make it easy for people looking at the pull request list to see which PR addresses which issue, without having to open individual ones and clicking through to the issues they belong to.

flukeout:
Hi @evantsyferov a few changes I'd like to request:

* Make the link color white
* Give it an underline
* Remove it from the element that has the pulsing animation

Here's how it appears to me, the link is very hard to see...

image

evantsyferov:
@flukeout I have tried to change the color of the link before through several methods and then the loading project would stop showing up for me. Can you suggest where to change the color of link.

flukeout:
I'm running into the same problem @evantsyferov 

cc @humphd @gideonthomas Do you have any ideas why changing styles in ``views/editor/bramble.html`` would cause the loading spinner overlay to stop working? It's like if the styles get changed, they're no longer being applied at all.

humphd:
It's [CSP error](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy):

```
http://mozorg.cdn.mozilla.net https://ajax.googleapis.com https://fonts.googleapis.com https://mozorg.cdn.mozilla.net https://netdna.bootstrapcdn.com https://pontoon.mozilla.org 'sha256-jxjTomDIR9qe7wntK24mAd+gIoz39DrBll8o6DEBALs='". Either the 'unsafe-inline' keyword, a hash ('sha256-nQV06gnHc1BLLuuL9dLj/l31prp0ChLQ6XjgDJewYBw='), or a nonce ('nonce-...') is required to enable inline execution.
```

Basically, we have a CSP directive to allow a **specific** version of this inline stylesheet, and changing anything in it breaks the hash:

https://github.com/mozilla/thimble.mozilla.org/blob/master/server/security.js#L50-L51

You'd have to change those 2 lines to:

```js
    // Inline style for the spinner
    "'unsafe-inline'"
```

And then I guess when you're done with this patch, we should update the hash again and lock it down.  @gideonthomas will know the drill, since he did it before in https://github.com/mozilla/thimble.mozilla.org/commit/43a072036f985699118313b5c7663cf7ad2ef763

gideonthomas:
Yeah change that line to `'unsafe-inline'` and this should work! I'll fix the security bit for this later.

flukeout:
Did that work for you @evantsyferov ? Also, just use this link as the help link please:

https://github.com/mozilla/thimble.mozilla.org/wiki/Project-not-Loading-Help

evantsyferov:
Yes changing that line to " 'unsafe-inline' " worked for me

views/editor/bramble.html
```diff
@@ -36,6 +36,10 @@
             animation: loadingpulse 1s infinite linear;
             -webkit-animation: loadingpulse 1s infinite linear;
         }
+	#spinner-container .loading-message-help-link {
```
flukeout: Please use spaces instead of tabs here and match the indentation in the rest of the file, this block appears like this in my editor...

image

views/editor/bramble.html
```diff
@@ -113,15 +117,26 @@
                 transform: rotateX(-360deg) rotateY(-360deg);
             }
         }
+	
+	a {
```
flukeout: Here also, please use spaces instead of tabs. Also, nest this rule in ``#spinner-container`` like the rest of the rules.

views/editor/bramble.html
```diff
@@ -113,15 +117,26 @@
...
+		color: white;
+		text-decoration: underline;
+	}
+
         </style>
 
     <div id="spinner-container">
         <div class="thimble-spinner"></div>
-        <div class="loading-message message">{{ gettext("loadingProject") }}</div>
+        <div class="loading-message message">
```
flukeout: Also please fix indentation for this section.

locales/en-US/messages.properties
```diff
@@ -73,6 +73,7 @@ mySixWordSummerTeachingKit=Six-Word Summer {{ teachingKit }}
 
 # Loading the editor
 loadingProject=Loading Project
+loadingProjectSupportSuggestion=If the project doesn't load, <a title="Thimble help page" href="https://support.mozilla.org/en-Us/products/webmaker/thimble"> get help here </a>.
```
flukeout: Please remove the spaces inside of the ``<a>`` tag. The trailing space is visible on the loading screen and looks a little off...

image

flukeout:
Thanks @evantsyferov - I left some comments, let me know if you have any questions.

views/editor/bramble.html
```diff
@@ -37,6 +37,16 @@
             -webkit-animation: loadingpulse 1s infinite linear;
         }
 
+        #spinner-container .loading-message-help-link {
+            animation: none;
+            -webkit-animation: none;
```
humphd: Always put non-prefixed versions of CSS rules at the end of your list, so that the spec standard one gets used if it's available.

views/editor/bramble.html
```diff
@@ -113,15 +123,21 @@
...
+         {{ gettext("loadingProject") }}
+	<div class="loading-message-help-link">
+            <p>{{ gettext("loadingProjectSupportSuggestion") | safe }}</p>
+    	</div>
+	</div>
         <div class="error-message message">
             <p>{{ gettext("errorLoadingProject") | safe }}</p>
             <p>{{ gettext("errorLoadingProjectSuggestion") }}</p>
-        </div>
+	 </div>
```
gideonthomas: space vs. tab

views/editor/bramble.html
```diff
@@ -113,16 +123,22 @@
...
+            <p>{{ gettext("loadingProjectSupportSuggestion") | safe }}</p>
+    	  </div>
+	 </div>
         <div class="error-message message">
             <p>{{ gettext("errorLoadingProject") | safe }}</p>
             <p>{{ gettext("errorLoadingProjectSuggestion") }}</p>
-        </div>
-    </div>
+       </div>
+     </div>
```
flukeout: Hi @evantsyferov - please fix up the indentation in the block from line 129 to 141. Please use 2 space indentation, and not tab characters.

Here's how the code appears right now in my editor...

image

Notice that the indentation is a mix of spaces and tabs. Please match the indentation so that it looks similar to the ``id=browser-support-warning`` code block below. 

Thanks!

flukeout:
Just a few nitpicks about the indentation and we'll be good to go. Let me know if you have any questions!

humphd:
@flukeout, @gideonthomas looks like this is ready to look at again.

flukeout:
Right on - merging! Thanks for your work on this @evantsyferov - I think it will help a lot of people who get stuck.

Issue #1729, opened 2017-02-11
Locales not present when installing via vagrant


LCartwright:
I'm attempting to start up thimble for the first time from a fresh clone of master, after running `vagrant up`, requesting http://localhost:3500/ returns a connection reset error from the browser. 

So looking at the pm2 app summary in the `vagrant up` output, I see that `localize-client` is stopped
```bash
==> default: ┌─────────────────┬────┬──────┬──────┬─────────┬─────────┬────────┬─────┬───────────┬──────────┐
==> default: │ App name        │ id │ mode │ pid  │ status  │ restart │ uptime │ cpu │ mem       │ watching │
==> default: ├─────────────────┼────┼──────┼──────┼─────────┼─────────┼────────┼─────┼───────────┼──────────┤
==> default: │ id              │ 1  │ fork │ 2215 │ online  │ 0       │ 12s    │ 23% │ 27.3 MB   │ disabled │
==> default: │ localize-client │ 3  │ fork │ 2227 │ stopped │ 1       │ 0      │ 0%  │ 0 B       │ disabled │
==> default: │ login           │ 0  │ fork │ 2209 │ online  │ 0       │ 12s    │ 42% │ 21.0 MB   │ disabled │
==> default: │ publish         │ 2  │ fork │ 2221 │ online  │ 0       │ 12s    │ 23% │ 39.9 MB   │ disabled │
==> default: │ thimble         │ 4  │ fork │ 2235 │ online  │ 0       │ 12s    │ 20% │ 47.5 MB   │ enabled  │
==> default: └─────────────────┴────┴──────┴──────┴─────────┴─────────┴────────┴─────┴───────────┴──────────┘
```
Digging into the log file `/home/vagrant/.pm2/logs/localize-client-out.log` shows the following error
```
Error: Cannot find module '/vagrant/dist/locales/en_US/messages.json'
    at Function.Module._resolveFilename (module.js:325:15)
    at Function.Module._load (module.js:276:25)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/vagrant/scripts/localize-client.js:17:12)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
```
Likewise there's a similar error in `/home/vagrant/.pm2/logs/thimble-error.log`
[...]

PR #1734, opened 2017-02-12
Fix Issue#1729

LCartwright:
Here are the changes as described in #1729


gideonthomas:
@LCartwright this looks great! Thank you very much for taking the time to figure out what was going on.

Issue #1702, opened 2017-02-04
Thimble default projects code assumes only folders, crashes on presence of a file.


simon66:
I ran into this issue last week and today while trying to start vagrant:

![screen shot 2017-01-31 at 12 33 06 am](https://cloud.githubusercontent.com/assets/1103528/22621278/ec014e80-eaed-11e6-9a02-4737825e44f1.png)

Looks like the issue is caused by one of those pesky `.DS_Store` file (I hate these).
I tried the following:

`vagrant destroy`
`vagrant up`
`vagrant reload --provision`

and I still ran into this issue.
I found out that this file was created when I was browsing the `default` folder (`./thimble.mozilla.org/default/`)
Deleting the file fixed my issue :)

I'm not too sure which script is trying to load `.DS_Store` and I'm short on time to continue debugging.

PS: The issue it self is not caused by the .DS_Store properties. I can create any file in the `default` folder and vagrant will fail to start. Looks like the script is trying to access the file as if it was a directory.
![screen shot 2017-02-04 at 3 39 33 pm](https://cloud.githubusercontent.com/assets/1103528/22621371/2d0c991e-eaf0-11e6-8d9d-c343397301d9.png)

humphd:
It's actually a bug in this function:

https://github.com/mozilla/thimble.mozilla.org/blob/master/default/index.js#L131-L145

We're implicitly assuming that everything in the `default/*` directory is a dir vs. a file, with the exception of the `index.js` file.  We should probably do an `fs.stat` on each item in that loop to see if it's a dir or a file, and skip any files.

humphd:
I've updated the title of this issue to better reflect what's happening.

LCartwright:
I'm interested in fixing this bug.

humphd:
@LCartwright that's great, thank you!  If you have questions or need any help, please let us know.

LCartwright:
@humphd you can find the fix in PR #1736, thanks!

gideonthomas:
Fixed with 540835e8b5df7c2a717b5262e4938e06bedf74d8

PR #1736, opened 2017-02-12
Handle presence of files in the default project

LCartwright:
This PR is to solve #1702


humphd:
This looks right to me.  I'll let @gideonthomas sign off on it, and land if he's happy.  Thanks for the fix.

gideonthomas:
Yep, this looks good! Thanks @LCartwright, much appreciated!

Issue #1718, opened 2017-02-09
Make it harder to accidentally Unpublish a project.


flukeout:
In the Publish dialog, the Unpublish link is really close to the link the published project, which can lead to misclicks and unintended consequences.

<img width="250" src="https://cloud.githubusercontent.com/assets/25212/22798522/3e128d02-eeb7-11e6-85cc-131b6c3d68de.png"/>

It's especially bad because once you do this, you can't get the original URL back.

**Ideas**
* Move the Button further away from the project link
* Add an extra step to confirm an Unpublish action
  * Maybe even a boring old Confirm
  * Some kind of custom "Are you sure?"
* Change icon to Trash icon.
* Make the button red scary.
* Change label to "Delete" instead of "Unpublish"
  * Unpublish / republish kind of makes it seem like you might get the same URL

Also somewhere within that UX we should indicate that Unpublishing causes you to lose the original URL you had.

gideonthomas:
+1 to adding an extra step

O-kasso:
hey! can I take a crack at this?

humphd:
@flukeout can you let @O-kasso know your thoughts on this plan?

O-kasso:
I was thinking the confirmation dialog could say something like "Are you sure you want to unpublish this project? Links that have already been shared will no longer work."  

Also, does localization need to be taken into account? is there any material available describing how locale text is handled?

gideonthomas:
@O-kasso go for it! I would say go with those changes for now and open up a PR and we can comment on any changes there. As for localization: https://github.com/mozilla/thimble.mozilla.org/wiki/Localization might be helpful.

O-kasso:
Basic fix here: https://github.com/mozilla/thimble.mozilla.org/pull/1739
Will address localization separately.

TheAlbinoPanda:
Afternoon,

Is this issue solved? Can I take a stab at it ? I'm new to all of this, so If i'm doing something wrong, just let me know !

O-kasso:
@TheAlbinoPanda workin' on it :)

gideonthomas:
@TheAlbinoPanda if you look at the list in https://github.com/mozilla/thimble.mozilla.org/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20label%3A%22good%20first%20bug%22%20-label%3A%22assigned%20to%20contributor%22 you might have a better chance finding issues that haven't been taken by someone yet.

[...]

PR #1739, opened 2017-02-14
Make unpublishing scary

O-kasso:
Addresses #1718 

- Space unpublish button away from link
- Basic confirm dialog
- my attempt at (English) copy: "Are you sure you want to unpublish this project? Links that have already been shared will no longer work."
- Change icon to trash can
- Change button to red (the color guide doesn't have a dark red, so I just added a 10% darken for hover)
- Opted out of changing "Unpublish" to "Delete" for now, but open to making the switch

Things to address in later PRs:
- Figure out Localization strings
- Replace confirm dialog with prettier


public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -173,52 +173,54 @@ define(function(require) {
...
-    buttons.unpublish.off("click", handlers.unpublish);
-    setState(false);
-    SyncState.syncing();
-
-    var request = publisher.generateRequest("unpublish");
-    request.done(function() {
-      if(request.status !== 200) {
-        console.error("[Thimble] Server was unable to unpublish project, responded with status ", request.status);
-        buttons.unpublish.on("click", handlers.unpublish);
+    if (window.confirm("Are you sure you want to unpublish this project? Links that have already been shared will no longer work.")) {
```
gideonthomas: store the result of the confirm dialog in a variable and let's do something like:
```js 
if(!didConfirm) {
  return;
}

...
```

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -173,52 +173,54 @@ define(function(require) {
...
+      if (publisher.unpublishing) {
```
gideonthomas: This should be outside the confirm `if` check because we do not want to show the confirm dialog if it is already in the process of unpublishing

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -173,52 +173,54 @@ define(function(require) {
...
         return;
       }
+      publisher.unpublishing = true;
 
-      buttons.parent.removeClass("hide");
+      var setState = function(done) {
```
gideonthomas: let the name be bound to the function. No need for a variable.

O-kasso: agreed, only added it to avoid a declaration inside an if statement, which I've moved as per your above comments 👍

O-kasso:
@gideonthomas I believe I've implemented your feedback :). I've also added the new icon provided [here](https://github.com/mozilla/thimble.mozilla.org/issues/1718#issuecomment-280397148) to `public/resources/img/icon/` -- I noticed a similar folder in `dist/` but I'm not sure what that's for. Should it be copied there as well?

flukeout:
Hey @O-kasso checked out the PR, looks good! I've got some suggestions:

**For the garbage can icon, use**
* opacity: 1
* get rid of the ``width`` declaration
* make the ``height`` 16px
* add ``top: 1px`` 

**For the Button**
* Change label to "Delete Published Version"
* User the current, darker, ``:hover`` state as the ``:active`` state.
* For ``:hover`` just set the opacity of the button to .9

**For the dialog**
* Change copy to: "Are you sure? The current link will stop working and if you publish again, the link will be different."

Thanks a lot!

I think down the road, we'll want to move the Confirm copy into the dialog itself. But I'll have to work out the design of that first.

flukeout:
Related to my previous comment, here's what I was thinking...

image

Not needed for this issue. But I think it would be better to put the copy outside of the Confirm, since it's kind of hard to read there. In the confirm we can just do "Are you sure?".

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -173,6 +173,11 @@ define(function(require) {
     var dialog = publisher.dialog;
     var buttons = dialog.buttons;
 
+    var didConfirm = window.confirm("Are you sure you want to unpublish this project? Links that have already been shared will no longer work.");
+
+    if (!didConfirm) {
+      return;
+    }
     if (publisher.unpublishing) {
```
gideonthomas: Can you put this check before the `didConfirm` declaration so that we don't show it if it is already unpublishing.

flukeout:
Hey @O-kasso - I'm getting a double confirm message when running this patch.

double-message

Can you update it so that there is just one message (let's use the copy from the second one).

Thanks!

O-kasso:
@flukeout I'm unable to repro the double confirm dialogue. I've also tried grepping my branch for the erroneous string with no luck. Do you have the most up-to-date version of the branch?

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -186,6 +186,13 @@ define(function(require) {
...
       return;
     }
+
+    var didConfirm = window.confirm("Are you sure? The current link will stop working and if you publish again, the link will be different.");
```
Pomax: Should this string be localized? It feels like it should be.

gideonthomas: Ah, nice catch! @O-kasso, you can find information about how to localize this here: https://github.com/mozilla/thimble.mozilla.org/wiki/Localization#how-it-works and https://github.com/mozilla/thimble.mozilla.org/wiki/Localization#ui-strings

flukeout:
@O-kasso let me know if you have troubles with the localization part. I was working on some stuff that included localization so I'm happy to help.

O-kasso:
@flukeout @Pomax I've added localization for en-US and en-GB. What's the process for getting translations in other locales? Is it my responsibility to use Google Translate or something? Right now it'll just print an empty string.

public/editor/scripts/editor/js/fc/publisher.js
O-kasso: I believe this has been addressed.

flukeout:
Hey awesome @O-kasso it should use the English version as a default if it hasn't yet been localized.

Did you run ``npm run localize`` in your vagrant instance? That's what creates the other locale files. Then you can run ``npm run localize-client`` to uhhh, do something else - at least those are the steps that worked for me.

To run those commands you have to first...

``vagrant ssh``
`` cd /vagrant``

In any case, I'll try this patch out now!

flukeout:
Okay - looks like the Reviews have been addressed and this works as expected.

flukeout:
Hey @O-kasso I'm collecting information for a new contributors page for the Thimble website and I'd like to include you but need some more details. Can you please email me at ``luke@mozillafoundation.org``? Thanks!

Issue #1715, opened 2017-02-08
Update nunjucks in Brackets + Thimble


humphd:
While running `npm install` I notice:

```
npm WARN deprecated nunjucks@2.3.0: potential XSS vulnerability in autoescape mode, and with escape filter was fixed in v2.4.3
```
We should update to `2.4.3` in both code bases.

Blackweda:
I'd like to try this problem over the weekend :)

gideonthomas:
@Blackweda it's all yours :)

Blackweda:
thanks :)

PR #1742, opened 2017-02-16
Issue 1715

Blackweda:
Hello, I am making a fix to the issue #1715 .

This should solve the problem. If there are any problems with this, please let me know so I can fix it later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "multer": "^1.0.1",
     "node-uuid": "~1.4.3",
     "npm-run-all": "^1.5.1",
-    "nunjucks": "^2.4.3",
+    "nunjucks": "2.4.3",
     "properties-parser": "0.3.1",
     "q-io": "1.13.2",
     "request": "2.21.0",
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "multer": "^1.0.1",
     "node-uuid": "~1.4.3",
     "npm-run-all": "^1.5.1",
-    "nunjucks": "^2.3.0",
+    "nunjucks": "^2.4.3",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -62,7 +62,7 @@
     "multer": "^1.0.1",
     "node-uuid": "~1.4.3",
     "npm-run-all": "^1.5.1",
-    "nunjucks": "^2.3.0",
+    "nunjucks": "2.4.3",
```
Pomax: Hm, `^2.3.0` means "any version 2.3.0 or higher, as long as the major version number stays the same", so this should have already found 2.4.3 when doing a new `npm install` - Due to how semantic versioning works, where the first number is a "major release", the second number is the "minor" release, and the third number is a patching number, anything has the same major version should be API compatible, so we should be able to say `^2.4.3`, keeping the caret because newer versions in the `2.*.*` range should be compatible with the calls we make to nunjucks.

flukeout: @Pomax Any suggestions on what to change or do here?

Pomax: mostly just "add `^` back in". We should not need to peg the specific 2.4.0 version.

humphd:
@Blackweda when you do your pull requests, add more info to the description.  For example, instead of just "Issue 1715" do something like "Issue 1715: update nunjucks to latest" so that it's clear what this is about.

humphd:
@Blackweda any update on this 1 character fix that's blocking this from landing?  When is it coming?

Pomax:
as this is a single character fix with no activity for weeks, I'm going to merge this in with an immediate in-branch update to add in the caret.

Issue #1633, opened 2017-01-16
Preview auto-refresh setting gets reset when browser is refreshed.


flukeout:
If you uncheck it and refresh the browser, it becomes checked again. It should stay unchecked.

![image](https://cloud.githubusercontent.com/assets/25212/22001085/b291ccc4-dbf7-11e6-8ebd-5bbbd89c567b.png)

humphd:
Similar fix to what I describe in https://github.com/mozilla/thimble.mozilla.org/issues/1631#issuecomment-274916966

timmoy:
Hi, I'd like to take a stab at fixing this bug.

flukeout:
Sounds good @timmoy

timmoy:
Hi, I'd like to take a stab at fixing this bug.

timmoy:
After going over the code a bit, it seems that the problem is that when we refresh the browser the views/editor/nav-options.html reloads from scratch using 

 `<div class="refresh-wrapper enabled">`
        `<div title="{{ gettext("navRefreshPreviewIconTitle") }}" id="preview-pane-nav-refresh">`
          `<img src="/img/icon/refresh.svg">`
        `</div>`
       ` <div title="{{ gettext("navToggleAutoUpdateTitle") }}" class="toggle-auto-update">`
          `<span class="checkbox">`
            `<img class="checkbox-icon" src="/img/icon/checkmark-white.svg"/>`
         ` </span>`
         ` {{ gettext("navToggleAutoUpdateLabel") }}`
        `</div>`
`</div>`

The part I "enabled" in the first line is what I think is causing the issue. Whenever the user loads the page, the "enabled" class is loaded into the refresh-wrapper div again so the box becomes checked again. 

My current train of thought is that we need to somehow store the setting of whether the user wants it auto-refresh checked or not.
 
I looked through the code and found this snippet 

`<div id="navbar-login" class="{% if username %}signed-in{% endif %} nav-chunk">`

 in views/editor/userbar.html which was used to show the username if the user is signed in.

Potentially, if we edit the line I mentioned earlier into 

 `<div class="refresh-wrapper {% if auto-refresh %}enabled{% endif %}">`

and store "auto-refresh" in the user's session, it should work. 

It will also require a bit of change to the  "// Preview auto-refresh toggle" section of public/editor/scripts/editor/js/fc/bramble-ui-bridge.js as we need to change "auto-refresh" each time the user toggles. 

[...]

PR #1751, opened 2017-02-17
added front end support for auto update preference to fix issue #1633

timmoy:
I added some code to handle issue #1633 to the bramble-ui-bridge.js file. The back end for this is https://github.com/mozilla/brackets/pull/595.

This code just does the front end part, checking on load to see the preference stored for autoUpdate and changing the .refreshwrapper element to reflect that (by removing the checkmark if needed).

Final testing has yet to be done and will be done once the back end has been finalized as well.

### DIFF
--- a/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
+++ b/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
@@ -156,9 +156,10 @@ define(function(require) {
 
     // Set initial auto-refresh toggle to last known setting
     if(bramble.getAutoUpdate() === true) {
-        $(".refresh-wrapper").addClass("enabled");
     } else {
         $(".refresh-wrapper").removeClass("enabled");
+        bramble.disableAutoUpdate();
+        analytics.event("disableAutoUpdate");
     }
 
     // Preview auto-refresh toggle
--- a/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
+++ b/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
@@ -154,6 +154,13 @@ define(function(require) {
       _inspectorEnabled = data.enabled;
     });
 
+    // Set initial auto-refresh toggle to last known setting
+    if(bramble.getAutoUpdate() === true) {
+        $(".refresh-wrapper").addClass("enabled");
+    } else {
+        $(".refresh-wrapper").removeClass("enabled");
+    }
+
     // Preview auto-refresh toggle
     $(".toggle-auto-update").on("click", function() {
       var refreshWrapper = $(".refresh-wrapper");
--- a/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
+++ b/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
@@ -155,8 +155,7 @@ define(function(require) {
     });
 
     // Set initial auto-refresh toggle to last known setting
-    if(bramble.getAutoUpdate() === true) {
-    } else {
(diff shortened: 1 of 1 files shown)


timmoy:
I found what was causing the back end part of the solution to not work, and have pushed a new commit to that pull request.

This one doesn't require any further changes to support the functionality of keeping the same auto update setting when refreshing the browser.

If there are any problems with the code or styling that I might have missed, please let me know and I will fix them as soon as I can.

flukeout:
Hi @timmoy I'm running both the Thimble and Brackets PRs locally and seeing a bunch of javascript errors, when I make changes to the file I'm editing....

image

Are you seeing anything like this? It looks like the changes I'm making aren't being saved locally in my project.

flukeout:
Okay, here's what I'm finding

* The editor correctly remembers the setting for the Auto-refresh feature
  * If I turn it off, and reload, it comes back turned off
* However, even though the UI remembers the state, it still auto-refreshes

Check out the GIF below. I have a javascript file that fires an alert every time the page is reloaded. Initially, auto-refresh is disabled, so my changes to the file don't trigger the alert. Then when I reload the browser you can see the 'auto refresh' is turned off - BUT - it's still refreshing the preview when I make changes to the doc.

alert

humphd:
@timmoy grab me tomorrow if you want to debug this together.

timmoy:
Good catch @flukeout, looks like the setting wasn't saved for the editor although it was for the UI.
I managed to reproduce it and added the proper function calls and it works as intended for me now. 
Let me know if you still get the bug.

public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
```diff
@@ -154,6 +154,14 @@ define(function(require) {
       _inspectorEnabled = data.enabled;
     });
 
+    // Set initial auto-refresh toggle to last known setting
+    if(bramble.getAutoUpdate() === true) {
```
gideonthomas: Let's simplify this logic:
```js
if(!bramble.getAutoUpdate()) {
  ...
}
```
so that we remove the need for `if...else`

flukeout:
Double checking functionality now.

flukeout:
Hey great work, just confirmed that the setting is remembered and works properly now. Once you make the changes requested by @gideonthomas in this PR and the brackets one we can go ahead and merge!

timmoy:
The code has been optimized to remove the else branch and not do the comparison. 
I've tested it and functionality is working on my machine.

flukeout:
Running this and the brackets patch together produces this error in the console for me...

image

timmoy:
@flukeout Indeed I got the same errors when I checked the console. I managed to debug the commandmanager error by taking out this line:
https://github.com/timmoy/brackets/blob/a88f84681b4c708c8a1d60fed990e1024c50b3c7/src/editor/EditorOptionHandlers.js#L107
and functionality still seems intact.

I still get the other error 
image

I suspect there might be other bits of code that aren't needed and with that in mind I will continue to try and debug it.

A side note, do you also get this error?
image

humphd:
@timmoy don't remove that line, we need that.  The errors are likely due to your `dist/bramble.js` being out of date.  Try running `npm run build` in Brackets before you run `npm start`.

The second set of errors you noticed are a known issue related to localization and running Brackets from `src/` (not localized) vs. `dist/` (localized) and missing a string.  You can ignore it.

gideonthomas:
See https://github.com/mozilla/brackets/pull/595#issuecomment-282136384 for further fixes.

gideonthomas:
Thanks for the contribution @timmoy!

Issue #1755, opened 2017-02-17
Show the project description (if available) on the Your Projects page.


flukeout:
If we're going to keep it in the Publish dialog, let's at least make it somewhat useful.
I think we should include the project description as a line of text below the title. If it's longer than one line, we can cut it off with an ``...``

cgsingh:
Can I please have this issue. This seems like something I can actually handle, heh.

humphd:
Yup, go for it.

cgsingh:
@flukeout In order for `text-overflow: ellipsis;` to work, I need a `width: XXXpx`. What value do you want in there, or is there another way I'm supposed to be truncating a description into one line?

flukeout:
Closed via https://github.com/mozilla/thimble.mozilla.org/pull/1759 - thanks @cgsingh

PR #1759, opened 2017-02-19
Issue #1755 - Your Projects Page shows Project Descriptions

cgsingh:
d236e34598d1228269a576fb554b9b7c

Let me know if I should change anything. I made sure the "Delete" <td> was centered, otherwise it looked weird positioned below everything, especially in comparison to a project that didn't have a description.

### DIFF
--- a/views/editor/projects.html
+++ b/views/editor/projects.html
@@ -30,9 +30,10 @@
             <a target="_self" href="/{{locale}}/user/{{username}}/{{project.id}}/{{queryString}}"> 
               {{project.title}}
               <span class="project-information"></span>
+              <div class="project-description">{{project.description}}</div>
             </a>
           </td>
-          <td>
+          <td class="project-delete-cell">
             <div class="project-delete" title="{{ gettext("prjListDeleteProjectBtnTitle") }}">
               <div class="icon-garbage-can">
                 <div class="icon-garbage-lid"></div>
--- a/public/editor/stylesheets/projects-style.less
+++ b/public/editor/stylesheets/projects-style.less
@@ -4,7 +4,7 @@ body {
 }
 
 #container {
-  width: 65%;
+  width: 80%;
   margin: 35px auto;
   position: relative;
 }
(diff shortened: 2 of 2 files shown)


views/editor/projects.html
```diff
@@ -29,10 +29,11 @@
           <td class="project-title">
             <a target="_self" href="/{{locale}}/user/{{username}}/{{project.id}}/{{queryString}}"> 
               {{project.title}}
-              <span class="project-information"></span>
+              <span class="project-information"></span><br />
```
humphd: Use a `<div>` if you need it to be a block element, and get rid of the `<br />`.

cgsingh: Fixed.

flukeout:
Hey @cgsingh - great work so far. Here are my recommendations for styling the description:

* Make the text color ``rgba(0,0,0,.5)``
* Make the font-size ``.9em``
* Give it a margin-top: of ``5px``
* Give it a padding-right of ``20px``
  * This will add some space between the end of the description and trash icon, which looks a bit better

I'm okay with the max-width ``500px`` for now. Ideally we'd use a percentage-based max-width, but that doesn't work well within a table cell. We can re-write the project list to be an ``<ul>`` instead of a ``<table>`` sometime later.

Pomax:
@flukeout as quick note on " Ideally we'd use a percentage-based max-width, but that doesn't work well within a table cell. " you can also use `vw` which is always relative to the view port width, rather than the box size of whatever you're in.

flukeout:
@cgsingh give @Pomax's suggestion a try and see if it works! Ideally we want the max width to be just a little narrower than the size of the cell containing the description. If it doesn't work, no problem!

cgsingh:
@flukeout @Pomax I've applied both of your suggestions. Max-width seems to be working better with vw. I set the value to 40vw.

1f55ae3c954a1ae07a0ca07a89504a68

flukeout:
Hey this looks great, just one final change to make... 

* Set the width of the ``#container`` to 80% instead of the current 65% - then the **Create New Project** button will stay aligned with the right edge of the table. 

Once that's in, we'll get it merged. Thank you!

cgsingh:
@flukeout Changed. Thanks for the help.

Issue #1764, opened 2017-02-21
Windows with VMWare and vagrat + virtualbox does not necessarily work


Pomax:
I use VMWare Workstation constantly, and trying to use the Thimble vagrant+virtualbox solution is causing all kinds of headaches. With VMWare services running, I get
```
==> default: Clearing any previously set forwarded ports...
Vagrant cannot forward the specified ports on this VM, since they
would collide with some other application that is already listening
on these ports. The forwarded port to 5432 is already in use
on the host machine.
```
With all VMWare services stopped, the error changes to
```
==> default: Checking if box 'ubuntu/trusty64' is up to date...
Vagrant cannot forward the specified ports on this VM, since they
would collide with some other application that is already listening
on these ports. The forwarded port to 5432 is already in use
on the host machine.

To fix this, modify your current project's Vagrantfile to use another
port. Example, where '1234' would be replaced by a unique host port:

  config.vm.network :forwarded_port, guest: 5432, host: 1234
```

Since vagrant doesn't really care about *which* VM technology it uses, it might be possible to make it use VMWare if already available, so we don't need to introduce the headache of trying to make VMWare and VirtualBox play nice together (which is a deep rabbit hole)

humphd:
It looks to me like we should be able to write configuration such that you can use either virutalbox or vmware in the same `Vagrantfile`: https://thornelabs.net/2013/09/19/multi-machine-vagrantfile-for-virtualbox-or-vmware-fusion.html.  Let's try and figure this out so you can just always use vmware, and I can always use virutalbox.

humphd:
https://www.vagrantup.com/docs/providers/basic_usage.html has more info on this.  This sounds doable.

humphd:
And some examples in the wild:

* https://github.com/nateswart/vagrant-nodejs/blob/cee9e357e95af90a8bd59172f7bfbf3f54507584/Vagrantfile
* https://github.com/Abdul0/OSSEC-Ansible/blob/d519f2ad575a74544d3566d2ab3ddd159c839014/provisioning/Vagrantfile

gideonthomas:
@Pomax 5432 is generally used for postgres. Are you running postgres in the background by any chance?

Pomax:
This is a dev machine, it has all kinds of services running. Postgres is definitely one of those because I work on other projects beside Thimble, almost half of which require postgres =)

humphd:
[...]

PR #1765, opened 2017-02-21
Don't leak port 5432 from Vagrant VM into host, part of #1764

humphd:
This stops us from forwarding the postgres port out of the VM, so that it won't conflict with services running in the host.

I tested it locally, and it still runs fine for me.  I'm not sure if it will fix Pomax's issue.


Pomax:
I'll have to test this when I get home later today

Pomax:
works like a charm.

PR #1775, opened 2017-02-22
Updated styling for Your Projects page

### LINKED ISSUE #1626: Your Projects List > Add link to project if already published.
On the Your Projects list, let's indicate if a project has been published, and if so, provide a link to the published project.

### LINKED ISSUE #1774: Style cleanup on My Projects page
Recently we've merged in some great improvements to the My Projects page. Unfortunately, as a result, the styling of the page is now a little bit off...

image

The fav stars are misaligned with the content as a result of the descriptions being shown. Filing this for visibility, but I'll take it on, as it's largely a CSS exercise.

flukeout:
**Changes**

* Added small animation to the star icon when it's turned on
* Updated text styling
* Markup & style cleanup
* Hiding description element when it's empty
* Added Published Url when available
* Hiding the list until the projects have been sorted by favorites to avoid the jarring change
* Added a little bit of responsive styling 

**Now looks like...**

image

Fixes #1774 
Fixes #1626

### DIFF
--- a/public/editor/scripts/project-list.js
+++ b/public/editor/scripts/project-list.js
@@ -18,7 +18,7 @@ require.config({
 });
 
 require(["jquery", "constants", "analytics", "moment"], function($, Constants, analytics, moment) {
-  var projects = document.querySelectorAll("tr.bramble-user-project");
+  var projects = document.querySelectorAll(".bramble-user-project");
   var locale = $("html")[0].lang;
   var isLocalStorageAvailable = !!(window.localStorage);
   var favorites;
--- a/public/editor/scripts/project-list.js
+++ b/public/editor/scripts/project-list.js
@@ -18,6 +18,7 @@ require.config({
 });
 
 require(["jquery", "constants", "analytics", "moment"], function($, Constants, analytics, moment) {
+  document.querySelector("#project-list").style.opacity = 1;
   var projects = document.querySelectorAll(".bramble-user-project");
   var locale = $("html")[0].lang;
   var isLocalStorageAvailable = !!(window.localStorage);
--- a/public/editor/scripts/project-list.js
+++ b/public/editor/scripts/project-list.js
@@ -46,7 +46,7 @@ require(["jquery", "constants", "analytics", "moment"], function($, Constants, a
       $(projectFavoriteButton).toggleClass("project-favorite-selected");
     }
 
-    $(projectSelector + " .project-favorite").on("click", function() {
+    $(projectSelector + " .project-favorite-button").on("click", function() {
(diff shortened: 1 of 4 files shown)


flukeout:
Paging @gideonthomas the coolest guy to ever live! So talented and fast at merging PRs, ooooh yeah 🎵🎵🎵🎵

gideonthomas:
Sorry :P
@flukeout can you rebase this patch onto master? I can't seem to test this because it doesn't have the vagrant code.

flukeout:
Just to clarify, @gideonthomas is having some kind of vagrant issue, and the patch can be tested. @Pomax ?

public/editor/scripts/project-list.js
```diff
@@ -18,7 +18,8 @@ require.config({
 });
 
 require(["jquery", "constants", "analytics", "moment"], function($, Constants, analytics, moment) {
-  var projects = document.querySelectorAll("tr.bramble-user-project");
+  document.querySelector("#project-list").style.opacity = 1;
```
humphd: Why not via a class or from a `.css` file?

flukeout: Huh yeah... no comment. Will fix.

public/editor/stylesheets/projects-style.less
```diff
@@ -33,189 +12,181 @@ h2 {
...
-.project-button:active {
-  background-color: #288952;
-  top: 2px;
-  border: none;
-}
+  .project-delete:hover :after {
+    transform-origin: left;
+    -webkit-transform-origin: left;
+    transform: rotate(-4deg) translateY(-1px);
+    -webkit-transform: rotate(-4deg) translateY(-1px);
```
humphd: Can you put the unprefixed versions of these two sets last vs. first, so they always get applied if available?

flukeout: Yeah totally, in fact I found a mixin that we already use in LESS to apply all the prefixes so I'll try to use that.

gideonthomas:
@flukeout really nice! Works pretty well.

Just one comment which might not be an issue - the clickable area for opening a project is only here:
screen shot 2017-02-24 at 3 14 44 pm 2
which means that if you click just below that above the demarcation, it does nothing.

public/editor/stylesheets/projects-style.less
gideonthomas: We might want to consider using autoprefixer and adding it to the less-middleware so that we don't even need to include vendor prefixes.

flukeout:
@gideonthomas Made the entire height of the row clickable.

Issue #1762, opened 2017-02-20
Funky Console Errors on Staging


flukeout:
Seeing lots of JS errors when I duplicate the link tag in the head of a brand new project on staging.

* Go to https://bramble.mofostaging.net/
* Start a new project
* Copy and paste the existing style tag
* Check the console and see...

![image](https://cloud.githubusercontent.com/assets/25212/23144620/b13850a4-f77d-11e6-83a8-b9c71d1df80f.png)

What is this stuff? @gideonthomas @humphd

humphd:
There's at least one, maybe more bugs here.  The first issue, with the exception in `DocumentRelated` is apparently not a new bug--I can reproduce this on production.  My best guess is that somehow we're throwing away (or regenerating) the `style.css` file's blob URL and one of the two that gets included is failing.  It's dying like so:

```
GET blob:http://localhost:8000/e08b6995-6250-47ec-8972-235d6cd6844a 
EventDispatcher.js:231 Exception in 'DocumentRelated' listener on Object {_EventDispatcher: true, _eventHandlers: Object} Error: Paths must be absolute: 'blob:http://localhost:8000/e08b6995-6250-47ec-8972-235d6cd6844a' Error: Paths must be absolute: 'blob:http://localhost:8000/e08b6995-6250-47ec-8972-235d6cd6844a'
    at FileSystem._normalizePath (http://localhost:8000/src/filesystem/FileSystem.js:489:19)
    at FileSystem._getEntryForPath (http://localhost:8000/src/filesystem/FileSystem.js:556:21)
    at FileSystem.getFileForPath (http://localhost:8000/src/filesystem/FileSystem.js:575:21)
    at Object.getFileForPath (http://localhost:8000/src/filesystem/FileSystem.js:988:25)
    at Object.getDocumentForPath (http://localhost:8000/src/document/DocumentManager.js:342:46)
    at _styleSheetAdded (http://localhost:8000/src/LiveDevelopment/LiveDevMultiBrowser.js:309:42)
    at http://localhost:8000/src/LiveDevelopment/LiveDevMultiBrowser.js:574:29
    at Array.forEach (native)
    at http://localhost:8000/src/LiveDevelopment/LiveDevMultiBrowser.js:573:30
    at Object.trigger (http://localhost:8000/src/utils/EventDispatcher.js:229:40)
```

Basically, we've deleted the URL for the `style.css` file and generated a new one.  Here's what the preview HTML looks like when rewritten:

[...]

PR #1777, opened 2017-02-23
Fix #1738 - `callback` is not a function when showing "Update Published Project" prompt

### LINKED ISSUE #1738: `callback` is not a function when showing "Update Published Project" prompt
Bug introduced in 392fb9dccc4a228b21dfe88cfe415cb56a6acded

https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/editor/js/fc/publisher.js#L65-L68 <-- these lines use the `showUnpublishedChangesPrompt` function from the `publisher`'s prototype. However, the events that it is attached to send parameters that are not a callback (which is what the function actually expects) and it fails.

We should probably not use function bindings for these event listeners. Instead, let's just use closures like:
```js
bramble.on("event", function() {
  publisher.showUnpublishedChangesPrompt();
});
```

lkisac:
Fix for Issue #1738

### DIFF
--- a/public/editor/scripts/editor/js/fc/publisher.js
+++ b/public/editor/scripts/editor/js/fc/publisher.js
@@ -62,11 +62,19 @@ define(function(require) {
       this.updateDialog(publishUrl, true);
     }
 
-    bramble.on("fileChange", publisher.showUnpublishedChangesPrompt.bind(publisher));
-    bramble.on("fileDelete", publisher.showUnpublishedChangesPrompt.bind(publisher));
-    bramble.on("fileRename", publisher.showUnpublishedChangesPrompt.bind(publisher));
-    bramble.on("folderRename", publisher.showUnpublishedChangesPrompt.bind(publisher));
-
+    bramble.on("fileChange", function() {
+      publisher.showUnpublishedChangesPrompt();
+    });
+    bramble.on("fileDelete", function() {
+      publisher.showUnpublishedChangesPrompt();
+    });
+    bramble.on("fileRename", function() {
+      publisher.showUnpublishedChangesPrompt();
+    });
+    bramble.on("folderRename", function() {
+      publisher.showUnpublishedChangesPrompt();
+    });
+        
     dialog.buttons.publish.on("click", publisher.handlers.publish);
 
     // Were there any files that were updated and not published?
--- a/public/editor/scripts/editor/js/fc/publisher.js
+++ b/public/editor/scripts/editor/js/fc/publisher.js
@@ -63,16 +63,16 @@ define(function(require) {
     }
 
     bramble.on("fileChange", function() {
-      publisher.showUnpublishedChangesPrompt();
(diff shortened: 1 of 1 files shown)


public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -62,11 +62,19 @@ define(function(require) {
       this.updateDialog(publishUrl, true);
     }
 
-    bramble.on("fileChange", publisher.showUnpublishedChangesPrompt.bind(publisher));
-    bramble.on("fileDelete", publisher.showUnpublishedChangesPrompt.bind(publisher));
-    bramble.on("fileRename", publisher.showUnpublishedChangesPrompt.bind(publisher));
-    bramble.on("folderRename", publisher.showUnpublishedChangesPrompt.bind(publisher));
-
+    bramble.on("fileChange", function() {
+      publisher.showUnpublishedChangesPrompt();
```
Pomax: remember to also pass all the arguments down in this rewrite:
```
publisher.showUnpublishedChangesPrompt.apply(publisher, arguments);
```
The old code returned a function that would then be called by `bramble` with any number of arguments, so to preserve that functionality we can't just call the function without any args, whatever bramble passes in needs to be passed on to the handler function

lkisac: Thanks @Pomax, I adjusted it in the latest commit.

gideonthomas:
@lkisac I would revert the change you just made. We don't want to pass in the arguments that the function receives because it does not match the arguments that `showUnpublishedChangesPrompt` accepts. `showUnpublishedChangesPrompt` only accepts a `callback` while the arguments passed in by the `bramble` filechange events are `fileName` and something else. Hence, why this change removes the `bind` which we used before. `apply` basically does the same as `bind` in this case.

public/editor/scripts/editor/js/fc/publisher.js
gideonthomas: > so to preserve that functionality

I think this change is meant to not preserve that functionality as that was what was causing the error :)

lkisac:
@gideonthomas  Thanks.  Good to know.  Please let me know if there is any more changes needed to be made.

humphd:
How close are we to landing this?  I'm getting a lot of confusion from devs who are hitting this and don't know how to overcome it.  It would be great to get the fix up.

lkisac:
@humphd  I couldn't replicate the issues in #1762   I've tested `fileChange`, `fileDelete`, `fileRename`, `folderRename` by changing a file, deleting, renaming, etc. but don't see any error messages.  What else can I try on my end to replicate one of issues devs are running into?

gideonthomas:
@lkisac I don't think there's anything you can do from your end. Sorry for the delay, I'm testing this right now to do one final review.

lkisac:
Ok, thanks @gideonthomas   Will wait for your review.

lkisac:
@gideonthomas Glad it works!

PR #1783, opened 2017-02-24
Added direct link to Thimble channel on Mattermost instance. Fixes #1782

### LINKED ISSUE #1782: README should link directly to the Thimble channel on Mattermost instance.
As it currently stands, the README points chat-friendly folks to http://chat.mozillafoundation.org/. While this link is (partially) correct, it does not fulfil its purpose properly. 

Those who followed the link _intend_ to talk about Thimble. Instead they get directed to the `~Town Square` channel, where they bumble and stumble before finding their way to the relevant `~Thimble` channel. 

Since a dedicated channel for Thimble exists, it only makes sense that this link point directly to the relevant channel.

babhishek21:
Fixes #1782.

Let me know if the text needs to change. :smiley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,6 @@ Thimble uses the [throng](https://www.npmjs.com/package/throng) module to levera
 Contact Us
 -----------
 
-We're a friendly group, so feel free to chat with us in the "Thimble" channel on Mozilla Chat running on [Mattermost](https://about.mattermost.com). To access Mozilla Chat head to [this link]( http://chat.mozillafoundation.org). Note that you will need to create an account first.
+We're a friendly group, so feel free to chat with us in the "Thimble" channel on Mozilla Chat running on [Mattermost](https://about.mattermost.com). To access Mozilla Chat head to [this link]( http://chat.mozillafoundation.org). Note that you will need to create an account first. If you have signed in already, head [here](https://chat.mozillafoundation.org/mozilla/channels/thimble) to directly visit the Thimble channel.
 
 You can also download a mobile or desktop client for Mattermost [here](https://about.mattermost.com/download/#mattermostApps).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -202,6 +202,6 @@ Thimble uses the [throng](https://www.npmjs.com/package/throng) module to levera
 Contact Us
 -----------
 
-We're a friendly group, so feel free to chat with us in the "Thimble" channel on Mozilla Chat running on [Mattermost](https://about.mattermost.com). To access Mozilla Chat head to [this link]( http://chat.mozillafoundation.org). Note that you will need to create an account first.
+We're a friendly group, so feel free to chat with us in the "Thimble" channel on Mozilla Chat running on [Mattermost](https://about.mattermost.com). To access Mozilla Chat head to [this link]( http://chat.mozillafoundation.org). Note that you will need to create an account first. If you have signed in already, head [here](https://chat.mozillafoundation.org/mozilla/channels/thimble) to directly visit the Thimble channel.
```
Pomax: it would be nicer if we can rephrase things here because we're pointing to a domain, and then a page on that domain (as far as URLs are concerned) so I'd suggest:
```
To access Mozilla Chat head over to [chat.mozillafoundation.org](http://chat.mozillafoundation.org). Note that you will be prompted to create an account if you do not already have one. If you already have an account, and you are already logged in from a previous visit, you can directly access the Thimble channel by clicking [this link](https://chat.mozillafoundation.org/mozilla/channels/thimble).
```

babhishek21: But isn't that too much information? 👀

humphd: I think saying more is good in this case, since we're trying to make it as easy as possible for people who are new to find and get connected with us.

babhishek21:
@Pomax @humphd  Changed the phrasing as discussed. Let me know if anything else is needed. :smile:

PR #1788, opened 2017-02-24
Adding the homepage Gallery

flukeout:
In-progress work for adding a project gallery to the homepage.
It's modeled after this protototype: https://flukeout.github.io/thimble-gallery/

Activity data lives in a separate repo, here:
* https://github.com/mozilla/thimble-homepage-gallery/blob/gh-pages/activities.json

**To-do**
* Responsive styling
* Handle cases where lots of tags in search box break the layout
* Rip out code related to old Starter Projects implementation
* Add loading indicator when external activity data is being fetched

@gideonthomas could you review the progress thus far?

Should appear like this, on the homepage...
image


flukeout:
Also @gideonthomas can you help me resolve the travis errors? Looks like it can't find some files. Do I have to run a build step of some sort in thimble?

public/homepage/scripts/main.js
```diff
@@ -17,6 +17,7 @@ require.config({
     "uuid": "/node_modules/node-uuid/uuid",
     "cookies": "/node_modules/cookies-js/dist/cookies",
     "analytics": "/node_modules/webmaker-analytics/analytics",
+    "gallery": "/{{ locale }}/homepage/scripts/gallery",
```
gideonthomas: If you're creating it as a module here, you'll also need to add it to the Gruntfile here: https://github.com/mozilla/thimble.mozilla.org/blob/master/Gruntfile.js#L37-L50
`gallery: "homepage/scripts/gallery"`

flukeout:
Now it looks like travis is failing on a template literal...

``>> Unexpected character '`' (line: 1009, col: 21, pos: 116354)``

...is that right? Should I change this?

gideonthomas:
@flukeout yeah we don't support ES6 code yet. I would also suggest moving that to the top as a variable maybe.

flukeout:
Okay! Moved the template markup to the gallery.html page instead. So long, template literals.

gideonthomas:
@TheoChevalier would you be willing to look at the l10n part of this PR please? I'm a bit unsure about some of the strings (like "By", "Lesson Plan", etc.  and the way they're ordered)

views/gallery.html
```diff
@@ -0,0 +1,48 @@
...
+
+    <div class="search-header">
+      <h2 class='title'></h2>
+
+      <div class='search-wrapper'>
+        <div class='search-wrapper-outer'>
+          <span class='search-tags'></span>
+          <input class='search' placeholder='{{ gettext("gallerySearchPlaceholder") }}' type='text' />
+          <div class='icon search-icon'></div>
+          <div title='Clear the search term' class='clear icon clear-icon'></div>
```
TheoChevalier: Wrap `'Clear the search term'` for l10n

locales/en-US/messages.properties
```diff
@@ -166,6 +166,17 @@ homeworkExcuseGeneratorRemixTitle=Remix the {{ homeworkExcuseGenerator }} projec
 homeworkExcuseGeneratorTeachingKitTitle=Read the {{ homeworkExcuseGenerator }} teaching kit
 mySixWordSummerRemixTitle=Remix the {{ mySixWordSummer }} project
 mySixWordSummerTeachingKitTitle=Read the {{ mySixWordSummer }} teaching kit
+lessonPlan=Lesson Plan
+projectBy=By
+remixGalleryTitle=Remix a project to get started...
+projectSingular=project
```
TheoChevalier: Let’s use something like `projectSingular={{ n }} project`, so that the number can be moved around.
Ideally, we should use proper plural form, since singular/plural won’t work for most languages:
`project[one]={{ n }} project`
`project[two]={{ n }} projects`
`project[few]={{ n }} projects`
`project[many]={{ n }} projects`

But I’m not sure what plural form support you have with this gettext implem (is this node-gettext?)

locales/en-US/messages.properties
```diff
@@ -166,6 +166,17 @@ homeworkExcuseGeneratorRemixTitle=Remix the {{ homeworkExcuseGenerator }} projec
...
+projectPlural=projects
```
TheoChevalier: Same here, let’s use something like `projectPlural={{ n }} projects`

locales/en-US/messages.properties
```diff
@@ -166,6 +166,17 @@ homeworkExcuseGeneratorRemixTitle=Remix the {{ homeworkExcuseGenerator }} projec
...
+found=found
```
TheoChevalier: If it’s concatenated with a number, we will need proper plural here, too. At least singular/plural depending on the support you have

flukeout:
Thanks for all of the great suggestions @TheoChevalier - I've removed the "Tags" content from the CSS entirely and reworked one of the headings to not rely on numbers so localization is easier.

TheoChevalier:
Thanks @flukeout, looks great!

flukeout:
Alright @gideonthomas want to look this over one last time?

Issue #1795, opened 2017-02-28
Fix incorrect use of "login" to "log in".


flukeout:
A contributor found this string somewhere within Pontoon...

![image](https://cloud.githubusercontent.com/assets/25212/23413704/6044751c-fd8e-11e6-90fe-b2e50b79afa1.png)

Must be some kind of system error, but let's get it updated anyway.

gideonthomas:
Fixed with 77d48ed0ff440bc1830fb5cb38a0cf263efc2a86

PR #1799, opened 2017-03-01
Fix incorrect use of "login" to "log in". #1795

Rishabh04-02:
[#1795](https://github.com/mozilla/thimble.mozilla.org/issues/1795)


gideonthomas:
Thanks for the contribution @Rishabh04-02

PR #1802, opened 2017-03-02
Fix #1801 - Update l10n keys for changed strings

### LINKED ISSUE #1801: Several strings changed without updating the string ID
Several strings have been updated over the past months, but the string ID has not been updated. If IDs are not updated, translations are not invalidated, and localizers are not warned that the source string has changed. Doing so results in updating en-US only, and not the other locales.

See also this page to understand when IDs need to be updated and when they don’t: https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#Changing_existing_strings
Happy to answer questions, too

Tonnes and I identified the following commits with string updates without an ID change:

https://github.com/mozilla/thimble.mozilla.org/commit/1f4b98799227fd80d6cf8345a9998ef00ced636f
https://github.com/mozilla/thimble.mozilla.org/commit/bc500a8550e0b3d82e7cff4eae365300c16b4b44
https://github.com/mozilla/thimble.mozilla.org/commit/adf9a5f79f7828e2b025cc673c412dff823d2928
https://github.com/mozilla/thimble.mozilla.org/commit/612e0eedff6a5dadf8c30fefb2697123fd5912b4
https://github.com/mozilla/thimble.mozilla.org/commit/59bb2d8696be5963dccef75d967f888c17b9c887

Side note, please don’t commit changes in other locales, even if in this case en-GB is really close to en-US: https://github.com/mozilla/thimble.mozilla.org/commit/bc500a8550e0b3d82e7cff4eae365300c16b4b44#diff-1b35da63addb17b2751083c61013aac6

gideonthomas:
Updates *most* of the keys for which values were changed in previous patches.


locales/en-US/messages.properties
```diff
@@ -135,10 +135,10 @@ publishedLink=Published Version
 yourProjectsLink=Your Projects
 loginWelcome=Hi, {{username}}
 newProjectLink=New Project
-signOutLink=Sign out
+signOutLink2=Sign out
```
TheoChevalier: (actually, should write that here…)
This one does not need to be updated (capitalization rules are specific to each language, so translations are already correct)

TheoChevalier:
One nit, otherwise, looks good :)

gideonthomas:
Thanks for the prompt review @TheoChevalier!

PR #1805, opened 2017-03-04
Fixing Incremental backoff for saving

### LINKED ISSUE #1554: Incremental backoff for saving isn't working correctly
on [line 31](https://github.com/mozilla/thimble.mozilla.org/blob/e6d8ba2162767b0de8aa48e8bad5cacf497bec1c/server/routes/files/delete.js#L31) we reference `status` which is undefined.

This slipped in with #1521 

cc/ @gideonthomas

evantsyferov:
Fixes #1554


Pomax:
Hi @evantsyferov it looks like your PR does not pass CI testing - can you run `npm test` on your local instance and fix any errors that might flag?

humphd:
This change doesn't deal with the problem here: https://github.com/mozilla/thimble.mozilla.org/blob/e6d8ba2162767b0de8aa48e8bad5cacf497bec1c/server/routes/files/delete.js#L31

It looks to me like you're pushing another branch for your earlier fix.  You want a new, clean branch:

```
git checkout master
git pull upstream master
git checkout -B issue1554
```

Now you can start working on the actual problem.  When you've fixed it and done your commit, push your branch up using `-f` so that it overwrites this other code:

```
git push origin issue1554 -f
```

Pomax:
Is this a work-in-progress PR, or is the change introduced in the PR the suggested solution for #1554?

server/routes/files/delete.js
```diff
@@ -28,7 +28,7 @@ module.exports = function(config, req, res, next) {
     }
 
     if(response.statusCode !== 204) {
-      res.status(response.statusCode);
+      res.status(req.status);
```
humphd: What you had before was correct.  Look at line 34 below, which also references `response.statusCode`.

humphd:
@evantsyferov can we please tie this up?  It's a one line fix.  You need to correct your code as I wrote in my comment, then rebase and squash these commits into one.

Pomax:
@evantsyferov the "fixes #...." is specific and exact syntax for github.com issue crosslinking, so either put "fixes #..." in the title or the PR comment (preferably the comment) but don't use a different word or github will not pick up on the fact that it needs crosslink issues to PRs.

humphd:
@evantsyferov please talk to me about this when we meet up again, so we can finish this.  It's really close.

humphd:
This was fixed in https://github.com/mozilla/thimble.mozilla.org/commit/75c4aea375005166e66ccb03ef58e40b3f5a42f6

Issue #1637, opened 2017-01-16
Always display a Tutorial in "Desktop" mode.


flukeout:
Currently, if the preview is set to Mobile view and you click to view the tutorial, it remains in the mobile view. This makes it hard to use...

![image](https://cloud.githubusercontent.com/assets/25212/22002229/1a51df3c-dbff-11e6-88bf-ff6c426700ff.png)

Let's always show the Tutorial file in desktop view.

This means that if we're previewing the ``index.html`` page in the mobile view, we need to return the preview to mobile after showing the tutorial.

humphd:
This will likely involve changes to https://github.com/mozilla/brackets/blob/master/src/extensions/default/bramble/lib/Tutorial.js and https://github.com/mozilla/brackets/blob/master/src/extensions/default/bramble/lib/UI.js#L207

rkgupta21:
I am interested in working on the issue.

humphd:
Great, @rkgupta21, it's yours.

gideonthomas:
@rkgupta21 how are you doing with this bug? Let us know if you need any guidance.

Rajat-dhyani:
hey @rkgupta21 if you are not working on this...can i try to fix this issue ?

gideonthomas:
@humphd can we re-assign this issue?

humphd:
@rkgupta21 can you let us know if you're working on this still?

rkgupta21:
@humphd I was actually working on some other issue. You can reassign it.

humphd:
Thanks @rkgupta21, that's helpful to know.  @Rajat-dhyani go for it if you want this one.

Rajat-dhyani:
okay thanks for informing @rkgupta21 ... i am taking this over 👍

Rajat-dhyani:
please review this PR : https://github.com/mozilla/thimble.mozilla.org/pull/1807

Pomax:
@Rajat-dhyani can you update your PR so that the PR comment says `fixes #1637` somewhere? That makes github crosslink the two issues, flagging that your PR fixes this specific issue.

Rajat-dhyani:
@Pomax Done :) 👍

PR #1807, opened 2017-03-05
Fix: 1637 Always display a Tutorial in Desktop mode 

Rajat-dhyani:
Currently, if the preview is set to Mobile view and you click to view the tutorial, it remains in the mobile view. This fix always show the Tutorial file in desktop view. This means that if we're previewing the index.html page in the mobile view, we return the preview to mobile after showing the tutorial.
Fixes : [1637](https://github.com/mozilla/thimble.mozilla.org/issues/1637)

### DIFF
--- a/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
+++ b/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
@@ -13,6 +13,7 @@ define(function(require) {
 
   var adapting = false;
   var adaptTimeoutMS = 200; // How often we adapt editor bar layout
+  var _isMobileView = false; // to check wheather mobile view is on/off
 
   function updateLayout(data) {
     $(".filetree-pane-nav").width(data.sidebarWidth);
@@ -287,6 +288,12 @@ define(function(require) {
       }
 
       bramble.hideTutorial(setNormalPreview);
+
+      if (_isMobileView)
+      {
+        //if mobile view is preserved
+        activatePreviewMode("mobile");
+      }
     });
     $("#tutorial-title").click(function() {
       if(bramble.getTutorialVisible()) {
@@ -294,6 +301,11 @@ define(function(require) {
       }
 
       bramble.showTutorial(setTutorialPreview);
+      if (_isMobileView)
+      {
+        //display tutorial always in desktop mode
+        activatePreviewMode("destop");
+      }
     });
 
     // Programmatic change to tutorial vs. regular preview mode from Bramble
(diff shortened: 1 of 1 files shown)


public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
```diff
@@ -13,6 +13,7 @@ define(function(require) {
 
   var adapting = false;
   var adaptTimeoutMS = 200; // How often we adapt editor bar layout
+  var _isMobileView = false; // to check wheather mobile view is on/off
```
Pomax: please don't use underscores in variables: there is nothing special about this variable and it deserves a normal name. Also on a typo issue, `wheather` -> `whether`

flukeout:
So just to clarify, a couple of changes are needed here...

* Remove the underscore from the variable
* Fix the "whether" spelling

Rajat-dhyani:
@humphd yes sir i actually took the cue from brackets code base to name the private variable starting with underscore.. @Pomax @flukeout i did the changes accordingly please check this pr.

humphd:
I'm confused about these changes.  It seems to me that we should be making them in Brackets, so that when you open a tutorial, we automatically switch to Desktop mode.  I thought you had a patch happening for that already?  Why this fix @Rajat-dhyani?

Rajat-dhyani:
actually sir @humphd first i did the changes in brackets only but then there i couldn't preserve the mobile state of the preview mode... as is stated in the problem statement that we have to switch back to mobile view in the preview mode... so i did the changes here instead.

humphd:
@Rajat-dhyani just a note to let you know that this is still on my radar, and among the many reviews I'm working through.   I'll try to get through them this week.

Rajat-dhyani:
@humphd sounds great sir.. thanks for letting me know :)

humphd:
OK, sorry for the delay in this.  I was looking for my notes on this, and I suspect that I must have said this in chat vs a bug, so let me repeat it.

What we need to do is fix this inside Brackets, not Thimble. The tutorial override is set here:

https://github.com/mozilla/brackets/blob/master/src/extensions/default/bramble/lib/Tutorial.js#L20-L36

We need to add some code there to also force us into Desktop mode, which happens here:

https://github.com/mozilla/brackets/blob/master/src/extensions/default/bramble/lib/UI.js#L225-L248

So you basically need to call `UI.showDesktopView()` when you set the tutorial override.  The reason this will fix Thimble is because we emit events to let it know about the change to tutorial/view modes:

https://github.com/mozilla/brackets/blob/master/src/extensions/default/bramble/lib/UI.js#L243

I'm going to close this PR, and ask that you instead fix it in Brackets according to what I've just written here.  Let me know if you have questions about that, and ideally in the bug or a PR so we don't lose track of the info.

Thanks for working on this.

PR #1809, opened 2017-03-06
Redirects the user to his projects page after he signs in from the home page.

### LINKED ISSUE #1760: Redirect Sign In on homepage to Your Projects list
It's a little annoying getting back to a project you were working on after signing in with Thimble. Once you sign in, you pretty much always end up navigating to the **Your Projects** view from the user dropdown.

**So..**

* When someone Signs In from the homepage, let's redirect them there right away.
* If they're already in the Editor though, let's keep them there.

hkirat:
In reference to  #1760 
Simply removes the code that redirected to the home page in case a user came from the same.

### DIFF
--- a/server/routes/auth/oauth2-callback.js
+++ b/server/routes/auth/oauth2-callback.js
@@ -145,12 +145,6 @@ module.exports = function(config, req, res, next) {
         );
       }
 
-      // Was this sign-in triggered from the home page?
-      if (req.session.home) {
-        delete req.session.home;
-        return res.redirect(301, path.join("/", locale));
-      }
-
       res.redirect(301, path.join("/", locale, "/editor"));
     });
   });


flukeout:
Hi @hkirat can you explain what this PR is for, what changes it makes, and how to test it?

Is this an fix for issue #1760? That issue was already being worked on by @sgupta7857.

Thank you!

hkirat:
@flukeout Yes it is a fix for the issue #1760.
I was on a bug hunt on this repo yesterday, and tried to fix some bugs. Though I do see people are assigned to many of them now. 

@sgupta7857 this change redirects the user to the projects page in all three cases as mentioned by you here : #1760. And initialises a new project if he doesn't have one already.

@flukeout Should I close the P.R then?

humphd:
Let's go ahead with this PR, since our goal is to fix the bug.  However, in future it would be ideal if we communicated with one another about who is doing what.  I realize it's not always easy/obvious, and I appreciate that you are both trying to contribute--we want to make sure you can both do it!

flukeout:
I'll test the changes now.

flukeout:
OK - this works great as far as I can tell! Thanks for sovling it @hkirat

PR #1818, opened 2017-03-06
Fix #1595: allow blob: URLs in CSP to download zip files from editor iframe

### LINKED ISSUE #1595: Unable to export zip
We got an email about a user being unable to export the project as a zip on Firefox on Linux and Windows. They are able to export on Chrome though.

Will provide more details once they are available.

@humphd any first guesses as to why this might be happening?

humphd:
Tested locally on Firefox, and this allows the `.zip` download to happen.

### DIFF
--- a/server/security.js
+++ b/server/security.js
@@ -14,11 +14,13 @@ let defaultCSPDirectives = {
   ],
   frameSrc: [
     "'self'",
-    "https://docs.google.com"
+    "https://docs.google.com",
+    "blob:"
   ],
   childSrc: [
     "'self'",
-    "https://pontoon.mozilla.org"
+    "https://pontoon.mozilla.org",
+    "blob:"
   ],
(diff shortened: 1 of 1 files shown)


server/security.js
```diff
@@ -14,7 +14,8 @@ let defaultCSPDirectives = {
   ],
   frameSrc: [
     "'self'",
-    "https://docs.google.com"
+    "https://docs.google.com",
+    "blob:"
```
gideonthomas: do you need to add it to `childSrc` below too?

humphd: http://stackoverflow.com/questions/30023608/how-to-use-frame-src-and-child-src-in-firefox-and-other-browsers seems to indicate that `frameSrc` is right (again)?  I could do both.

humphd:
I added it to both, tested and landed.

PR #1820, opened 2017-03-07
Added extra error message for developer troubleshooting

### LINKED ISSUE #1758: Provide useful error when bramble.js can't be loaded in Thimble
If you setup Thimble but forget (or don't know you have to) run Brackets as well, you get our default black error page saying that the project couldn't be loaded.  In reality, it has failed to load `bramble.js` from Brackets, which it needs to load the editor.

We should provide some kind of error for developers/devops running Thimble so they know how to fix it.  I've seen quite a few people get stuck on this.

edyedy123:
Added errorLoadingProjectBramble error message to messages.properties
Added errorLoadingProjectBramble error to bramble.html

I saw that the error message already existed and showed in developers tools console.
I added a error message that tells the user to look in their browsers developer tools console to see error details.  
This will make is so that developers know if bramble is the cause of the error right away alongside of any other errors they might not have known about. 
Here is a picture of the dev tool error message.
https://gyazo.com/e1a4ffbbe37a11aaf8200c8273327a77

fixes #1758

### DIFF
--- a/locales/en-GB/messages.properties
+++ b/locales/en-GB/messages.properties
@@ -33,7 +33,6 @@ errorLoadingProjectSuggestion=NOTE: if you are using Private Browsing mode, plea
 errorUnsupportedBrowser=Sorry, Thimble won't work in your Browser
 errorUnsupportedBrowserSuggestion=Thimble works in all of the following browsers: <a title="Download Mozilla Firefox" href="https://www.mozilla.org/{{ locale }}/firefox/new/">Firefox</a>, <a title="Download Google Chrome" href="https://www.google.com/chrome/browser/desktop/index.html">Chrome</a>, Internet Explorer 11, Microsoft Edge, Safari (8+), and <a title="Download Opera" href="http://www.opera.com/download">Opera</a>.
 errorUnsupportedBrowserIgnore=Let me try anyway!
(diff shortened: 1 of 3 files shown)


locales/en-GB/messages.properties
```diff
@@ -33,6 +33,7 @@ errorLoadingProjectSuggestion=NOTE: if you are using Private Browsing mode, plea
 errorUnsupportedBrowser=Sorry, Thimble won't work in your Browser
 errorUnsupportedBrowserSuggestion=Thimble works in all of the following browsers: <a title="Download Mozilla Firefox" href="https://www.mozilla.org/{{ locale }}/firefox/new/">Firefox</a>, <a title="Download Google Chrome" href="https://www.google.com/chrome/browser/desktop/index.html">Chrome</a>, Internet Explorer 11, Microsoft Edge, Safari (8+), and <a title="Download Opera" href="http://www.opera.com/download">Opera</a>.
 errorUnsupportedBrowserIgnore=Let me try anyway!
+errorLoadingProjectBramble=For error details open your browser's Developer Tools console.
```
Pomax: locale strings should only be changed in `en-US`, every other locale gets overwritten based on that so your changes here will be lost when you sync locales.

Pomax:
Thanks for the fix!

PR #1821, opened 2017-03-07
change to link to help wiki to address issue 1647- Add a help page

jgmac1106:
Changed the link in line 110 to point toward the GitHub wiki rather than the SUMO page. Will start cloning the Q+A over to SUMo and we can move back if that is the direction team decides.


public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
```diff
@@ -108,7 +108,7 @@ define(function(require) {
     // *******EVENTS
     // User bar menu help
     $("#navbar-help").click(function() {
-      window.open("https://support.mozilla.org/" + locale + "/products/webmaker/thimble");
+      window.open("https://github.com/mozilla/thimble.mozilla.org/wiki/Using-Thimble-FAQ");
```
humphd: What if we open this in a new tab/window vs. the current window?  Add `, "_blank"` after your closing quote.

humphd:
@flukeout can you sign off on this?  I left one comment.

flukeout:
Merged. @humphd - we can make the improvement you suggested in a follow up, or, better yet, we should just make this a normal ``<a>`` element in the markup. Do you know why we did it this way?

Thanks for the change @jgmac1106

Issue #1778, opened 2017-02-23
Create a default index.html when the project folder does not have one


dangkhue27:
This is a follow up from issue #1652. @humphd said:

> This isn't a complete fix, because we probably need to inject a default index.html file in the filesystem if there isn't one, and do it prior to this code being called. However, this will at least stop the crash, and let the initial loading complete.

The solution I am proposing is to create a default `index.html` file [after looking for the index.html](https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/project/project.js#L249). If the file is not found, I will create a default index.html after this code.

Please let me know if I should do this approach or something else.

humphd:
NOTE: we have default files already, see https://github.com/mozilla/thimble.mozilla.org/tree/master/default/empty-project.  We would just need to a) `GET` that file from the server; b) `writeFile` it to the user's filesystem.

I think this is a good idea, go for it.

dangkhue27:
Thank you for the advice. I will start working on it.

humphd:
@dangkhue27 how is this coming?

babhishek21:
I'm working on #1776 , which deals with showing a tutorial file by default over any other html file when a project is loaded.
Now as it stands, this issue was fixed in #1822 by making sure that an `index.html` file is created if none exists.
My question is: What do I do, when `tutorial.html` exists, but `index.html` does not? Do I let the code create an `index.html` file, but still show `tutorial.html`?

humphd:
This was fixed in https://github.com/mozilla/thimble.mozilla.org/pull/1822, closing.

PR #1822, opened 2017-03-07
Fix Issue#1778: Create a default index.html file if there isn't one when the browser refreshes

dangkhue27:
I tried to "ask" Bramble to create the index.html file once it's ready. For this fix, the following message will still shows up when there was no index.html file previously:

image

I also have a different fix that looks something like this:
````
var file = getRoot() + "/index.html";
fs.writeFile(file, data, function(err) {
  if (err) {
    return console.error("Cannot write file to project: " + err);
  }
});
````
This is obviously simpler and probably won't produce the above message. However, I'm not sure why I cannot declare `var fs = require("fs");` in [this file](https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/project/project.js#L2). This is the error message on the browser console:
````
GET http://localhost:3500/en-US/en_US/editor/scripts/editor/js/fs.js 404 (Not Found)
Uncaught Error: Script error
http://requirejs.org/docs/errors.html#scripterror
    at H (require.min.js:9)
    at HTMLScriptElement.onScriptError (require.min.js:32)
````
I am wondering if this is due to the project structure of Thimble or if I am missing something?

### DIFF
--- a/public/editor/scripts/project/project.js
+++ b/public/editor/scripts/project/project.js
@@ -239,37 +239,23 @@ define(function(require) {
           }
 
           // Find the index.html file in the project root to open
-          _fs.readdir(getRoot(), function(err, found) {
-            if(err) {
-              return callback(err);
-            }
-
-            // Try to grab the index.html
-            var indexPos = 0;
-            var foundIndexHTML = false;
-            found.forEach(function(path, idx) {
-              if(Path.basename(path) === "index.html") {
-                indexPos = idx;
-                foundIndexHTML = true;
-              }
-            });
-
-            if (foundIndexHTML) {
-              callback(null, found[indexPos]);
-            }
-
-            // Create a default index.html file
-            if (!foundIndexHTML) {
+          var indexLocation = Path.join(getRoot(),"index.html")
+          _fs.exists(indexLocation, function(exist) {
+            if(exist) {
+              callback(null,indexLocation)
+              return;
+            } else {
+              // Create a default index.html file
               var location = "/default-files/html.txt";
               $.get(location).then(function(data) {
-                var file = getRoot() + "/index.html";
-                _fs.writeFile(file, data, function(err) {
+                _fs.writeFile(indexLocation, data, function(err) {
                   if (err) {
                     console.error("Cannot write file to project: " + err);
                     callback(err);
                     return;
                   }
-                  callback(null, file);
+                  callback(null, indexLocation);
+                  return;
                 });
               }, function(err) {
                 if (err) {
--- a/public/editor/scripts/project/project.js
+++ b/public/editor/scripts/project/project.js
@@ -255,6 +255,10 @@ define(function(require) {
               }
             });
 
+            if (foundIndexHTML) {
+              callback(null, found[indexPos]);
+            }
+
             // Create a default index.html file
             if (!foundIndexHTML) {
               var location = "/default-files/html.txt";
@@ -262,8 +266,11 @@ define(function(require) {
                 var file = getRoot() + "/index.html";
                 _fs.writeFile(file, data, function(err) {
                   if (err) {
-                    return console.error("Cannot write file to project: " + err);
+                    console.error("Cannot write file to project: " + err);
+                    callback(err);
+                    return;
                   }
+                  callback(null, file);
                 });
               }, function(err) {
                 if (err) {
@@ -274,7 +281,6 @@ define(function(require) {
                 callback();
               });
             }
-            callback(null, found[indexPos]);
           });
         });
       });
--- a/public/editor/scripts/project/project.js
+++ b/public/editor/scripts/project/project.js
@@ -239,10 +239,10 @@ define(function(require) {
           }
 
           // Find the index.html file in the project root to open
-          var indexLocation = Path.join(getRoot(),"index.html")
+          var indexLocation = Path.join(getRoot(),"index.html");
           _fs.exists(indexLocation, function(exist) {
             if(exist) {
-              callback(null,indexLocation)
+              callback(null,indexLocation);
               return;
             } else {
               // Create a default index.html file
--- a/public/editor/scripts/project/project.js
+++ b/public/editor/scripts/project/project.js
@@ -263,16 +263,11 @@ define(function(require) {
               };
               var location = "/default-files/html.txt";
               $.get(location).then(function(data) {
-                fileOptions.contents = data;
-                Bramble.once("ready", function(bramble) {
-                  bramble.addNewFile(fileOptions, function(err) {
-                    if (err) {
-                      console.error("[Bramble] Failed to write new file", err);
-                      callback(err);
-                      return;
-                    }
-                    callback();
-                  });
+                var file = getRoot() + "/index.html";
+                _fs.writeFile(file, data, function(err) {
+                  if (err) {
+                    return console.error("Cannot write file to project: " + err);
+                  }
                 });
               }, function(err) {
                 if (err) {
--- a/public/editor/scripts/project/project.js
+++ b/public/editor/scripts/project/project.js
@@ -264,7 +264,6 @@ define(function(require) {
                   if (err) {
                     return console.error("Cannot write file to project: " + err);
                   }
-                  callback();
                 });
               }, function(err) {
                 if (err) {
--- a/public/editor/scripts/project/project.js
+++ b/public/editor/scripts/project/project.js
@@ -239,33 +239,32 @@ define(function(require) {
           }
 
           // Find the index.html file in the project root to open
-          var indexLocation = Path.join(getRoot(),"index.html");
-          _fs.exists(indexLocation, function(exist) {
-            if(exist) {
-              callback(null,indexLocation);
+          var indexLocation = Path.join(getRoot(), "index.html");
+          _fs.exists(indexLocation, function(exists) {
+            if(exists) {
+              callback(null, indexLocation);
               return;
-            } else {
-              // Create a default index.html file
-              var location = "/default-files/html.txt";
-              $.get(location).then(function(data) {
-                _fs.writeFile(indexLocation, data, function(err) {
-                  if (err) {
-                    console.error("Cannot write file to project: " + err);
-                    callback(err);
-                    return;
-                  }
-                  callback(null, indexLocation);
-                  return;
-                });
-              }, function(err) {
+            }
+            // Create a default index.html file
+            var location = "/default-files/html.txt";
+            $.get(location).then(function(data) {
+              _fs.writeFile(indexLocation, data, function(err) {
                 if (err) {
-                  console.error("Failed to download " + location, err);
+                  console.error("Cannot write file to project: " + err);
                   callback(err);
                   return;
                 }
-                callback();
+                callback(null, indexLocation);
+                return;
               });
-            }
+            }, function(err) {
+              if (err) {
+                console.error("Failed to download " + location, err);
+                callback(err);
+                return;
+              }
+              callback();
+            });
           });
         });
       });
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by humphd** [comment]
You get the `fs` like this: https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/project/project.js#L145

**Comment by dangkhue27** [comment]
@humphd Thank you for your help!

File: public/editor/scripts/project/project.js [hunk 1]
```diff
@@ -246,12 +247,33 @@ define(function(require) {
...
 
+            // Create a default index.html file
+            if (!foundIndexHTML) {
+              var location = "/default-files/html.txt";
+              $.get(location).then(function(data) {
+                var file = getRoot() + "/index.html";
+                _fs.writeFile(file, data, function(err) {
+                  if (err) {
+                    return console.error("Cannot write file to project: " + err);
+                  }
```
humphd: You'd still need a `callback();` after this, so that it doesn't fail to continue if it does write the file.

humphd:
I've looked over the code, and made a comment.  I'd appreciate a review from @gideonthomas on the overall method.

public/editor/scripts/project/project.js
```diff
@@ -1,4 +1,5 @@
 define(function(require) {
+  var $ = require("jquery");
```
Pomax: this should not be necessary: we already have jquery listed as a "known library" so you can instead say:
```
define(['jQuery'], function($) {
 ....
});
```
and have jQuery available that way.

humphd: It looks like we do it the first way throughout the rest of the Thimble code: https://github.com/mozilla/thimble.mozilla.org/search?q=jquery&type=Code&utf8=%E2%9C%93

public/editor/scripts/project/project.js
dangkhue27: I just re-tested the code and noticed that when I had the `callback()`, it created an error in the console like this:
screen shot 2017-03-14 at 10 22 25 am

With this screen showed up after the thimble loading screen for a split second and it disappeared:
screen shot 2017-03-14 at 11 21 57 am

Therefore, I don't think we need a `callback()` here. The code will still work without it.

humphd: It's a bit more subtle than that--we do need it for sure, but we also need to call it in two spots, and we have to pass along the path to the file to open.  The block of code should look something like this (I haven't tested this code, but it's the basic idea you want):

```js
            if(foundIndexHTML) {
              callback(null, found[indexPos]);
              return;
            }

            // Create a default index.html file since we didn't find any
            var location = "/default-files/html.txt";
            $.get(location).then(function(data) {
              var file = getRoot() + "/index.html";
              _fs.writeFile(file, data, function(err) {
                if (err) {
                  console.error("Cannot write file to project: " + err);
                  return callback(err);
                }
                callback(null, file);
              });
            }, function(err) {
              if (err) {
                console.error("Failed to download " + location, err);
                callback(err);
                return;
              }
            });
```

public/editor/scripts/project/project.js
Pomax: Hmm... I would recommend we stop doing that, but if we want to keep this PR in line with the rest of the code that's fine with me.

public/editor/scripts/project/project.js
dangkhue27: I understand what you mean now and I just made the change to the code. Can you take a look?

flukeout:
Testing now.

flukeout:
Okay, this works mostly great. One issue I found is that if I have a folder with an index.html in it, like ``testfolder/index.html`` it will not longer add an index.html page to the root of the project, but I think it should.

humphd:
We could fix the problem of deep-index.html files being found but not one in the root by not doing an `sh.find` that's recursive here https://github.com/mozilla/thimble.mozilla.org/pull/1822/files#diff-eac81253dbbab1c9d63e6efd5fb4768bL242 and instead doing an `fs.readdir()` just on the root dir.

dangkhue27:
@flukeout I've changed code so it is now only looks for the index.html in the project root and will create one if there isn't one, will also ignore index.html within a folder.

public/editor/scripts/project/project.js
```diff
@@ -237,22 +238,48 @@ define(function(require) {
...
             if(err) {
               return callback(err);
             }
 
-            // Look for an HTML file to open, ideally index.html
+            // Try to grab the index.html
             var indexPos = 0;
+            var foundIndexHTML = false;
             found.forEach(function(path, idx) {
               if(Path.basename(path) === "index.html") {
```
humphd: You can skip this check, since `readdir()` will return only basenames without parent dir info.  Also, you don't need both `indexPos` and `foundIndexHTML`.  Just set `indexPos = -1` when you declare it, then you can see if it's greater than `-1` on 257 below.  Better to reduce your state and drop a variable vs. risking a bug by using 2 to do the job of 1.

public/editor/scripts/project/project.js
```diff
@@ -237,22 +238,48 @@ define(function(require) {
             return callback(err);
           }
 
-          // Find an HTML file to open in the project, hopefully /index.html
-          var sh = new _fs.Shell();
-          sh.find(getRoot(), {name: "*.html"}, function(err, found) {
+          // Find the index.html file in the project root to open
+          _fs.readdir(getRoot(), function(err, found) {
```
humphd: I would rename `found` to `children`, since `readdir()` will return all children of a dir vs. some query (like `find` does).

public/editor/scripts/project/project.js
```diff
@@ -237,22 +238,48 @@ define(function(require) {
...
             found.forEach(function(path, idx) {
               if(Path.basename(path) === "index.html") {
                 indexPos = idx;
+                foundIndexHTML = true;
               }
             });
 
-            callback(null, found[indexPos]);
+            if (foundIndexHTML) {
+              callback(null, found[indexPos]);
```
humphd: You'll need a `return` here too, or you'll also run all the code below.

public/editor/scripts/project/project.js
humphd: I'm not thinking clearly...the right thing to do here is `_fs.exists(Path.join(getRoot(), "index.html"), function(exists) {` which will replace all the variables and your loop, and just report on whether this file is there.

dangkhue27:
Referencing #1778
@humphd I just made the change following your instruction. 
I noticed that the `exists` function is deprecated on [Node.js documentation](https://nodejs.org/api/fs.html#fs_fs_exists_path_callback). I wonder if I am actually calling that function indirectly? I followed the source code of this function but didn't understand at all.

public/editor/scripts/project/project.js
```diff
@@ -237,22 +238,34 @@ define(function(require) {
...
-          sh.find(getRoot(), {name: "*.html"}, function(err, found) {
-            if(err) {
-              return callback(err);
+          // Find the index.html file in the project root to open
+          var indexLocation = Path.join(getRoot(),"index.html");
+          _fs.exists(indexLocation, function(exist) {
+            if(exist) {
+              callback(null,indexLocation);
+              return;
+            } else {
```
humphd: No-else-after-return:

```js
if(exists) {
    callback(null, indexLocation);
    return;
}

// Continue with your code here.
```

public/editor/scripts/project/project.js
```diff
@@ -237,22 +238,34 @@ define(function(require) {
             return callback(err);
           }
 
-          // Find an HTML file to open in the project, hopefully /index.html
-          var sh = new _fs.Shell();
-          sh.find(getRoot(), {name: "*.html"}, function(err, found) {
-            if(err) {
-              return callback(err);
+          // Find the index.html file in the project root to open
+          var indexLocation = Path.join(getRoot(),"index.html");
```
humphd: You need spaces after your commas

public/editor/scripts/project/project.js
```diff
@@ -237,22 +238,34 @@ define(function(require) {
...
+          _fs.exists(indexLocation, function(exist) {
```
humphd: The variable should be named `exists` not `exist`.  The node.js code for `fs.exists` is deprecated, but we aren't actually using node's code. We wrote a library for the browser that acts like node's `fs` but is different.  You could also use `fs.stat`, but I think it's fine to just do `fs.exists` instead.

public/editor/scripts/project/project.js
```diff
@@ -237,22 +238,34 @@ define(function(require) {
...
+            if(exist) {
+              callback(null,indexLocation);
```
humphd: Space after `,`

dangkhue27:
@humphd I have made the changes. Please have a look!

dsych:
@dangkhue27 I am actually working on the API that will allow you to query brackets for project metadata like existence of index.html in project root, project size, file count, etc. Take a look, that might help you : https://github.com/mozilla/brackets/pull/620

public/editor/scripts/project/project.js
```diff
@@ -237,22 +238,33 @@ define(function(require) {
...
+            var location = "/default-files/html.txt";
+            $.get(location).then(function(data) {
+              _fs.writeFile(indexLocation, data, function(err) {
+                if (err) {
+                  console.error("Cannot write file to project: " + err);
+                  callback(err);
+                  return;
+                }
+                callback(null, indexLocation);
+                return;
```
humphd: This `return` isn't needed, because you're already at the end of the function.

public/editor/scripts/project/project.js
```diff
@@ -237,22 +238,33 @@ define(function(require) {
...
+              return;
             }
-
-            // Look for an HTML file to open, ideally index.html
-            var indexPos = 0;
-            found.forEach(function(path, idx) {
-              if(Path.basename(path) === "index.html") {
-                indexPos = idx;
+            // Create a default index.html file
+            var location = "/default-files/html.txt";
```
humphd: Move this up to the top of the file, and rename it to `DEFAULT_INDEX_HTML_URL`

dangkhue27:
@humphd I hoped that's all the change we need

public/editor/scripts/project/project.js
```diff
@@ -237,22 +240,31 @@ define(function(require) {
...
+            $.get(DEFAULT_INDEX_HTML_URL).then(function(data) {
+              _fs.writeFile(indexLocation, data, function(err) {
+                if (err) {
+                  console.error("Cannot write file to project: " + err);
+                  callback(err);
+                  return;
+                }
+                callback(null, indexLocation);
+              });
+            }, function(err) {
```
gideonthomas: there's no need to log here because:
1. The log doesn't really show anything too useful.
2. The browser will automatically log network failures.

Instead of this function, I would just pass `callback` here directly.

humphd:
@dangkhue27 can you finish this soon?  It would be great to get it in.

dangkhue27:
I hope that's all the change needed

public/editor/scripts/project/project.js
```diff
@@ -259,7 +259,6 @@ define(function(require) {
               });
             }, function(err) {
               if (err) {
-                console.error("Failed to download " + DEFAULT_INDEX_HTML_URL, err);
```
humphd: No, this still isn't right.  What @gideonthomas asked you to do above is to change this:

```js
}, function(err) {
  if(err) {
    return callback(err);
  }
  callback();
});
```

Into this:

```js
}, callback);
```

The only reason to "wrap" the callback would be if you wanted to do something special (like log or handle the error somehow).  In this case, you just want to pass it on unchanged.  This is a common anti-pattern in node.js callback code.

humphd:
@dangkhue27 looks like you fixed this, you ready for me to merge this?

dangkhue27:
Thank you for your clarification. I am ready to merge.

PR #1825, opened 2017-03-07
fix: Remove jQuery dependency from injected remix button code

### LINKED ISSUE #1327: Remove jQuery dependency from injected remix button code
It would be nice if we could remove the jQuery-dependency of the [injected remix button code](https://thimble.mozilla.org/resources/remix/index.js) from published code, as it interferes non-trivially with jQuery loading (from user code).

We don't use any jQ plugins and that source is not so vast that and complicated taht this should be an issue. Also we would win on performance, reduce page load time and size, and favoring the use of vanilla js would be in line with general Mozilla principles, so it's a win on multiple grounds.

If this is deemed okay, I would like to have a stab at this at [`vanilla`](https://github.com/flaki/thimble.webmaker.org/tree/vanilla) branch.

aayushsanghavi:
Fixes #1327 
Changed jQuery dependent code to plain JS code and used DOM methods in some places

### DIFF
--- a/public/resources/remix/index.js
+++ b/public/resources/remix/index.js
@@ -55,6 +55,7 @@
 
   function injectDetailsBar(metadata, callback) {
     var xmlhttp = new XMLHttpRequest();
+    xmlhttp.responseType = "text";
     var url = metadata.host + "/projects/remix-bar";
     var data = {
       host: metadata.host,
@@ -64,12 +65,12 @@
       updated: metadata.dateUpdated
     };
 
-    xmlhttp.onreadystatechange = function(response, textStatus) {
+    xmlhttp.onreadystatechange = function() {
       if (xmlhttp.readyState == XMLHttpRequest.DONE ) {
+	  var response = xmlhttp.response;
+	  var textStatus = xmlhttp.statusText;
         if (xmlhttp.status == 200) {
-          var responseTag = document.createElement("span");
-          responseTag.innerHTML = response;
-          document.body.insertBefore(responseTag, document.body.firstChild);
+          document.body.insertBefore(response, document.body.firstChild);
 	    callback(null);
         }
         else if (xmlhttp.status == 400) {
--- a/public/resources/remix/index.js
+++ b/public/resources/remix/index.js
@@ -70,7 +70,7 @@
     xmlhttp.onreadystatechange = function() {
       if (xmlhttp.readyState == XMLHttpRequest.DONE ) {
         if (xmlhttp.status == 200) {
-          body.insertBefore(response, body.firstChild);
+          document.body.insertBefore(response, document.body.firstChild);
 	    callback(null);
         }
         else if (xmlhttp.status == 400) {
@@ -106,7 +106,7 @@
     var metadata = {};
     var metaTags = document.getElementsByTagName("meta");
 
-    var set = grep(metaTags, function(metaTag) {
+    grep(metaTags, function(metaTag) {
       return /^data-remix-.+/.test((metaTag).getAttribute("name"));
     })
     .forEach(function(metaTag) {
@@ -131,7 +131,14 @@
         injectStyleSheets(metadata);
         injectDetailsBar(metadata, setupBar);
       }
-    }
+    };
   }
   
+  (function(doc, script) {
+    script = doc.createElement("script");
+    script.type = "text/javascript";
+    script.onload = run;
+    doc.getElementsByTagName("head")[0].appendChild(script);
+  }(document));
+
 }(document, document.head));
--- a/public/resources/remix/index.js
+++ b/public/resources/remix/index.js
@@ -14,43 +14,44 @@
 
   function setupBar(err) {
     if (err) {
-      console.log("[Thimble Error] Unable to inject Remix UI. Error was: `" + err + "`");
+      console.log("[Thimble Error] Unable to inject Remix UI. Error was: ", err);
       return;
     }
 
     var detailsBar = document.querySelector(".details-bar");
     
-    if (detailsBar) {
-      detailsBar.setAttribute("style", "");
-      detailsBar.classList.add("mouse-mode");
-
-      detailsBar.addEventListener("click", function(event) {
-        if (event.target.classList.contains("thimble-button")) {
-          detailsBar.classList.remove("collapsed");
-          event.stopPropagation();
-        }
-      });
-
-      detailsBar.addEventListener("click", function(event) {
-        if (event.target.classList.contains("close-details-bar")) {
-          detailsBar.classList.add("collapsed");
-          event.stopPropagation();
-        }
-      });
-
-      document.querySelector(".mouse-mode").addEventListener("mouseenter", function(){
-        detailsBar.classList.remove("collapsed");
-      });
-
-      document.querySelector(".mouse-mode").addEventListener("mouseleave", function(){
-        detailsBar.classList.add("collapsed");
-      });
+    if (!detailsBar) {
+      return;
     }
+
+    detailsBar.setAttribute("style", "");
+    detailsBar.classList.add("mouse-mode");
+
+    detailsBar.addEventListener("click", function(event) {
+      if (event.target.classList.contains("thimble-button")) {
+        detailsBar.classList.remove("collapsed");
+        event.stopPropagation();
+      }
+    });
+
+    detailsBar.addEventListener("click", function(event) {
+      if (event.target.classList.contains("close-details-bar")) {
+        detailsBar.classList.add("collapsed");
+        event.stopPropagation();
+      }
+    });
+
+    document.querySelector(".mouse-mode").addEventListener("mouseenter", function(){
+      detailsBar.classList.remove("collapsed");
+    });
+
+    document.querySelector(".mouse-mode").addEventListener("mouseleave", function(){
+      detailsBar.classList.add("collapsed");
+    });
   }
 
   function injectDetailsBar(metadata, callback) {
-    var xmlhttp = new XMLHttpRequest();
-    xmlhttp.responseType = "text";
+    var request = new XMLHttpRequest();
     var url = metadata.host + "/projects/remix-bar";
     var data = {
       host: metadata.host,
@@ -60,26 +61,23 @@
       updated: metadata.dateUpdated
     };
 
-    xmlhttp.onreadystatechange = function() {
-      if (xmlhttp.readyState === XMLHttpRequest.DONE ) {
-        var response = xmlhttp.response;
-        var textStatus = xmlhttp.statusText;
-        if (xmlhttp.status === 200) {
-          var fragmentBody = document.createElement("body");
-          var docFragment = document.createDocumentFragment();
-          docFragment.appendChild(fragmentBody);
-          fragmentBody.innerHTML = response;
-          var remixBar = fragmentBody.querySelector("div");
-          document.body.appendChild(remixBar);
-          callback(null);
-        } else if (xmlhttp.status >= 400) {
-          callback(textStatus);
+    request.onreadystatechange = function() {
+      if (request.readyState === XMLHttpRequest.DONE ) {
+        var response = request.response;
+        if (request.status !== 200) {
+          return callback({ status: request.status, response: response });
         }
+        document.querySelector("body").insertAdjacentHTML("afterbegin", response);
+        callback(null);
       }
     };
 
-    xmlhttp.open("GET", url, true);
-    xmlhttp.send(data);
+    var querystring = Object.keys(data).map(function(key) {
+      return encodeURIComponent(key) + "=" + encodeURIComponent(data[key]);
+    }).join("&");
+    url = url +  "?" +querystring;
+    request.open("GET", url);
+    request.send(data);
   }
 
   function injectStyleSheets(metadata) {
@@ -94,13 +92,10 @@
     document.head.appendChild(stylesheet);
   }
 
-  var hasDataRemixAttribute = function(element) {
+  function hasDataRemixAttribute(element) {
     var name = element.getAttribute("name");
-    if (!name) {
-      return false;
-    }
-    return name.startsWith("data-remix-");
-  };
+    return /^data-remix-.+/.test(name);
+  }
 
   function getMetadata() {
     var metadata = {};
--- a/public/resources/remix/index.js
+++ b/public/resources/remix/index.js
@@ -67,7 +67,7 @@
       updated: metadata.dateUpdated
     };
 
-    xmlhttp.onreadystatechange = function() {
+    xmlhttp.onreadystatechange = function(response, textStatus) {
       if (xmlhttp.readyState == XMLHttpRequest.DONE ) {
         if (xmlhttp.status == 200) {
           document.body.insertBefore(response, document.body.firstChild);
--- a/public/resources/remix/index.js
+++ b/public/resources/remix/index.js
@@ -7,16 +7,9 @@
   var gaTrackingId = 'UA-68630113-1';
 
   function injectAnalytics() {
-    var analyticsHtml =
-    '\n<script>\n' +
-    '(function(i,s,o,g,r,a,m){i[\'GoogleAnalyticsObject\']=r;i[r]=i[r]||function(){\n' +
-    '(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),\n' +
-    'm=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)\n' +
-    '})(window,document,\'script\',\'//www.google-analytics.com/analytics.js\',\'ga\');\n' +
-    'ga(\'create\', \'' + gaTrackingId + '\', \'auto\');\n' +
-    'ga(\'send\', \'pageview\');\n' +
-    '</script>\n';
-
+    var analyticsHtml = document.createElement("script");
+    analyticsHtml.type = "text/javascript";
+    analyticsHtml.innerHTML = '(function(i,s,o,g,r,a,m){i[\'GoogleAnalyticsObject\']=r;i[r]=i[r]||function(){\n' + '(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),\n' + 'm=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)\n' + '})(window,document,\'script\',\'//www.google-analytics.com/analytics.js\',\'ga\');\n' + 'ga(\'create\', \'' + gaTrackingId + '\', \'auto\');\n' + 'ga(\'send\', \'pageview\');\n';
     head.appendChild(analyticsHtml);
   }
 
@@ -27,7 +20,7 @@
     }
 
     var isTouchDevice = 'ontouchstart' in document.documentElement;
-    var detailsBar = document.getElementsByClassName("details-bar");
+    var detailsBar = document.querySelector(".details-bar");
     detailsBar.setAttribute("style", "");
 
     if(isTouchDevice) {
@@ -37,21 +30,25 @@
       detailsBar.classList.add("mouse-mode");
     }
 
-    detailsBar.on("click", ".thimble-button", function(){
+    detailsBar.addEventListener("click", function(event) {
+      if (event.target.className == "thimble-button") {
+        detailsBar.classList.remove("collapsed");
+        return false;
+      }
+    });
+
+    detailsBar.addEventListener("click", function(event) {
+      if (event.target.className == "close-details-bar") {
+        detailsBar.classList.add("collapsed");
+        return false;
+      }
+    });
+
+    document.querySelector(".mouse-mode").addEventListener("mouseenter", function(){
       detailsBar.classList.remove("collapsed");
-      return false;
     });
 
-    detailsBar.on("click", ".close-details-bar", function(){
-      detailsBar.classList.add("collapsed");
-      return false;
-    });
-
-    document.getElementsByClassName("mouse-mode").on("mouseenter",function(){
-      detailsBar.classList.remove("collapsed");
-    });
-
-    document.getElementsByClassName("mouse-mode").on("mouseleave",function(){
+    document.querySelector(".mouse-mode").addEventListener("mouseleave", function(){
       detailsBar.classList.add("collapsed");
     });
   }
@@ -70,7 +67,9 @@
     xmlhttp.onreadystatechange = function(response, textStatus) {
       if (xmlhttp.readyState == XMLHttpRequest.DONE ) {
         if (xmlhttp.status == 200) {
-          document.body.insertBefore(response, document.body.firstChild);
+          var responseTag = document.createElement("span");
+          responseTag.innerHTML = response;
+          document.body.insertBefore(responseTag, document.body.firstChild);
 	    callback(null);
         }
         else if (xmlhttp.status == 400) {
@@ -87,7 +86,7 @@
     var stylesheets =
       "<link href=\"" + metadata.host + "/resources/remix/clean-slate.min.css\" rel=\"stylesheet\">\n" +
       "<link href=\"" + metadata.host + "/resources/remix/style.css\" rel=\"stylesheet\">\n";
-    document.getElementsByTagName("head").appendChild(stylesheets);
+    document.head.appendChild(stylesheets);
   }
 
   function grep(items, callback) {
@@ -124,14 +123,12 @@
   }
 
   function run() {
-    document.onreadystatechange = function () {
-      if (document.readyState === "complete") {
-        var metadata = getMetadata();
-        injectAnalytics();
-        injectStyleSheets(metadata);
-        injectDetailsBar(metadata, setupBar);
-      }
-    };
+    document.addEventListener("DOMContentLoaded", function() {
+      var metadata = getMetadata();
+      injectAnalytics();
+      injectStyleSheets(metadata);
+      injectDetailsBar(metadata, setupBar);
+    });
   }
   
   (function(doc, script) {
--- a/public/resources/remix/index.js
+++ b/public/resources/remix/index.js
@@ -7,10 +7,9 @@
   var gaTrackingId = 'UA-68630113-1';
 
   function injectAnalytics() {
-    var analyticsHtml = document.createElement("script");
-    analyticsHtml.type = "text/javascript";
-    analyticsHtml.innerHTML = '(function(i,s,o,g,r,a,m){i[\'GoogleAnalyticsObject\']=r;i[r]=i[r]||function(){\n' + '(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),\n' + 'm=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)\n' + '})(window,document,\'script\',\'//www.google-analytics.com/analytics.js\',\'ga\');\n' + 'ga(\'create\', \'' + gaTrackingId + '\', \'auto\');\n' + 'ga(\'send\', \'pageview\');\n';
-    head.appendChild(analyticsHtml);
+    var analytics = document.createElement("script");
+    analytics.textContent = '(function(i,s,o,g,r,a,m){i[\'GoogleAnalyticsObject\']=r;i[r]=i[r]||function(){\n' + '(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),\n' + 'm=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)\n' + '})(window,document,\'script\',\'//www.google-analytics.com/analytics.js\',\'ga\');\n' + 'ga(\'create\', \'' + gaTrackingId + '\', \'auto\');\n' + 'ga(\'send\', \'pageview\');\n';
+    head.appendChild(analytics);
   }
 
   function setupBar(err) {
@@ -21,36 +20,37 @@
 
     var isTouchDevice = 'ontouchstart' in document.documentElement;
     var detailsBar = document.querySelector(".details-bar");
-    detailsBar.setAttribute("style", "");
+    
+    if (detailsBar) {
+      detailsBar.setAttribute("style", "");
+      if (isTouchDevice) {
+        detailsBar.classList.add("touch-mode");
+      } else {
+        detailsBar.classList.add("mouse-mode");
+      }
 
-    if(isTouchDevice) {
-      detailsBar.classList.add("touch-mode");
-    }
-    else {
-      detailsBar.classList.add("mouse-mode");
-    }
+      detailsBar.addEventListener("click", function(event) {
+        if (event.target.classList.contains("thimble-button")) {
+          detailsBar.classList.remove("collapsed");
+          event.stopPropagation();
+        }
+      });
 
-    detailsBar.addEventListener("click", function(event) {
-      if (event.target.className == "thimble-button") {
+      detailsBar.addEventListener("click", function(event) {
+        if (event.target.classList.contains("close-details-bar")) {
+          detailsBar.classList.add("collapsed");
+          event.stopPropagation();
+        }
+      });
+
+      document.querySelector(".mouse-mode").addEventListener("mouseenter", function(){
         detailsBar.classList.remove("collapsed");
-        return false;
-      }
-    });
+      });
 
-    detailsBar.addEventListener("click", function(event) {
-      if (event.target.className == "close-details-bar") {
+      document.querySelector(".mouse-mode").addEventListener("mouseleave", function(){
         detailsBar.classList.add("collapsed");
-        return false;
-      }
-    });
-
-    document.querySelector(".mouse-mode").addEventListener("mouseenter", function(){
-      detailsBar.classList.remove("collapsed");
-    });
-
-    document.querySelector(".mouse-mode").addEventListener("mouseleave", function(){
-      detailsBar.classList.add("collapsed");
-    });
+      });
+    }
   }
 
   function injectDetailsBar(metadata, callback) {
@@ -66,13 +66,13 @@
     };
 
     xmlhttp.onreadystatechange = function() {
-      if (xmlhttp.readyState == XMLHttpRequest.DONE ) {
+      if (xmlhttp.readyState === XMLHttpRequest.DONE ) {
         var response = xmlhttp.response;
         var textStatus = xmlhttp.statusText;
-        if (xmlhttp.status == 200) {
+        if (xmlhttp.status === 200) {
           document.body.insertBefore(response, document.body.firstChild);
 	   callback(null);
-        } else if (xmlhttp.status == 400) {
+        } else if (xmlhttp.status >= 400) {
 	   callback(textStatus);
         }
       }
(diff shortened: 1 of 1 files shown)


public/resources/remix/index.js
```diff
@@ -17,110 +17,127 @@
     'ga(\'send\', \'pageview\');\n' +
     '</script>\n';
 
-    $(head).append(analyticsHtml);
+    head.appendChild(analyticsHtml);
```
humphd: This seems wrong to me, since `appendChild()` expects a node vs. a string of HTML.  You'll first need to create a `script` element, and then set its content

public/resources/remix/index.js
```diff
@@ -17,110 +17,127 @@
...
+  function setupBar(err) {
     if(err) {
       console.log("[Thimble Error] Unable to inject Remix UI. Error was: `" + err + "`");
       return;
     }
 
     var isTouchDevice = 'ontouchstart' in document.documentElement;
-    var detailsBar = $(".details-bar");
-    detailsBar.attr("style", "");
+    var detailsBar = document.getElementsByClassName("details-bar");
```
humphd: Prever `querySelector()` here so we get just one single element.

public/resources/remix/index.js
```diff
@@ -17,110 +17,127 @@
...
-      detailsBar.addClass("mouse-mode");
+      detailsBar.classList.add("touch-mode");
+    }
+    else {
+      detailsBar.classList.add("mouse-mode");
     }
 
-    detailsBar.on("click", ".thimble-button",function(){
-      detailsBar.removeClass("collapsed");
+    detailsBar.on("click", ".thimble-button", function(){
```
humphd: You'll have to use `addEventListener` vs. `on`

public/resources/remix/index.js
```diff
@@ -17,110 +17,127 @@
...
 
     detailsBar.on("click", ".close-details-bar", function(){
-      detailsBar.addClass("collapsed");
+      detailsBar.classList.add("collapsed");
       return false;
     });
 
-    $(".mouse-mode").on("mouseenter",function(){
-      detailsBar.removeClass("collapsed");
+    document.getElementsByClassName("mouse-mode").on("mouseenter",function(){
```
humphd: Again, use `querySelector` and use `addEventListener`

public/resources/remix/index.js
```diff
@@ -17,110 +17,127 @@
...
+      id: metadata.projectId,
+      title: metadata.projectTitle,
+      author: metadata.projectAuthor,
+      updated: metadata.dateUpdated
+    };
+
+    xmlhttp.onreadystatechange = function(response, textStatus) {
+      if (xmlhttp.readyState == XMLHttpRequest.DONE ) {
+        if (xmlhttp.status == 200) {
+          document.body.insertBefore(response, document.body.firstChild);
```
humphd: `insertBefore` expects a node not a response string.

public/resources/remix/index.js
```diff
@@ -17,110 +17,127 @@
...
+    xmlhttp.send(data);
   }
 
-  function injectStyleSheets($, metadata) {
+  function injectStyleSheets(metadata) {
     var stylesheets =
       "<link href=\"" + metadata.host + "/resources/remix/clean-slate.min.css\" rel=\"stylesheet\">\n" +
       "<link href=\"" + metadata.host + "/resources/remix/style.css\" rel=\"stylesheet\">\n";
-    $("head").append(stylesheets);
+    document.getElementsByTagName("head").appendChild(stylesheets);
```
humphd: `document.head`

public/resources/remix/index.js
```diff
@@ -17,110 +17,127 @@
...
   function run() {
-    var $$ = $.noConflict(true);
-    $$(document).ready(function($) {
-      var metadata = getMetadata($);
-      injectAnalytics($);
-      injectStyleSheets($, metadata);
-      injectDetailsBar($, metadata, setupBar);
-    });
+    document.onreadystatechange = function () {
+      if (document.readyState === "complete") {
```
humphd: What about `DOMContentLoaded` instead of all this?

public/resources/remix/index.js
aayushsanghavi: Should I wrap the response string in a span or p tag?

aayushsanghavi:
@humphd thanks a lot for the review. I'll get back to you with the changes in some time.

aayushsanghavi:
@humphd I've made the changes. How do I test them? I had run "npm test" earlier too and it said "done without any errors". Is it that npm test checks only the linting and syntactical issues?

aayushsanghavi:
@humphd what do you think about the edits I made?

humphd:
In order to test this, you need to get Thimble + Brackets running locally, and then create a project from scratch, then publish it, and try to open that published project so you can see what the remix stuff looks like, and how it works.  Could you try that and report back?  I'll get @flukeout to also review your changes.

aayushsanghavi:
Yes definitely. I'll do that and get back to you in some time.

aayushsanghavi:
@humphd when I click on the sign in link to be able to publish a project I am redirected to a blank page with title webmaker login v4 instead of a log in form. What should I do? (I run the project on vagrant)

flukeout:
That's (unfortunatly) the correct sign in page, can you try to create an account there and then log in with it?

aayushsanghavi:
@flukeout I created an account on thimble before trying to sign in on localhost:3500 but it didn't work.

flukeout:
Did you create an account on ``thimble.mozilla.org`` or on your local environment?

aayushsanghavi:
I tried both. When I tried to create an account on the localhost server I was redirected to the same blank page as I was when I tried to sign in through localhost.

flukeout:
@aayushsanghavi I'll try it now locally and get back to you

flukeout:
So if you click "Create new Account" on localhost:3500 you dont' see this page?
image

aayushsanghavi:
Nope. I see a blank page

flukeout:
Paging @gideonthomas - any ideas?

aayushsanghavi:
@flukeout could it have anything (even remotely) to do with the fact that I didn't set up the webmaker repository according to the manual installation and used vagrant instead?

flukeout:
I'm using the basic vagrant setup and things are working fine, do you want to jump into the Thimble channel so the others can help us troubleshoot this?

https://chat.mozillafoundation.org/mozilla/channels/thimble

humphd:
@aayushsanghavi using the Vagrant setup is ideal, and you should continue to do it.  I'd like you to try re-provisioning it though, in case something failed:

```
$ vagrant destroy
$ vagrant up
```

That will force it to rebuild, and then you can try this again.  Let us know what happens.

flukeout:
Trying it out now

flukeout:
@aayushsanghavi have you been able to sign in locally and test this patch yet? I'm running it now and it's not working as intended. Let me know when you've had a chance to test it yourself and then we can troubleshoot together.

flukeout:
For a start, the ``script`` in block at the bottom of the file never 'loads' anymore because you've removed the ``src=`` attribute that used to load jQuery, so then the ``run`` function that sets up the rest of the remix stuff never runs...

```
(function(doc, script) {
    script = doc.createElement("script");
    script.type = "text/javascript";
    script.onload = run;
    doc.getElementsByTagName("head")[0].appendChild(script);
  }(document));
```

aayushsanghavi:
@flukeout thanks for trying it out.
I'm working on that. By tomorrow I'll try to set up thimble again outside of any proxy and test this code.

flukeout:
Sounds good!

aayushsanghavi:
@flukeout I can now login and publish projects (the set up went smoothly). How do I test this code?
Also, for some reason the existing projects that we can remix, aren't being displayed. It says loading projects ... but doesn't load them.

flukeout:
@aayushsanghavi - when you publish a project in your local environment, are you able to see the Remix button in the top right corner when you're running the code in this PR?

aayushsanghavi:
@flukeout No. I can't see it. So that means it isn't working correctly.

Pomax:
as I recall this got resolved in the chat - are you still blocked on this @aayushsanghavi?

public/resources/remix/index.js
```diff
@@ -83,10 +83,15 @@
   }
 
   function injectStyleSheets(metadata) {
-    var stylesheets =
-      "<link href=\"" + metadata.host + "/resources/remix/clean-slate.min.css\" rel=\"stylesheet\">\n" +
-      "<link href=\"" + metadata.host + "/resources/remix/style.css\" rel=\"stylesheet\">\n";
-    document.head.appendChild(stylesheets);
+    var stylesheet1 = document.createElement("link");
+    stylesheet1.rel = "stylesheet";
+    stylesheet1.href = "\"" + metadata.host + "/resources/remix/clean-slate.min.css\"";
```
Pomax: you don't need quotes inside an href, so just `metadata.host + "/resources/remix/clean-slate.min.css"`. The quotes above are only because the HTML code needs quotes in it, and is itself built as a string in JavaScript so already has quotes around it =)

public/resources/remix/index.js
```diff
@@ -83,10 +83,15 @@
...
+    document.head.appendChild(stylesheet1);
+
+    var stylesheet2 = document.createElement("link");
+    stylesheet2.rel = "stylesheet";
+    stylesheet2.href = "\"" + metadata.host + metadata.host + "/resources/remix/style.css\"";
```
Pomax: same note here, no quotes inside the href.

public/resources/remix/index.js
```diff
@@ -83,10 +83,15 @@
   }
 
   function injectStyleSheets(metadata) {
-    var stylesheets =
-      "<link href=\"" + metadata.host + "/resources/remix/clean-slate.min.css\" rel=\"stylesheet\">\n" +
-      "<link href=\"" + metadata.host + "/resources/remix/style.css\" rel=\"stylesheet\">\n";
-    document.head.appendChild(stylesheets);
+    var stylesheet1 = document.createElement("link");
```
Pomax: if you use `var styleheet` here, you can simply reuse the `stylesheet` variable for the second sheet, too. Once you've added the first to the document head, `stylesheet` can be freely reassigned.

aayushsanghavi:
@flukeout @Pomax @humphd I had made a big blunder earlier. I hadn't defined the response and statusText of the xmlHttp request. I'm sorry about that; I have fixed that now.
I printed some console logs as @flukeout suggested yesterday and what I find is that the xmlHttp.status is 0, which is why it doesn't enter any of the conditionals and hence the request doesn't execute successfully. Any help with this?

public/resources/remix/index.js
```diff
@@ -6,122 +6,137 @@
...
-      data: {
-        host: metadata.host,
-        id: metadata.projectId,
-        title: metadata.projectTitle,
-        author: metadata.projectAuthor,
-        updated: metadata.dateUpdated
+  function injectDetailsBar(metadata, callback) {
+    var xmlhttp = new XMLHttpRequest();
+    xmlhttp.responseType = "text";
+    var url = metadata.host + "/projects/remix-bar";
```
humphd: @cadecairos, @gideonthomas  what's the right way for us to do XHR from publish to a remix page hosted on thimbleprojects cross domain?

public/resources/remix/index.js
```diff
@@ -6,122 +6,137 @@
...
+      title: metadata.projectTitle,
+      author: metadata.projectAuthor,
+      updated: metadata.dateUpdated
+    };
+
+    xmlhttp.onreadystatechange = function() {
+      if (xmlhttp.readyState == XMLHttpRequest.DONE ) {
+	  var response = xmlhttp.response;
+	  var textStatus = xmlhttp.statusText;
+        if (xmlhttp.status == 200) {
```
humphd: Watch your indents, this needs to be moved to the right.

public/resources/remix/index.js
```diff
@@ -6,122 +6,137 @@
...
+          document.body.insertBefore(response, document.body.firstChild);
+	    callback(null);
```
humphd: Same deal with these lines, you have 3 different indent levels.

public/resources/remix/index.js
```diff
@@ -6,122 +6,137 @@
...
+        }
+        else if (xmlhttp.status == 400) {
```
humphd: `} else if {`

aayushsanghavi:
Thanks @humphd. I'll be careful of the indentation next time.

public/resources/remix/index.js
```diff
@@ -6,122 +6,136 @@
...
-    '})(window,document,\'script\',\'//www.google-analytics.com/analytics.js\',\'ga\');\n' +
-    'ga(\'create\', \'' + gaTrackingId + '\', \'auto\');\n' +
-    'ga(\'send\', \'pageview\');\n' +
-    '</script>\n';
-
-    $(head).append(analyticsHtml);
+  function injectAnalytics() {
+    var analyticsHtml = document.createElement("script");
+    analyticsHtml.type = "text/javascript";
+    analyticsHtml.innerHTML = '(function(i,s,o,g,r,a,m){i[\'GoogleAnalyticsObject\']=r;i[r]=i[r]||function(){\n' + '(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),\n' + 'm=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)\n' + '})(window,document,\'script\',\'//www.google-analytics.com/analytics.js\',\'ga\');\n' + 'ga(\'create\', \'' + gaTrackingId + '\', \'auto\');\n' + 'ga(\'send\', \'pageview\');\n';
```
Pomax: I am not intimately familiar with the code here, but can't we keep this code in its own file and then require it in, replace the bits of string that need replacing, and then set the textContent?

On that note, even if we don't: this is not HTML code and so we should not be using `analyticsHtml.innerHTLM=`. For plain text content we should be using `analytics.textContent=` instead.

public/resources/remix/index.js
```diff
@@ -6,122 +6,136 @@
...
-    '(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),\n' +
-    'm=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)\n' +
-    '})(window,document,\'script\',\'//www.google-analytics.com/analytics.js\',\'ga\');\n' +
-    'ga(\'create\', \'' + gaTrackingId + '\', \'auto\');\n' +
-    'ga(\'send\', \'pageview\');\n' +
-    '</script>\n';
-
-    $(head).append(analyticsHtml);
+  function injectAnalytics() {
+    var analyticsHtml = document.createElement("script");
```
Pomax: technical nit: this is not going to be "HTML", this is a script, so naming variables after what they are is still an excellent idea. Let's rename this to `analytics` or `gaScript` or something.

public/resources/remix/index.js
```diff
@@ -6,122 +6,136 @@
...
+  function setupBar(err) {
     if(err) {
       console.log("[Thimble Error] Unable to inject Remix UI. Error was: `" + err + "`");
       return;
     }
 
     var isTouchDevice = 'ontouchstart' in document.documentElement;
-    var detailsBar = $(".details-bar");
-    detailsBar.attr("style", "");
+    var detailsBar = document.querySelector(".details-bar");
```
Pomax: Note that you're going from jQuery code to plain JS, and so some of the conveniences of jQuery get lost: a jQuery `$(...)` array may be empty, but will still allow you to call `.attr()` on it (it just doesn't do anything). In plain JS, however, `document.querySelector()` can yield `null`, and calling `null.setAttribute` will crash the JS script, so you need to make sure that you do something smart here in case `detailsBar` turns out to be null.

public/resources/remix/index.js
```diff
@@ -6,122 +6,136 @@
...
+    }
+    else {
+      detailsBar.classList.add("mouse-mode");
     }
 
-    detailsBar.on("click", ".thimble-button",function(){
-      detailsBar.removeClass("collapsed");
-      return false;
+    detailsBar.addEventListener("click", function(event) {
+      if (event.target.className == "thimble-button") {
```
Pomax: you have no guaranteed that this is the only class on the button, so use the classList API, as you were above:
```
if (event.target.classList.contains("thimble-button")) { ... }
```

public/resources/remix/index.js
```diff
@@ -6,122 +6,136 @@
...
+        detailsBar.classList.remove("collapsed");
+        return false;
```
Pomax: in HTML5 `return false` gets ignored. If you want to stop propagation you will need to call `event.stopPropagation();` and if you want to prevent browser-default behaviour you'll have to call `event.preventDefault();`.

public/resources/remix/index.js
```diff
@@ -6,122 +6,136 @@
...
+      }
     });
 
-    detailsBar.on("click", ".close-details-bar", function(){
-      detailsBar.addClass("collapsed");
-      return false;
+    detailsBar.addEventListener("click", function(event) {
+      if (event.target.className == "close-details-bar") {
```
Pomax: see the note above about using classList.contains

public/resources/remix/index.js
```diff
@@ -6,122 +6,136 @@
...
+        detailsBar.classList.add("collapsed");
+        return false;
```
Pomax: see the note above about returning false

public/resources/remix/index.js
```diff
@@ -6,122 +6,136 @@
...
+    var data = {
+      host: metadata.host,
+      id: metadata.projectId,
+      title: metadata.projectTitle,
+      author: metadata.projectAuthor,
+      updated: metadata.dateUpdated
+    };
+
+    xmlhttp.onreadystatechange = function() {
+      if (xmlhttp.readyState == XMLHttpRequest.DONE ) {
```
Pomax: you don't use `==` in JavaScript, unless you are *absolutely certain* you need a converting value. As such, always test with `===` unless you can explain and justify why JS should be allowed to try to automatically convert values for the purposes of boolean comparison in your code.

```
// this is true, because JS was told to convert things until both left and right are boolean:
"" == 0 

// this is also true, for the same reason
"cake" == 1

// this is false, because "" is string and 0 is number and these can't possibly be the same 
"" === 0 
```

public/resources/remix/index.js
```diff
@@ -6,122 +6,136 @@
...
+        var response = xmlhttp.response;
+        var textStatus = xmlhttp.statusText;
+        if (xmlhttp.status == 200) {
```
Pomax: `===`, not `==`

public/resources/remix/index.js
```diff
@@ -6,122 +6,136 @@
...
+          document.body.insertBefore(response, document.body.firstChild);
+	   callback(null);
+        } else if (xmlhttp.status == 400) {
```
Pomax: `===`, not `==`, also this should really be `>= 400` because any error code in the 4xx and 5xx range should trigger an error callback. They're *all* bad =D

public/resources/remix/index.js
```diff
@@ -6,122 +6,136 @@
...
+    stylesheet.href = metadata.host + "/resources/remix/clean-slate.min.css";
+    document.head.appendChild(stylesheet);
+
+    stylesheet = document.createElement("link");
+    stylesheet.rel = "stylesheet";
+    stylesheet.href = metadata.host + "/resources/remix/style.css";
+    document.head.appendChild(stylesheet);
+  }
+
+  function grep(items, callback) {
```
Pomax: JavaScript has [filtering already built into the Array prototype](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/filter):
```
function grep(items, preserveItem) {
  return items.filter( function(item, position) {
    return preserveItem(item);
  });
}
```
Other array functions worth knowing: `.map()`, `.some()`, `every()`, and `.reduce()`

public/resources/remix/index.js
```diff
@@ -6,122 +6,136 @@
...
 
-  function getMetadata($) {
+  function getMetadata() {
     var metadata = {};
-    var metaTags = $("meta");
+    var metaTags = document.getElementsByTagName("meta");
 
-    $.grep(metaTags, function(elem, index) {
-      return /^data-remix-.+/.test($(metaTags[index]).attr("name"));
+    grep(metaTags, function(metaTag) {
```
Pomax: if this is the only place we're using `grep` then we do not need the separate function, we can simply say
```
var hasDataRemixAttribute = function(element) {
  var name =element.getAttribute("name");
  if (!name) return false;
  return name.startsWith("data-remix-");
};

var metaElements = document.getElementsByTagName("meta");
var remixElements = Array.from(metaElements).filter( hasDataRemixAttribute );
```

Note that this also removes the regular expression, which is always safer than leaving a regular expression.

public/resources/remix/index.js
```diff
@@ -6,122 +6,136 @@
...
+      return /^data-remix-.+/.test((metaTag).getAttribute("name"));
     })
     .forEach(function(metaTag) {
-      metaTag = $(metaTag);
-      var key = metaTag.attr("name").replace(/^data-remix-/, "");
+      var key = metaTag.getAttribute("name").replace(/^data-remix-/, "");
```
Pomax: if we use the preceding code, we are already guaranteed that any element that makes it through the filter has a value for `name` that starts with `data-remix-` so we don't need a regular expression here anymore either, we can simply say `var key = metaTag.getAttribute("name").replace("data-remix-", "");`

That said, the argument to this function should probably be metaElement, since we're passing around elements, not tags (tags are strings)

public/resources/remix/index.js
```diff
@@ -6,122 +6,136 @@
...
   }
 
-  (function(doc, script) {
-    script = doc.createElement("script");
-    script.type = "text/javascript";
-    script.onload = run;
-    script.src = "https://ajax.googleapis.com/ajax/libs/jquery/2.1.4/jquery.min.js";
-    doc.getElementsByTagName("head")[0].appendChild(script);
-  }(document));
+  run();
```
Pomax: if this function doesn't need to be called anywhere else, and it is immediately invoked, then we don't need this function. We just need the code inside of it to immediately run:

```
function dothis() {
  console.log("test");
}
dothis();
```
executes identically to:
```
console.log("test");
```
but without the function allocation and call overhead.

public/resources/remix/index.js
aayushsanghavi: @Pomax I'll change it to analytics.textContent for now.

aayushsanghavi:
@Pomax thanks for the awesome review!

aayushsanghavi:
@Pomax I apologise for the delayed response. I made the changes as you suggested but we're still stuck there. Still no remix button on the published page.
thimble-remix

Have a look at the Developer tools console.

public/resources/remix/index.js
```diff
@@ -21,36 +20,37 @@
 
     var isTouchDevice = 'ontouchstart' in document.documentElement;
     var detailsBar = document.querySelector(".details-bar");
-    detailsBar.setAttribute("style", "");
+    
+    if (detailsBar) {
+      detailsBar.setAttribute("style", "");
+      if (isTouchDevice) {
```
Pomax: we might need to consider what we do here a bit more:  since we add additional functionality based on `mouse-mode`, what we *want* to know is "is there a mouse cursor or not", so we should figure out how to test for that. Just because something is a touch device does not mean there won't also be a mouse cursor available.

(My tablet, for instance, has a dock with a mouse pad on it. Despite being a touch device, mouseenter and mouseleave will work just fine)

thoughts @flukeout, @humphd?

aayushsanghavi:
@Pomax besides your suggestion about the mouse-mode functionality, do you suggest any other changes in the patch?
Also, what do you think about the console error, "Cross Origin Request Blocked".

Pomax:
that's all the suggests I have right now - as the CORS error I'm not too sure because I don't know what you're doing when you're seeing it. Brackets runs on port 8000 and Thimble runs on port 3500, whereas the CORS error is about something on port 8001: I don't know what that would be. Were you running something on a different port?

aayushsanghavi:
I'm not very sure of that. I might have run some other project before Thimble but I stop them before running Thimble and they're on other ports like 3000 or 8008 so I guess there was no interaction with port 8001.

flukeout:
Hey @aayushsanghavi - just checking in - is this PR running on your machine now successfully and if so, should I test it now? Let me know, thanks.

public/resources/remix/index.js
flukeout: What's the experience for you on that device using the current implementation of the remix button?

I think all that happens is that for touch-capable devices, the remix button moves to a different spot and doesn't rely on hovering over it to work, so it's still functional.

aayushsanghavi:
@flukeout not really. I still can't get the remix button to work. I'm guessing that it has something to do with the CORS issue. But it would be great if you could try it and give your review on it. Thanks!

Pomax:
@aayushsanghavi if you weren't running anything on 8001, then what is this part of your screenshot?

screenshot 180

This shows the devtools for *something* running on port 8001, which should not be involved in any thimble/brackets testing.

Pomax:
Okay, port 8001 is confirmed noxmox (the local hosting of what in deployment is S3 content). That said if I publish something locally with brackets running in a terminal and thimble etc running through vagrant, I do not get any CORS errors when using thimble and brackets `master` branches.

public/resources/remix/index.js
```diff
@@ -6,122 +6,126 @@
...
-    .done(function(response) {
-      $(response).prependTo("body");
-      callback(null, $);
-    })
-    .fail(function(jqXHR, textStatus) {
-      callback(textStatus);
-    });
+    };
+
+    xmlhttp.open("POST", url, true);
```
Pomax: This is absolutely not a `POST` operation: we are *getting* a file from the server, we are not pushing data into the server for storage.

aayushsanghavi:
@Pomax thanks for digging into it. Honestly, I don't have much idea about xhr headers so I looked up the documentation on MDN but I still don't know what exactly are we missing.
Are you referring to these - https://github.com/mozilla/thimble.mozilla.org/blob/master/public/resources/remix/index.js#L61-L63 ?
Please help me out with this.

public/resources/remix/index.js
```diff
@@ -6,122 +6,126 @@
...
+      author: metadata.projectAuthor,
+      updated: metadata.dateUpdated
+    };
+
+    xmlhttp.onreadystatechange = function() {
+      if (xmlhttp.readyState === XMLHttpRequest.DONE ) {
+        var response = xmlhttp.response;
+        var textStatus = xmlhttp.statusText;
+        if (xmlhttp.status === 200) {
+          document.body.insertBefore(response, document.body.firstChild);
```
Pomax: that is not what `$(response).prependTo("body")` does - the prependTo call is actually hack that prepends the `entire body` to the element given. What you actually need to do here is
```
var fragmentBody = document.createElement("body");
var docFragment = document.createDocumentFragment();
docFragment.appendChild(fragmentBody);
fragmentBody.innerHTML = response;
var remixBar = fragmentBody.querySelector("div");
document.body.appendChild(remixBar);
callback(null);
```

public/resources/remix/index.js
```diff
@@ -6,122 +6,126 @@
...
   }
 
-  function setupBar(err, $) {
+  function setupBar(err) {
     if(err) {
       console.log("[Thimble Error] Unable to inject Remix UI. Error was: `" + err + "`");
       return;
     }
 
     var isTouchDevice = 'ontouchstart' in document.documentElement;
```
Pomax: this returns `true` on my desktop. We cannot use this code to test whether this is actually a touch device or not. However, more important: this has nothing to do with removing jQuery, and so should be its own PR.

Please take this out as part of your PR update, and file a new issue (like "add in touch vs mouse input detection to the remix bar") and then form a new PR specifically for that issue.

Pomax:
Alright, after trying to debug this, the CORS problem is due to using the wrong HTTP verb (you want to `GET` the remix bar script, not `POST` data to the server), but a bigger issue is that your PR is trying to solve two problems at once: 

1. the actual issue of removing jQuery
2. add in touch-input awareness.

You PR should only do the first, not also the second, so please remove all the touch rewrites so that *first* we have a clean, no-jQuery remix script, and then we can work on adding in touch device detection and event binding separately.

flukeout:
Thanks for the review @Pomax - agreed on tackling one issue at a time.

humphd:
@aayushsanghavi how are you doing with this?  Did you get stuck on something we can help with, or just busy?  You're pretty close to having this work completed, it would be great to get in.

aayushsanghavi:
@humphd and @Pomax sorry for the late response. I had college exams till yesterday and project evaluations today. I'm on it. Today I'll make the changes and in case I get stuck somewhere I'll let you know.

public/resources/remix/index.js
aayushsanghavi: So basically I'd have to remove the setupBar function and all instance of making a call to it as a callback function in the injectDetailsBar function. Am I on the correct lines?

aayushsanghavi: Also, in the injectDetailsBar, we do -
`callback(textStatus)` when the return status of the xmlhttp request is bad. If I remove that line (because we're going to remove the callback function), should we have a console.log saying there was some error?

public/resources/remix/index.js
aayushsanghavi: My bad! I had developed the notion that sending data in an ajax call means it is a POST request.

aayushsanghavi:
@Pomax @humphd I made the changes suggested and now the Remix button finally shows up!

remix

@Pomax please have a look. Also, I apologise for delaying the work.

flukeout:
Testing out the latest changes, thanks @aayushsanghavi

flukeout:
Okay, cool - yeah looks like the remix bar expands now, which is great. However all of the information from it missing (Title, author, etc).

image

Is that something you're still working on @aayushsanghavi? Thanks!

aayushsanghavi:
@flukeout I'm not working on it separately. I thought that the data that we send after the GET request did the job of sending information related to the project and author.
I'll try looking into it. @Pomax @humphd could you'll also please look into this?

public/resources/remix/index.js
```diff
@@ -6,122 +6,126 @@
...
+    document.head.appendChild(stylesheet);
   }
 
-  function getMetadata($) {
+  var hasDataRemixAttribute = function(element) {
+    var name = element.getAttribute("name");
+    if (!name) {
+      return false;
+    }
+    return name.startsWith("data-remix-");
```
gideonthomas: You can't use `startsWith` because it isn't supported by IE11.

Instead, do:
```js
return /^data-remix-.+/.test(name);
```

gideonthomas: This way you don't have to do the null check you do above either because `.test(null)` and `.test(undefined)` automatically returns false.

public/resources/remix/index.js
```diff
@@ -6,122 +6,126 @@
...
+    var analytics = document.createElement("script");
+    analytics.textContent = '(function(i,s,o,g,r,a,m){i[\'GoogleAnalyticsObject\']=r;i[r]=i[r]||function(){\n' + '(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),\n' + 'm=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)\n' + '})(window,document,\'script\',\'//www.google-analytics.com/analytics.js\',\'ga\');\n' + 'ga(\'create\', \'' + gaTrackingId + '\', \'auto\');\n' + 'ga(\'send\', \'pageview\');\n';
+    head.appendChild(analytics);
   }
 
-  function setupBar(err, $) {
-    if(err) {
+  function setupBar(err) {
+    if (err) {
       console.log("[Thimble Error] Unable to inject Remix UI. Error was: `" + err + "`");
```
gideonthomas: change this to: `console.log("[Thimble Error] Unable to inject Remix UI. Error was: ", err);`

public/resources/remix/index.js
```diff
@@ -6,122 +6,126 @@
...
       return;
     }
 
-    var isTouchDevice = 'ontouchstart' in document.documentElement;
```
gideonthomas: why do you remove this functionality?

public/resources/remix/index.js
```diff
@@ -6,122 +6,126 @@
...
-    var detailsBar = $(".details-bar");
-    detailsBar.attr("style", "");
-
-    if(isTouchDevice) {
-      detailsBar.addClass("touch-mode");
-    } else {
-      detailsBar.addClass("mouse-mode");
+    var detailsBar = document.querySelector(".details-bar");
+    
+    if (detailsBar) {
```
gideonthomas: change this to:
```js
if(!detailsBar) {
  return;
}

...
```

public/resources/remix/index.js
```diff
@@ -6,122 +6,126 @@
...
+      detailsBar.setAttribute("style", "");
+      detailsBar.classList.add("mouse-mode");
+
+      detailsBar.addEventListener("click", function(event) {
+        if (event.target.classList.contains("thimble-button")) {
+          detailsBar.classList.remove("collapsed");
+          event.stopPropagation();
```
gideonthomas: just `return false` instead

public/resources/remix/index.js
aayushsanghavi: @Pomax earlier suggested that the touchDevice functionality has issues so we should handle that separately because removing jQuery from this file is independent of that. He suggested that I should fix the jQuery part in this PR and open a separate PR for the touchDevice.

public/resources/remix/index.js
aayushsanghavi: I was returning false previously but in an earlier review [Pomax suggested](https://github.com/mozilla/thimble.mozilla.org/pull/1825#discussion_r106050023) that HTML5 ignores `return false`. So I changed it to `event.stopPropagation()`

public/resources/remix/index.js
```diff
@@ -6,122 +6,126 @@
...
-      $(response).prependTo("body");
-      callback(null, $);
-    })
-    .fail(function(jqXHR, textStatus) {
-      callback(textStatus);
-    });
+    };
+
+    xmlhttp.open("GET", url, true);
+    xmlhttp.send(data);
```
aayushsanghavi: Makes sense. The GET request does essentially send the data in the url.

aayushsanghavi:
Thanks a lot @gideonthomas for the review. I'll get back with the changes soon.

public/resources/remix/index.js
```diff
@@ -6,122 +6,126 @@
...
-      accepts: "text/html",
-      dataType: "html",
-      data: {
-        host: metadata.host,
-        id: metadata.projectId,
-        title: metadata.projectTitle,
-        author: metadata.projectAuthor,
-        updated: metadata.dateUpdated
+  function injectDetailsBar(metadata, callback) {
+    var xmlhttp = new XMLHttpRequest();
```
aayushsanghavi: @gideonthomas I tried making the request on metadata.host but that didn't work. I got an error saying request was blocked, so I'm sticking with `url = metadata.host + "/projects/remix-bar";`

public/resources/remix/index.js
aayushsanghavi: Actually I need to keep the line `request.send(data)` too. Simply `url + "?" + querystring` isn't working.

aayushsanghavi:
@gideonthomas I've made the changes (as you suggested). I left out the `return false` vs `event.stopPropagation()`. Please have a look.
Also, @flukeout now we're getting the project metadata on hovering on the remix button (thanks to @gideonthomas) -

screenshot from 2017-04-27 14-13-12

public/resources/remix/index.js
gideonthomas: Alright, @aayushsanghavi can you file a separate issue for that if it isn't filed already?

public/resources/remix/index.js
```diff
@@ -6,122 +6,121 @@
...
-      callback(textStatus);
-    });
+    };
+
+    var querystring = Object.keys(data).map(function(key) {
+      return encodeURIComponent(key) + "=" + encodeURIComponent(data[key]);
+    }).join("&");
+    url = url +  "?" +querystring;
+    request.open("GET", url);
+    request.send(data);
```
gideonthomas: @aayushsanghavi what browser are you using? This is still incorrect. You shouldn't be sending data to send if it isn't in a valid form (in fact for GET requests you shouldn't be sending request bodies at all).

Take a look at https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest which notes this. If you think that is incorrect, you might have another bug in here. Otherwise, you should link to documentation that says that you need to send an object to `send` with GET requests.

public/resources/remix/index.js
```diff
@@ -6,122 +6,121 @@
...
-    })
-    .fail(function(jqXHR, textStatus) {
-      callback(textStatus);
-    });
+    };
+
+    var querystring = Object.keys(data).map(function(key) {
+      return encodeURIComponent(key) + "=" + encodeURIComponent(data[key]);
+    }).join("&");
+    url = url +  "?" +querystring;
```
gideonthomas: space after `+`

public/resources/remix/index.js
```diff
@@ -6,122 +6,121 @@
...
-    var metaTags = $("meta");
-
-    $.grep(metaTags, function(elem, index) {
-      return /^data-remix-.+/.test($(metaTags[index]).attr("name"));
-    })
-    .forEach(function(metaTag) {
-      metaTag = $(metaTag);
-      var key = metaTag.attr("name").replace(/^data-remix-/, "");
+    var metaElements = document.getElementsByTagName("meta");
+    var remixElements = Array.from(metaElements).filter(hasDataRemixAttribute);
```
gideonthomas: Change this to `Array.prototype.filter.call(metaElements, hasDataRemixAttribute);

aayushsanghavi:
@gideonthomas you were right. It does and has to work without explicitly sending data. Its working now. I guess, earlier when I tried I didn't do `request.send()`. I also made the other changes as you suggested.

gideonthomas:
Nice work @aayushsanghavi! Thanks for the PR! Filed https://github.com/mozilla/thimble.mozilla.org/issues/2041 as a follow-up for the touch events.

aayushsanghavi:
Thanks for merging this @gideonthomas and sorry I didn't file the touch events issue when you told earlier.
I'll get onto it too.

PR #1828, opened 2017-03-07
Updated node deps for package.json

omytryniuk:
fixes #1743

I have updated the dependencies. 
I have tested "npm install":

- without virtualization
- running Thimble on VM as well. No errors. 

Thank you for your advice to use npm-check, it was a very useful tool in terms of checking updated versions of dependencies. 

I have not had any problems, except grunt, and that's why I decided to stick with same version of the dependency we had before (however most of grunt peerdependencies were updated).

Although I did not see any errors, I would really like you, if you have a chance, to run it as well, to be sure that there are no errors. 

Thank you!
screenshot 2017-03-07 17 55 13

PR #1847, opened 2017-03-08
Fix #1829: Added warning supression for new versions of jshint

### LINKED ISSUE #1829: npm test for new deps failed because of "A constructor name should start with an uppercase letter"
I was trying to update node deps for the project and CI Travis shows me an [error](https://travis-ci.org/mozilla/thimble.mozilla.org/builds/208818821)

screenshot 2017-03-07 17 55 13

According to it, the next [function](https://github.com/mozilla/thimble.mozilla.org/blob/master/server/lib/environment.js)
```js
var habitat = require('habitat');
habitat.load(require('path').resolve(__dirname, '../../.env'));
var env = new habitat();
``` 
is wrong, because jshint expects that an identifier following the new operator(when the newcap option is set to true)  should start with an upperCase.

omytryniuk:
Here is the fix of the issue that happens with [deps update](https://github.com/mozilla/thimble.mozilla.org/pull/1828). 
This request updates deps, and the newest version of jshint, treats this code as an error, instead of just as a warning. 
```js
var habitat = require('habitat');
habitat.load(require('path').resolve(__dirname, '../../.env'));
var env = new habitat();
```
There is anything wrong with this code. That's why the easist solution to pass new updates of jhint, is to supress the warning. In addition, this syntax check will be deprecated in the next version of jshint.

### DIFF
--- a/server/lib/environment.js
+++ b/server/lib/environment.js
@@ -1,3 +1,5 @@
+/*jshint newcap:false */
+
 var habitat = require('habitat');
 habitat.load(require('path').resolve(__dirname, '../../.env'));
 var env = new habitat();
--- a/server/lib/environment.js
+++ b/server/lib/environment.js
@@ -1,5 +1,4 @@
-/*jshint newcap:false */
-
+/* jshint newcap:false */
 var habitat = require('habitat');
 habitat.load(require('path').resolve(__dirname, '../../.env'));
 var env = new habitat();


server/lib/environment.js
```diff
@@ -1,3 +1,5 @@
+/*jshint newcap:false */
+
```
humphd: Can you remove this blank line, since it relates directly to line 3.

humphd:
One small nit to fix, and this is good to go.

server/lib/environment.js
omytryniuk: @humphd Sure! Thank you. I just saw in  the previous examples with jshint directives a blank line was left and that is why I left it; however I will fix it. 
Just a small question. 
Here:
```js
/*jshint newcap:false */
```
Should I keep the extra space between false and * (same as it is used by jshint developers), or just get rid of the space? Thank you

humphd: If it were me, I'd but a space before/after both `*` but it's all optional.

humphd:
Thanks for researching and fixing this.  You should rebase #1828 on top of master now that this has landed.

omytryniuk:
You are welcome, @humphd. I will do rebase. Thank you for your help!

PR #1856, opened 2017-03-09
Fixes the issue of popup window overflowing to the left side

### LINKED ISSUE #1824: "Add a file to your project" popup window overflows to the left side
Hey

I came across this bug while working on thimble recently.
If we resize the left pane to make its width small enough on the editor page, the "add a file to your project popup window" popup window overflows to the left.

screenshot from 2017-03-07 11-26-22

I would like to work on fixing this.

hkirat:
Fixes #1824. 
Also replaces the labels of menu elements.

### DIFF
--- a/public/editor/scripts/editor/js/fc/bramble-popupmenu.js
+++ b/public/editor/scripts/editor/js/fc/bramble-popupmenu.js
@@ -15,13 +15,13 @@ define(function(require) {
         // Determine where to horizontally place menu based on button's icon location
         var menuWidth = self._menu$.width();
         var leftOffset = self._button$.offset().left - menuWidth/2 + 11;
-        var iconWidth = $("#filetree-pane-nav-add").width();
+        var iconWidth = self._button$.width();
         var arrowOffset = menuWidth/2 - iconWidth/2;
         if(leftOffset < 0) {
           arrowOffset = menuWidth/2 - iconWidth/2 + leftOffset;
           leftOffset = 0;
         }
-        $("#filetree-pane-nav-arrow-tip").css("left", arrowOffset);
+        self._menu$.find(".arrow-tip").css("left", arrowOffset);
         self._menu$.css("left", leftOffset);
       };
     }
--- a/public/editor/scripts/editor/js/fc/bramble-popupmenu.js
+++ b/public/editor/scripts/editor/js/fc/bramble-popupmenu.js
@@ -15,9 +15,10 @@ define(function(require) {
         // Determine where to horizontally place menu based on button's icon location
         var menuWidth = self._menu$.width();
         var leftOffset = self._button$.offset().left - menuWidth/2 + 11;
-        var arrowOffset = 86;
+        var iconWidth = $("#filetree-pane-nav-add").width();
+        var arrowOffset = menuWidth/2 - iconWidth/2;
         if(leftOffset < 0) {
-          arrowOffset = 86 + leftOffset;
+          arrowOffset = menuWidth/2 - iconWidth/2 + leftOffset;
           leftOffset = 0;
         }
         $("#filetree-pane-nav-arrow-tip").css("left", arrowOffset);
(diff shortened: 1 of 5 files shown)


flukeout:
Okay, this works well for the New File dialog. I've noticed that there are two different menus built using the 'createWithOffset` method..

```    var menu = PopupMenu.createWithOffset("#filetree-pane-nav-add", "#filetree-pane-nav-add-menu");```

The second one is the editor options menu (with the gear). The code you wrote only moves the arrow for the "File" menu - can you make it generic so that it moves the arrows within both menus?

**Comment by hkirat** [comment]
Thanks for the review @flukeout. 
Made the width general according to the language.

**Comment by flukeout** [comment]
Looking now!

File: public/editor/scripts/editor/js/fc/bramble-popupmenu.js [hunk 1]
```diff
@@ -15,6 +15,13 @@ define(function(require) {
         // Determine where to horizontally place menu based on button's icon location
         var menuWidth = self._menu$.width();
         var leftOffset = self._button$.offset().left - menuWidth/2 + 11;
+        var iconWidth = $("#filetree-pane-nav-add").width();
```
flukeout: Because this code is used for both of the file menu and the editor gear menu, we can't just reference the ``filetree-pane-nav-add`` element.

Instead, you can use   ``var iconWidth = self._button$.width();`` - the $button element is the icon that is clicked to open both menus.

public/editor/scripts/editor/js/fc/bramble-popupmenu.js
```diff
@@ -15,6 +15,13 @@ define(function(require) {
...
+        var arrowOffset = menuWidth/2 - iconWidth/2;
+        if(leftOffset < 0) {
+          arrowOffset = menuWidth/2 - iconWidth/2 + leftOffset;
+          leftOffset = 0;
+        }
+        $("#filetree-pane-nav-arrow-tip").css("left", arrowOffset);
```
flukeout: Here, again - we should not reference the filetree element, but instead find an ``.arrow`` element in the ``$menu`` element. I bet something like ``self._menu$.find('.arrow')`` will work. 

However, since the gear menu doesn't yet have an arrow element as a ``<div>`` you'll also have to add one there.

flukeout:
Hi @hkirat - I've left some comments about how to make this code work for both of the menus that use it. Let me know if you have questions.

flukeout:
I'll test later today.

hkirat:
@flukeout Could you review?

humphd:
@hkirat I see you're asking for review on various bugs.  We did a triage yesterday, and assigned reviewers already--you'll see that @flukeout is already assigned.  We've got about 30 reviews on the way right now, and another 50 or so bugs in progress.  We'll get to all of it for sure, but give us time :)

hkirat:
Thanks for the reply @humphd
Apologies for my impulsiveness. Will keep that in mind.

flukeout:
Testing now!

flukeout:
I'm sorry I let this one slide @hkirat - I'm testing it now again and will update the issue ASAP, thanks for your patience.

public/editor/stylesheets/editor.less
```diff
@@ -393,7 +381,8 @@ body {
   }
 }
 
-#editor-pane-nav-options-menu:after {
+
+.arrow-tip {
   content: '';
```
flukeout: Let's remove the "content" rule, we only needed that for the ``:after`` pseudo-element

flukeout:
@hkirat - left a small note about the ``content`` rule in the CSS which we don't need anymore. Also, let's revert all of the changes in ``locales/en-GB/messages.properties`` - even thought they are in English, we should only modify the ``en-us`` strings.

Thanks!! 💯 Once those changes are made we can land this change!

hkirat:
Thanks for the review @flukeout 
I've made the required changes.

flukeout:
Thanks a lot @hkirat - we did it! 👍 👍

PR #1863, opened 2017-03-10
Contributing guidelines based on current WIKI page

flukeout:
Once this lands, we can remove the corresponding Wiki Page


Pomax:
make it happen.

Issue #1862, opened 2017-03-10
Get rid of two javascript errors we always see


flukeout:
When you load the editor, these two fellas here are always popping up...

![image](https://cloud.githubusercontent.com/assets/25212/23813343/553e9852-0593-11e7-86c0-5004128ccad9.png)

...and misleading us into thinking they're significant or related to problems we might be experiencing at the time.

humphd:
paging @gideonthomas... these are because the strings aren't in the `src/` build but are going to be there when we do full localization and serve from `dist/`

humphd:
The code in question is here: https://github.com/mozilla/brackets/blob/master/src/extensions/default/bramble-move-file/main.js#L16-L18

We are using strings from the Brackets `Strings` module, and for some reason it seems to think that they aren't there.  @gideonthomas I wonder if we should change our localization merge code to run on `checkout` vs. as a build step, so your `src/` dir is always right?

gideonthomas:
@humphd I'm not sure I follow. I think the problem here is that we run `npm run unlocalize` which erases the changes we make to `src/nls/root/strings.js` which includes the string for "Move File". I wonder if there is a way to tell it to use strings from `dist` instead.

dangkhue27:
I think I will fix this one before fixing #1227, but i'm not sure what needs to be done here. Can you tell me how can I start with this issue?

dangkhue27:
@humphd Can you give me some details on this issue? I can dig further into the problem.

humphd:
cc'ing @Shank09, @MargarytaChepiga

@gideonthomas can you please shed some light on this issue and what we should do to solve it?

For the benefit of those that I'm including in this bug, here is what I know:

* Adobe uses its own strings in `src/nls/*` and any parts of the Brackets code that use the `Strings` module, are basically using this.
* Thimble (and our fork of Brackets) has some strings that Adobe doesn't need.  Our localizers translate strings for Thimble, and we have scripts that download and merge strings with what Adobe has.  The code is https://github.com/mozilla/brackets/blob/master/scripts/properties2js.js and we call it in the Gruntfile here https://github.com/mozilla/brackets/blob/master/Gruntfile.js#L478 (it's defined here https://github.com/mozilla/brackets/blob/master/Gruntfile.js#L424
[...]

PR #1891, opened 2017-03-15
Issue1745 - add allow edit svg file as xml toggle UI

peiying16:
Need bracket side PR https://github.com/mozilla/brackets/pull/650

The toggle to switch svg or xml is under setting.
1. Turn the toggle to xml
2. Reselect the svg file on the left side
3. You can see the xml code

There is a bug. To edit as xml, you must reload the page after you add the svg file.

Should we add an alert to tell the user to reload the page when the toggle is turned to xml?

https://github.com/mozilla/brackets/blob/master/src/editor/ImageViewer.js#L378 This function determines if the image viewer is a suitable view to open a file. Line 378 reads the preference to determine  it.


views/editor/nav-options.html
```diff
@@ -77,6 +77,16 @@
                   <img id="sun-white" class="hide" src="/img/icon/theme-sun-white.svg" width="20px" class="hide" height="20px">
                   <img id="sun-green" src="/img/icon/theme-sun-green.svg" width="20px" height="20px">
                 </div>
+              </li>
+              <li>
+                <span>SVG/XML</span>
```
Pomax: This should be a string in the locale/en-US messages.properties file, fetched with the same `{{ gettext(......) }}` calls you see above/below these lines

views/editor/nav-options.html
```diff
@@ -77,6 +77,16 @@
...
+                <div title="XML/SVG" class="switch-toggle switch-enabled" id="edit-SVG-toggle">
```
Pomax: Please follow the same HTML structure as the other menu items. You are not using the same div nesting here, and your new code closes off the `<ul>` list, which it should not be doing. 

Also, as above, please make the title a `gettext` string.

peiying16: Hi, there is something wrong with my git bash. I will figure it out.

peiying16:
I think I fix all the requested changes. Please review it.

views/editor/nav-options.html
```diff
@@ -93,6 +93,13 @@
                 <div class="toggle-button"></div>
               </div>
             </li>
+            <li>
+              <span>{{ gettext("XML/SVG") }}</span>
```
Pomax: where did these strings go? I don't see an update to the locale/en-US/messages.properties file in the PR now, which means `gettext()` won't do anything here.

peiying16: Do you mean I should edit all locale/*/messages.properties files manually?

peiying16: If there is no function generate this codes, I can add zh-CN and zh-TW. : )

Pomax: No, *specifically* the `en-US` one only. That's the master file. You don't edit any others, that is handled through by the localizers.

peiying16:
@Pomax I add the strings, please review it.

locales/en-US/messages.properties
```diff
@@ -155,6 +156,7 @@ navDarkThemeIconTitle=Switch to the dark theme
 navLightThemeIconTitle=Switch to the light theme
 navWrapTextTitle=Toggle wrapping long lines of code
 navAllowJSTitle=Toggle running JavaScript code in the preview
+navXMLSVGTitle=Toggle openning svg file with code editor or image viewer
```
Pomax: Given that this toggle is an on/off, we can say what this does when turned on: `View SVG files as code instead of as image`

Pomax:
apologies for the delay in review =(

peiying16:
@Pomax Fixed.

Pomax:
I noticed you closed https://github.com/mozilla/brackets/pull/650, so I'm not quite sure what you'd like to do here. Is that PR still necessary? If so, why was it closed? If it's not, is there anything else I need to do to brackets to make sure this PR works?

peiying16:
@Pomax I tested the toggle UI many weeks before, it could switch to off. I'm not sure why it stuck when you were testing. I will test it again.

The reason why I closed brackets is that there is still a bug that I couldn't fix, and I gave up.

I think this thimble PR is still useful because this PR provides the UI for this SVG feature. If anyone wants to add this feature, they can pull this PR.

Pomax:
if there is no brackets code to enable the functionality, just having the UI is not enough for us to land it - We cannot land any changes to the UI if there is no brackets functions that UI can tie in to. I'll have to discuss with @gideonthomas, @flukeout, and @humphd what should happen with this issue and the closed PR code: you were very close to a finished working PR, so perhaps we can find someone else to finish it up, and then we can revisit this patch again to tie the UI into it. This PR will probably be sitting for a few days until someone can pick up the brackets PR.

peiying16:
@Pomax OK, I will test this UI again to make sure it runs perfectly.

peiying16:
@Pomax I fixed the UI bug, you can switch it now. Please review it.

Pomax:
@peiying16 I'm still unable to use this in combination with our braackets on its master branch. The console gives me the error:
```
Attempting to register a command with a missing name, id, or command function:undefined bramble-move-file.moveFile
register @ CommandManager.js:195
Menus.js:582 addMenuItem(): commandID not found: bramble-move-file.moveFile
Menu.addMenuItem @ Menus.js:582
```
I will consult with @humphd on what the best way forward for this PR is

peiying16:
@Pomax The error is from this issue
https://github.com/mozilla/thimble.mozilla.org/issues/1862#issuecomment-289541226

peiying16:
@Pomax 
Hi, it works on both my IE and Chrome.

peiying16:
I discover that the bug not only happens on SVG file. It happens on all nontext files.

humphd:
Closing in favour of https://github.com/mozilla/thimble.mozilla.org/pull/1970.

PR #1892, opened 2017-03-15
Makes sure the adapt method gets called after resizing

flukeout:
The old implementation waits 200ms after adapting the layout before it calls the adapt method again. A lot can happen in that 200ms, so we want to also make sure we call it once after the last UI resize.

This feels a bit janky @Pomax so I'm curious about your thoughts.

The goal is to call the adapt method frequently, but not more often than 200ms, while also calling it one last time at the end.

To test this, resize the windows quickly - the icons should always adapt.

### DIFF
--- a/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
+++ b/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
@@ -13,16 +13,22 @@ define(function(require) {
 
   var adapting = false;
   var adaptTimeoutMS = 200; // How often we adapt editor bar layout
+  var adaptTimeout;
 
   function updateLayout(data) {
     $(".filetree-pane-nav").width(data.sidebarWidth);
     $(".editor-pane-nav").width(data.firstPaneWidth);
     $(".preview-pane-nav").width(data.secondPaneWidth);
 
-    // Only adapt the layout every once in a while
+    window.clearTimeout(adaptTimeout);
+    adaptTimeout = setTimeout(function(){
+      adaptLayout();
+    },adaptTimeoutMS);
+
     if(!adapting) {
       adapting = true;
       adaptLayout();
+      window.clearTimeout(adaptTimeout);
       setTimeout(function(){
         adapting = false;
       },adaptTimeoutMS);


Pomax:
This looks pretty decent, another option that we might want to have a look at is something we've done a few times in other code as well:


/**
 * wrap the code in a closure so the lock-var doesn't pollute
 */
function slowlyListenForEvent(el, evtName, runThisDuringUpdate, minProcessingInterval) {
  if (!el || !evtName || !runThisDuringUpdate) return;

  var minProcessingInterval = minProcessingInterval || 200;

  // lock-var that tracks whether or not there is an event waiting.
  var processing = false;

  // lock-var that tracks whether or not an event-handling chain call is in progress.
  var chainRunning = false;

  // This gets called by the event listener, and acts as controller 
  // that determines whether or not to start an update chain call.
  function queueEvent(evt) {
    // track the latest event. We don't really care about the now-out-of-date events
    processing = evt;

    // And only start a chain call if we're not already running one.
    if (!chainRunning) {
      chainRunning = true;
      updateScroll();
    }
  }

  // This function does the actual updating and keeps scheduling calls
  // to itself until "processing" stays false across a timeout call, which
  // means no new events came in in the mean time.
  var update = function (evt) {
    // no need to do any processing if we're already processing
    if (processing) {
      return (chainRunning = false);
    }

    // before anything, grab the latest-event-data
    var latestEvent = processing;

    // And then assume we'll be done (which might be a false assumption
    // based on what happens between now and the next scheduled call)
    processing = false;

    // Then we do whatever it is we need to do. While we do that,
    // a new event might cause "processing" to be set back to "true".
    runThisDuringUpdate(latestEvent);

    // Then schedule a call to update again, at some point in the
    // future: if the event triggering caused "processing" to become
    // true between now and then, we'll rerun. If it stays false,
    // no new events came in and update() will terminate the chain call.
    setTimeout(update, minProcessingInterval); 
  };

  // And finally after all that set up, we bind the listener to whatever element
  el.addEventListener("scroll", updateScroll);
}
```

And then that can be used anywhere as:

```
slowlyListenForEvent(someElement, "scroll", (evt) => {...}, 200);
```

humphd:
@flukeout this got stalled, want to finish it?

flukeout:
Yeah, it's on my list - prioritizing smaller PRs and contributor stuff at the moment.

humphd:
OK, smart.  Thanks for the update.  I was just crawling through PRs to see what needed help.

PR #1897, opened 2017-03-17
Site-wide font changes & updated styling for Publish dialog

flukeout:
Lots of changes here, I'll highlight a few

* Removed Fira entirely and moved everything to Open Sans
  * Less stuff to load, yey!
* Bumped the default font-weight for Open Sans to 400
  * 300 was too faint, this should help overall legibility
* Updated styling for the Publish dialog
  * Mostly just tweaked hierarchy and link styling
  * Moved "Update Published" section above the link
* Removed duplicate stylesheet link from Editor

### DIFF
--- a/public/editor/stylesheets/editor.less
+++ b/public/editor/stylesheets/editor.less
@@ -1,6 +1,7 @@
 body {
   font-family: 'Open Sans', sans-serif;
   font-size: 1em;
+  font-weight: 400;
   overflow: hidden;
 }
 
@@ -14,14 +15,14 @@ body {
 @transitionTime: 0.15s;
 
 #webmaker-bramble {
-    position: absolute;
-    display: block;
-    top: @brambleTop;
-    left: 0;
-    width: 100%;
-    bottom: 0;
-    border: 0;
-    background-color: #f8f8f8;
+  position: absolute;
+  display: block;
+  top: @brambleTop;
+  left: 0;
+  width: 100%;
+  bottom: 0;
+  border: 0;
+  background-color: #f8f8f8;
 }
 
 .bramble-toolbar {
@@ -73,7 +74,7 @@ body {
 
 .filetree-pane-nav-title {
   color: white;
-  font-weight: bold;
+  font-weight: 600;
   font-size: 1em;
   opacity: 0.45;
   filter: alpha(opacity=45);
@@ -208,7 +209,7 @@ body {
 
   .editor-pane-nav-title {
     color: white;
-    font-weight: bold;
+    font-weight: 600;
     opacity: 0.45;
     filter: alpha(opacity=45);
     margin-right: 10px;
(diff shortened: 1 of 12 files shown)


humphd:
@gideonthomas or @Pomax, would one of you be willing to review this?

public/editor/stylesheets/editor.less
```diff
@@ -14,14 +15,14 @@ body {
...
-    left: 0;
-    width: 100%;
-    bottom: 0;
-    border: 0;
-    background-color: #f8f8f8;
+  position: absolute;
+  display: block;
+  top: @brambleTop;
+  left: 0;
+  width: 100%;
```
Pomax: rather than `width:100%`, since we're already using left and bottom, can we say `right: 0`?

public/editor/stylesheets/editor.less
```diff
@@ -379,16 +380,19 @@ body {
...
-     padding: 20px 100px 20px 20px;
-     position: relative;
+      padding: 20px 100px 20px 20px;
+      position: relative;
+      color: rgba(0,0,0,.6);
     }
+
     li+li {
-     border-top: 1px solid #E5E5E5;
+      border-top: 1px solid #E5E5E5;
```
Pomax: feels like this should be an `@value` of some sort like `@border-colour`

public/editor/stylesheets/projects-style.less
```diff
@@ -6,13 +6,12 @@
 }
 
 h2 {
-  font-weight: bold;
-  font-size: 22px;
-  padding: 6px 0 0 4px;
+  font-weight: 600;
+  font-size: 1.4em;
```
Pomax: do we want `em` or `rem`? For font sizes, `rem` tends to be more stable.

public/homepage/stylesheets/gallery.less
```diff
@@ -242,26 +243,13 @@
         font-size: 22px;
         padding: 0;
         margin: 5px 0 2px 0;
+        color: rgba(0,0,0,.7);
```
Pomax: we should define a mixin that we can just call as `color: flukegreyout(0.7)`

public/homepage/stylesheets/style.less
```diff
@@ -1,21 +1,12 @@
-@import "variables.less";
-
-@font-face {
-  font-family: "Fira Medium";
-  src: url("https://code.cdn.mozilla.net/fonts/woff/FiraSans-Regular.woff") format("woff");
-  font-weight: normal;
-  font-style: normal;
+body {
+  font-family: 'Open Sans', sans-serif;
```
Pomax: this feels like a pretty shallow stack, do we not want a few more websafe sans fonts in there? Just falling straight to "whatever is sans-serif" when Open Sans fails is a bit drastic.

views/editor/index.html
```diff
@@ -1,10 +1,5 @@
 {% extends "./layout.html" %}
 
-{% block head %}
```
Pomax: 👍

Pomax:
It feels like there's something odd going on for me on Windows 10 at not-retina-dpi - the help, locale picker, and username all look super fuzzy; hovering over sharpens them up, but as someone with glasses the styling for these strings feel too out of focus to be comfortable. Can we keep them sharp?

fuzzy

public/homepage/stylesheets/gallery.less
flukeout: Yeah, I'd like to tackle stuff like this in a future CSS pass.

flukeout:
I wonder what is causing the fuzzy text. If you remove the lowered opacity in dev tools, does it appear sharper?

Issue #1766, opened 2017-02-21
Give me option to remix project on My Project Page


jgmac1106:
It currently takes me four clicks to remix my own project. It should be less
* Either put a remix button on the my projects page
* Count renaming project as saving a new project kind of like hitting "Save As"

I know this issue exists but I couldn't find it so feel free to delete this as duplicate.

I have to use a lot of templates I make for students and I often start editing the original project and forget I need to open project, hit publish project, click on the link, and then hit remix.

flukeout:
#1626 deals with adding the link to the published project on that page too which would be helpful, these issues can definitely be tackled together.

rajivjhaimin:
@flukeout  I would like to start contributing in thimble, with this being my first.

humphd:
@rajivjhaimin It looks like @flukeout did some of the work for this in https://github.com/mozilla/thimble.mozilla.org/pull/1775.  @flukeout can you update this bug with what needs to happen?

flukeout:
Hey @jgmac1106 - we've made some changes that will be landing shortly that show you the Published Project link on the projects list. Now, if something is published, you can click on that and remix it. It will be way simpler than before. Does this solve your need, or are you looking to be able to duplicate projects that aren't published yet, for example.

<img width="679" alt="image" src="https://cloud.githubusercontent.com/assets/25212/23561313/9bd9fc4e-fff2-11e6-999b-0c382ff177d7.png">

Let us know. Cloning projects on the projects list makes a lot of sense too.

jgmac1106:
two clicks is better than four but one is better than two

also there is an icon for delete but not one for published and remix. If you can fit in all three why not go with live (possible earth icon) remix (arrows or something) delete (already has can)

What you suggests will work and fit my use case.

jgmac1106:
Way out of left field here but what if we started using vocab of the discipline. Remix has never localised well.

Published Version=Production
Saved but published=Staging
Remix=Fork

flukeout:
Closing as this has been addressed in #1900 - woo!

PR #1900, opened 2017-03-18
Issue #1766- Remix from projects list

kramsamujh:


flukeout:
Looking now

flukeout:
Hey @kramsamujh - nice work! I think for now, please just style the link in the same way as the "View Published" link and remove the icon. We don't really have a standard Remix icon right now.

Thank you!

kramsamujh:
@flukeout Made the changes you requested.

public/editor/scripts/project-list.js
```diff
@@ -78,6 +78,13 @@ require(["jquery", "constants", "analytics", "moment"], function($, Constants, a
 
   $("#project-list").prepend(favoriteProjectsElementList);
 
+  $(".remix-link").click(function() {
```
flukeout: @kramsamujh Instead of adding it via a ``click``handler. Could you please set the URL as an ``href`` attribute to the ``a.remix-links``. You can do that in the ``forEach`` call on line ``67``. I think this is better since then you can open the link in a new window if you want, and you can also see the URL to the remix when you hover your mouse over the link.

Thanks!

kramsamujh:
@flukeout Changed the link to use an href attribute.

flukeout:
Great work @kramsamujh - thank you for making the changes! 🎺 :1st_place_medal:

PR #1901, opened 2017-03-19
Deleted L-15 to L-19 as frame src has been deprecated.

shubham0794x:
Updated CSP definations , as directive ‘frame-src’ has been disapproved , we are using directive ‘child-src’ instead. 
As far as I am able to understand that browsers we support don't need it by referring to mozilla and microsoft developer CSP documentations .

### DIFF
--- a/server/security.js
+++ b/server/security.js
@@ -12,11 +12,7 @@ let defaultCSPDirectives = {
     "https://pontoon.mozilla.org",
     "https://mozilla.github.io/thimble-homepage-gallery/activities.json"
   ],
-  frameSrc: [
-    "'self'",
-    "https://docs.google.com",
-    "blob:"
-  ],
+
   childSrc: [
     "'self'",
     "https://pontoon.mozilla.org",


Pomax:
this depends on how we collectively feel about not supporting older browsers: `child-src` is a CSP2 directive, which is not supposed by IE11. /cc @gideonthomas @humphd

humphd:
We do support IE11 in our code, so probably not a great idea to break them.  I might be wrong about removing this.  cc @gideonthomas, @cadecairos

cadecairos:
I think we should keep it in. The directive is only deprecated, so theres no harm leaving it in for the benefit of IE11.

shubham0794x:
So Sir , I guess we don't need to delete  L-15 to L-19 , for benefit of IE11.

humphd:
@shubham0794x yes, let's leave it.  Let's modify this PR slightly.  @cadecairos, @gideonthomas, @Pomax can we do a quick audit of all the CSP rules in that file, and get @shubham0794x to update any we want altered or removed?

Pomax:
There's https://github.com/mozilla/thimble.mozilla.org/blob/master/server/security.js#L53-L55 that we might be able to get rid of, in conjuction with improving our spinner so that it doesn't need to inline the `style` attribute. Other than that short of reexamining the script, font, and style domains we rely on nothing immediately looks in need of maintenance in the CSP directives file to me.

humphd:
The inline spinner is necessary in order to have it show that before it loads the massive amounts of CSS we need, otherwise it will be white for a long time while it waits on the styles to load.

Pomax:
That sounds unlikely: much like script elements, style `<link>` elements can be placed anywhere on a page to optimize resource load.  Instead of loading a massive amount of CSS in the `<head>` element, before a document can start to be rendered, a simple solution there would be to load the styles only once necessary, such that this block, currently in the head:
```


    <link rel="stylesheet" href="/editor/stylesheets/normalize.css">
    <link rel="stylesheet" href="/resources/stylesheets/userbar.css">
    <link rel="stylesheet" href="/editor/stylesheets/publish.css">

    
    <link rel="stylesheet" href="/editor/stylesheets/editor.css">
    <link rel="stylesheet" href="/resources/stylesheets/userbar.css">
```
(with duplication, but work is already underway to dedupe a little there) is instead placed in the body, after the `spinner-container` div is set up, so that the browser shows the spinner before subsequent style resources are loaded in.

humphd:
HTTP/2 makes this better, but currently, CSS loads of external files block rendering, leaving you with the flash of unstyled content problem.  The advice in the progressive web app community, is for you to provide a minimal amount of inline styling to "boot" your page, and load secondary content styling while that is already rendered.  This is why we do it.

shubham0794x:
Sir , so we need to audit the CSP rules or we need to make some different changes ?

humphd:
@shubham0794x I think we need to close this, based on what's being discussed above.  Thanks for doing it, and apologies for not being clear and changing our mind.

shubham0794x:
@humphd  Okay Sir , It's been learning experience . I am getting more familiar with the source code.

humphd:
@shubham0794x excellent.  I agree that it wasn't a waste of time, and learning your way around the code, git, github, etc. is useful.  Let me know if you see another bug you want to try.

PR #1910, opened 2017-03-23
Fix broken headings in readme

AnshulMalik:
Heading were earlier displayed as ####something, changed to 
#### something


humphd:
Great catch, thank you @AnshulMalik!

PR #1929, opened 2017-03-31
Added buttons to toggle the value of auto-complete

### LINKED ISSUE #1920: Make auto-complete an "on-by-default" editor setting that can be turned off.
Let's make autocomplete (as seen below) an editor setting, and put it in the gear dropdown menu.

It should be on by default, but if a user turns it off, it should stay off until the next time that it is turned off. This can be stored in localStorage along with the other editor options found in that menu.

For reference, this is autocomplete in action...

hkirat:
In reference to #1920.
Needs to be tested along with [this](https://github.com/mozilla/brackets/pull/678) PR in Brackets
Fixes #1920.


flukeout:
Testing now, thanks!

flukeout:
Hey @hkirat - in testing this I found that you can turn the autocomplete on and off for HTML files, but it is always on for CSS & JavaScript files. See below...

image

Can we disable / enable the autocomplete for ALL file types?

hkirat:
Thanks for pointing that out @flukeout 
Fixed the same [here](https://github.com/mozilla/brackets/pull/678/commits/c82a850178e75da0b24daa9942168a18faf864bc)

flukeout:
Okay @hkirat - I just re-tested this and it works as expected. The editor remembers the Autocomplete setting and respects the setting for HTML, CSS & JS files! Nice.

@humphd if you can review this and the related brackets PR (https://github.com/mozilla/brackets/pull/678) that would be awesome.

locales/en-US/messages.properties
```diff
@@ -60,6 +60,7 @@ navTextSize=Text size
 navColorTheme=Color theme
 navWrapText=Wrap long lines
 navAllowJS=Allow JavaScript
+navAutoComplete=Autcomplete code
```
humphd: See my feedback in the Brackets bug about "Autocomplete" vs. "AutoComplete" vs. "Auto Complete" vs. "Auto-Complete" etc.  Let's standardize, and I'd suggest "Autocomplete."

public/editor/scripts/editor/js/fc/bramble-menus.js
```diff
@@ -79,6 +79,29 @@ define(function(require) {
...
+    }
+    // Enable/Disable Autocomplete
+    $("#autocomplete-toggle").click(function() {
+      // Toggle current value
+      var $autoCompleteToggle = $("#autocomplete-toggle");
+      var toggle = !($autoCompleteToggle.hasClass("switch-enabled"));
+
+      if(toggle) {
+        $autoCompleteToggle.addClass("switch-enabled");
+        bramble.stopAutoComplete();
```
humphd: I asked for a naming change on these methods in the Brackets PR, since we tend to use `enable/disable` in the rest of the methods.

humphd:
This is good to go, and I've left a comment in the Brackets PR for a small fix.  Once that is done, both of these can land.

humphd:
See https://github.com/mozilla/brackets/pull/678#issuecomment-292171204 for details about updating Thimble/Brackets for this code.

Great job, @hkirat!

Issue #1909, opened 2017-03-22
Prompt user when file gets saved to the browser


Taguiar:
**What I propose to do**
This is the continuation of an UI update request by @humphd. I will be implementing a new feature that display icon that shows whenever a user is not logged into his thimble account and its document are saved locally to the browser.

**Steps to Reproduce**
User must be logged out
User makes changes and wait for 5 seconds
User jumps through files

**Expected behavior.**
I want the user to be prompt whenever a save occur.

**Actual behavior.**
Nothing is displayed when the document is changed of user moves between files.

**Instrument and Methodologies**
I will be using Firefox and google Chrome browsers to guarantee cross-browser.
I will be using Visual studio Code to ensure that my indentation is compatible with what the development team requires.

Pomax:
Is this tied to a specific pre-existing issue? It feels like @humphd and @flukeout probably need to weigh in on this issue

Taguiar:
Yes, it is tied to the one that I complete in brackets.

Pomax:
@Taguiar can you mention which one a comment, so that github can setup an automatic crosslink?

Taguiar:
Is the way that I did what you wanted?!

Pomax:
looks like github does not place crosslinks for old comments... so:

see: https://github.com/mozilla/brackets/issues/607

Pomax:
yes, thank you. Now there is a way for everyone to quickly get to the original issue, and the original (now closed) issue has a linkback to this one.

PR #1932, opened 2017-03-31
[WIP] Issue #1909 Adding GUI display when saving logic is  triggered 

Taguiar:
@Pomax and @humphd 
I finished with the initial logic for GUI display whenever the user is logged out and his modifications are saved to the browser.
In terms of the design, I tried to play safe since and used whatever is already been display on the case that the user is logged on. I much rather have @flukeout giving me his thoughts about the design before I spend time in something not desirable.

**When modifications are been made:**
untitled

**When modifications have been successfully saved:**
capture1


Taguiar:
Ok I will see what I can do!

flukeout:
Testing this out now.

flukeout:
Hey OK, so is this PR only meant to address the Signed Out case when we save the files to the browser but not "to the cloud"? If so, I think there are just two states to worry about...

* Changes detected, saving begins...
  * Indicator says "Saving..."
* Changes saved to locally to browser only
  * Indicator says "Saved in browser" <- this stays until the previous case is detected again.

However, this doesn't seem all that important because it basically always saves to the browser. The more important case would be to address the "All files saved to Cloud" scenario mentioned by @humphd above. 

Will you address that case here @Taguiar ?

humphd:
I think the not-signed-in case is still really important, @flukeout, since it's not common for me to go to a site and assume that whatever I just spent an hour typing will automatically be saved when I refresh or close the tab--we need to provide some assurance.

I like your idea of always having "Saved in browser" and then changing it to "Saving..." when that is happening.  @Taguiar please evolve your code to do that, and to do it in ways that work with what I mentioned in my review.

RE: "All files saved to the Cloud" is another bug we can do on top of, and in addition to this, and it will happen in the signed-in case.

flukeout:
That sounds like what we need @humphd

Taguiar:
It is working a little bit more as you guys expected @humphd and @flukeout 
Although the internal communication is a bit hard to detect even me doing exactly the same thing as previous implementations!
I was thinking to follow this instructions here. But I would like some guidance before I do something very wrong!

humphd:
@gideonthomas is the localization master, I'd ask him for help.

gideonthomas:
@Taguiar I'm not sure I understand your question. What you have in your PR is correct. The steps you have linked to is the right way to generate the localized strings. You might just need to restart the server with `npm run restart-server`

flukeout:
Within thimble, to get strings running this also works. In the terminal form your thimble folder...

* vagrant ssh
* cd /vagrant
* npm run localize

locales/en-US/messages.properties
```diff
@@ -42,6 +42,7 @@ renameProjectSaveBtn=Save
 renameProjectSavingIndicator=Saving...
 fileSavingIndicator=Saving...
 fileSavingFailedIndicator=Saving failed, retrying...
+dirtyFileSaved=Saved in Browser.
```
humphd: `projectFilesSaved=All files saved offline.`

views/editor/userbar.html
```diff
@@ -17,6 +17,10 @@
       <div id="navbar-save-indicator" class="hide">
         {{ gettext("fileSavingIndicator") }}
       </div>
+      <div id="navbar-dirt-file">
```
humphd: Rename this to `id="navbar-saved-offline-indicator"`

public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
```diff
@@ -443,6 +443,17 @@ define(function(require) {
       }
     });
 
+    bramble.on("projectDirty", function(){
+      $("#navbar-dirt-file").addClass("hide");
+
```
humphd: Remove this blank line.

Taguiar:
Changes requested by @humphd.

views/editor/userbar.html
```diff
@@ -17,6 +17,9 @@
       <div id="navbar-save-indicator" class="hide">
         {{ gettext("fileSavingIndicator") }}
       </div>
+      <div id="navbar-saved-offline-indicator">
```
gideonthomas: you want this `div` only when the user is logged in. So this code needs to wrapped in a nunjucks `if` block similar to https://github.com/mozilla/thimble.mozilla.org/blob/master/views/editor/userbar.html#L101-L106

public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
```diff
@@ -443,6 +443,16 @@ define(function(require) {
       }
     });
 
+    bramble.on("projectDirty", function(){
+      $("#navbar-dirt-file").addClass("hide");
```
gideonthomas: You need to do this only for the "not-signed-in" case. I bet that can be done with a `!Project.getUser()` check

humphd:
Closing in favour of https://github.com/mozilla/thimble.mozilla.org/pull/2209, where I'm finishing this.

PR #1933, opened 2017-03-31
Added an error message string for archive max size limit.

MargarytaChepiga:
Regarding that for localazation: https://github.com/mozilla/brackets/pull/635


humphd:
See my note in the other PR, since I think we can reduce this down to a single string that takes a variable.

gideonthomas:
@MargarytaChepiga can you update this PR with the new string change?

humphd:
Excellent, thank you!

PR #1935, opened 2017-04-01
[WIP] Replaces the Current vagrant setup, with a puppet set up for Issue #1849

### LINKED ISSUE #1849: Figure out a better Vagrant Provisioner than shell scripts
We see a lot of contributors getting stuck when their Vagrant provisioning doesn't work for some reason.  Sometimes that's due to network errors, filesystem permissions, host OS/bios issues, the list is long.

We currently run our provisioner as a [shell script](https://github.com/mozilla/thimble.mozilla.org/blob/master/scripts/setup-services.sh), which is 100% better than what we had to do before with manual setup.  We could go further, though, and switch to a better provisioner, see https://www.vagrantup.com/docs/provisioning/

Things like Ansible, Chef, Puppet, etc. all have advanced state checking in addition to commands to be run, which means that they will ensure that your environment has whatever it needs vs. just assuming that running the command once is good enough.  Things can and do fail, and we need to deal with that.

If anyone has a background in this, we could use some guidance and help.

JohnEJames:
Fixes #1849 
The files that I replaced and added are created from the website [PuPHPet](https://puphpet.com/) where I took the information from the old  [Vagrant File](https://github.com/mozilla/thimble.mozilla.org/blob/master/Vagrantfile) and from the [Setup Service](https://github.com/mozilla/thimble.mozilla.org/blob/master/scripts/setup-services.sh).

I had a lot of pathing issues since the file I was provided was not meant to be put into another folder, but that should all be fixed.

Here is what I get now when I run vagrant up 
0b44b6d93613e20232585489faa334d4

I kept all the folders the package gave me since I was'nt sure if some I could delete or not. I think there are some I can such as all the files in the "files" folder.


humphd:
Looks like you're getting some kind of error related to the disk being full.

JohnEJames:
I was able to fix that, by deleting the Vm on virtual box, but now it's saying that port "3500 is not an available address

Here is the error:
6c2cfca11cd267b75df98d60f6f81ad9

I could try different port numbers, but that is going away from what the old vagrant system used.

JohnEJames:
@humphd  and I have talked and we believed that this is not a good time to try to implement this enhancement.

PR #1938, opened 2017-04-02
Fix #1908: Issue 1908 glob update

### LINKED ISSUE #1908: major-version update several dependencies
based on https://github.com/mozilla/thimble.mozilla.org/pull/1828, let's find all the dependencies that require a major version update, and then tackle them one by one:

- [ ] cryptr
- [ ] fs-writefile-promise
- [ ] glob
- [ ] grunt-cli
- [ ] grunt-contrib-csslint
- [ ] grunt-lesslint
- [ ] jquery
- [ ] helmet
- [ ] wolfy87-eventemitter

badrmodoukh:
I have updated the glob dependency 
I tested using "npm install" and all seems to work fine from my end
I also ran "vagrant up" and loaded thimble and all worked fine


Pomax:
difference between v7 and v6: https://github.com/isaacs/node-glob/blob/master/changelog.md#70

humphd:
Nice catch @Pomax. @badrmodoukh can you do a fix for this in this PR?

badrmodoukh:
@Pomax @humphd I added the error handler in the utils.js file
Sorry for the many commits I made

humphd:
@badrmodoukh if you need to commit something, and don't want to add more commits, use `git commit --amend --no-edit` and it will just combine the previous commit with this new stuff.  This is *only* the right thing to do when you're working on your own stuff, and haven't shared the commits with other people, since it will replace the old commit, thus destroying history (i.e., you don't want to alter public history, but private history is fine to rewrite).

server/lib/utils.js
```diff
@@ -5,11 +5,14 @@ let path = require("path");
...
-      nodir: true
-    }).map(file => path.join(root, file));
+    try {
+      return glob.sync(pattern, {
+        cwd: root,
+        matchBase: true,
+        nodir: true
+      }).map(file => path.join(root, file));
+    }
+    catch (err) { }
```
humphd: Never, ever, ever.  Not ever.  Never. Not even once.

This pattern will cause you endless amounts of grief, when you have some bug where it silently fails and you don't know why.  Always do something with the error.

badrmodoukh:
@humphd sorry about that 
I just modified it right now

Pomax:
this is good to go, thanks @badrmodoukh

PR #1941, opened 2017-04-02
Fix #1908: Issue 1908 grunt-lesslint update

### LINKED ISSUE #1908: major-version update several dependencies

badrmodoukh:
@Pomax @humphd I have updated the grunt-lesslint dependency 
I tested using "npm install" and all seems to work fine from my end
However when I ran "npm test" I got many warnings from the less and css files 

This is what it showed 
screen shot 2017-04-01 at 9 52 05 pm
screen shot 2017-04-01 at 9 52 40 pm
screen shot 2017-04-01 at 9 53 15 pm
screen shot 2017-04-01 at 9 53 45 pm

The warnings are from the rules `known-properties` and `order-alphabetical`
I'm guessing all of these files need to be modified to meet with the new update
However, I am not sure if I should change these files or ignore these rules for now by modifying the Gruntfile.js file or return to the previous version

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -72,7 +72,8 @@ module.exports = function( grunt ) {
         "./public/homepage/stylesheets/*.css",
         "./public/resources/remix/*.less",
         "./public/resources/tutorial/*.less",
-        "./public/resources/tutorial/*.css"
+        "./public/resources/tutorial/*.css",
+        "!./public/editor/stylesheets/normalize.less"
       ],
       options: {
         csslint: {
@@ -89,7 +90,6 @@ module.exports = function( grunt ) {
           "font-sizes": false,
           "ids": false,
           "important": false,
-          "known-properties": false,
           "order-alphabetical": false,
           "outline-none": false,
           "overqualified-elements": false,
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -89,6 +89,8 @@ module.exports = function( grunt ) {
           "font-sizes": false,
           "ids": false,
           "important": false,
+          "known-properties": false,
+          "order-alphabetical": false,
           "outline-none": false,
           "overqualified-elements": false,
           "qualified-headings": false,
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "grunt-contrib-requirejs": "^1.0.0",
     "grunt-execute": "0.2.2",
     "grunt-git": "^0.3.4",
-    "grunt-lesslint": "^1.1.14",
+    "grunt-lesslint": "^4.0.0",
     "grunt-npm": "git://github.com/sedge/grunt-npm.git#branchcheck",
     "grunt-update-submodules": "^0.4.1",
     "habitat": "3.1.2",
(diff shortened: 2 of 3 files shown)


Pomax:
Now would be the perfect time to fix the associated linting issues, too.

badrmodoukh:
@Pomax does that mean I need to fix all these files to meet with the new update?
Or should I just modify the Gruntfile.js file and turn the `known-properties` and `order-alphabetical` rules off

humphd:
@badrmodoukh in each of this separate PRs, fix whatever fall out you find from changing things.  This is the value of doing them all as separate PRs vs. one giant change, since you can easily deal with things one problem at a time.  Also, later on we can `git bisect` to figure out which thing is causing some unknown regression or bug based on a change that affects us more deeply than we realized at the time.

badrmodoukh:
I tried to order alphabetically for the css and less files but I kept getting warnings for the files `public/editor/stylesheets/editor.less` and `public/editor/stylesheets/publish.less` because they call a class directly. For example in the `public/editor/stylesheets/publish.less` file in `#publish-dialog` it uses `.menu-appear;` which results in a warning. This is used throughout most of the file to call other classes. 

Also for the known-properties warning the file `public/editor/stylesheets/editor.less` uses `alpha(opacity=45);` which results in a warning. This is also used throughout most of the file.

Therefore as a result, I decided to turn these rules off in the Gruntfile.

humphd:
Assigning @flukeout to see what he wants to do here.

humphd:
cc @dsych.  Let's build on this PR to also add prettier fix like we are doing in Brackets.  https://github.com/mozilla/brackets/pull/703

humphd:
Wrong bug, ignore my last comment here.

Gruntfile.js
```diff
@@ -89,6 +89,8 @@ module.exports = function( grunt ) {
           "font-sizes": false,
           "ids": false,
           "important": false,
+          "known-properties": false,
```
gideonthomas: let's remove this and fix the less errors (it seems like they are mostly all `filter: alpha(opacity);` which we can safely change to `opacity: ##`). Note that when you change filter to opacity, you will need to divide the filter value by 100, so for e.g. `45` will become `0.45`.

There is one error generated by normalize.less. We should probably not lint that file since they lint it already (we probably want to switch to npm installing this at some point). You just need to add an entry to `src` above with `"!./public/editor/stylesheets/normalize.less"`

badrmodoukh:
@gideonthomas I modified Gruntfile.js so that we can use the known-properties rule. I also added `!./public/editor/stylesheets/normalize.less` to src so that it does not get checked. I also modified `public/editor/stylesheets/editor.less` to meet with the `known-properties` rule.

gideonthomas:
I'm good with this change, nice work @badrmodoukh! @flukeout wanna do a quick review?

gideonthomas:
We would do this with webpack since that's what we're switching to. Feel free to contribute to that change. Closing in favor of that.

PR #1951, opened 2017-04-04
added active class to self._button$ element

bbhagwandeen:
Added the activate class to element self._button @flukeout 

out

### DIFF
--- a/public/editor/scripts/editor/js/fc/bramble-popupmenu.js
+++ b/public/editor/scripts/editor/js/fc/bramble-popupmenu.js
@@ -22,6 +22,8 @@ define(function(require) {
     self.close = function(e) {
       if(e) {
         e.stopPropagation();
+        self._button$.removeClass('active');
+        $("#editor-pane-nav-options").css("opacity", "0.5");
       }
       if(!self.showing) {
         return;
@@ -48,8 +50,10 @@ define(function(require) {
     // Toggle the menu on/off when the button is clicked.
     self._button$.on("click", function(e) {
       e.stopPropagation();
-
+  
       if(!self.showing) {
+        self._button$.addClass('active');
+        $("#editor-pane-nav-options").css("opacity", "0.7");
         self.show();
       } else {
         self.close();
--- a/public/editor/scripts/editor/js/fc/bramble-popupmenu.js
+++ b/public/editor/scripts/editor/js/fc/bramble-popupmenu.js
@@ -49,7 +49,7 @@ define(function(require) {
     // Toggle the menu on/off when the button is clicked.
     self._button$.on("click", function(e) {
       e.stopPropagation();
-  
+
       if(!self.showing) {
         self._button$.addClass('active');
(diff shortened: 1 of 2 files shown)


flukeout:
Testing it out now, thanks!

public/editor/scripts/editor/js/fc/bramble-popupmenu.js
```diff
@@ -22,6 +22,8 @@ define(function(require) {
     self.close = function(e) {
       if(e) {
         e.stopPropagation();
+        self._button$.removeClass('active');
+        $("#editor-pane-nav-options").css("opacity", "0.5");
```
flukeout: Now that we are adding and removing the ``active`` class to the element, we can get rid of the code that changes the CSS and instead add some styles to the ``editor.less`` file [here](https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/stylesheets/editor.less#L351)

Like this...
```
#editor-pane-nav-options.active,
#editor-pane-nav-options:hover {
    opacity: 0.7;
    filter: alpha(opacity=70);
}
```

flukeout:
Okay, great stuff - left a comment up above about how to improve this.

flukeout:
Hey @bbhagwandeen - just pinging you so you're aware I left some comments, thanks!

bbhagwandeen:
Hey @flukeout . Thanks for the comments! I'll update it ASAP. I was trying to do it that way but didn't know how to write it using the propey CSS syntax.

flukeout:
Thanks, looking forward to it @bbhagwandeen

public/editor/scripts/editor/js/fc/bramble-popupmenu.js
bbhagwandeen: Ok. With that addition, I can remove the `$("#editor-pane-nav-options").css("opacity", "0.5");`
and `$("#editor-pane-nav-options").css("opacity", "0.7");` . I still have to use `.addClass('active')` and `.removeClass('active')` with `self._button$`, however. Is that the right way? If not, let me know and i'll figure it out!

flukeout:
Testing again now, thank you!

public/editor/scripts/editor/js/fc/bramble-popupmenu.js
```diff
@@ -48,8 +49,9 @@ define(function(require) {
     // Toggle the menu on/off when the button is clicked.
     self._button$.on("click", function(e) {
       e.stopPropagation();
-
+
```
flukeout: Please get rid of the whitespace here and we can merge this in!

flukeout:
Hey @bbhagwandeen this works great, I left one final request to get rid of the whitespace that was introduced into the file. In your editor you can find an option that shows whitespace which is handy.

Thanks!

flukeout:
Great, thank you @bbhagwandeen! Good work 👍

PR #1957, opened 2017-04-05
Update to Google Analytics

### LINKED ISSUE #1794: Update our Google Analytics event tracking approach
Currently, we have methods in our code that send events to Google Analytics. All of them have an Event Category of ``thimble.mozilla.org`` and only use the **Event Action** to convey information. 

GA Events can have the following properties...

* Category
* Action
* Label (optional, but recommended)
* Value (optional)

...and we can make better use of this structure.

Our current analytics setup enables us to send an event and specify the label and value fields, like this...

```analytics.event("NormalPreview", {label: "User switched to normal preview mode vs. tutorial"});```

However, they're all stuck under the ``thimble.webmaker.org`` Category. We should refactor our code to give us control over the category, then we could divide events into logical groups...

* Home page actions 
  * Search, filter, which projects are remixed
* Editor UI
  * Font size, mobile preview, etc.
* Publishing & Project actions
  * Renaming, publishing, updating publishing etc

### PR DESCRIPTION
**Pull request by flukeout**
Fixes #1794 

**Changes**

* Ripped out the ``webmaker-analytics`` node module and put the code into ``public/editor/analytics.js`` instead
* Removed Optimizely references from the analytics code
* Removed the Fallback old way of doing Events
* Rewrote all events to have separate cateogries
* Linked events to UI actions
  * We only count a theme toggle when the user does it explicitly, not when we set it on refresh
* Paired events together so there is a "Enabled" and "Disabled" label with the same action like "Toggle Javascript"
* Added some analytics events to the new homepage Gallery

I think that's it - can I get a look-over @humphd

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -46,7 +46,7 @@ module.exports = function( grunt ) {
             "PathCache": "editor/scripts/path-cache",
             "constants": "editor/scripts/constants",
             "EventEmitter": "../node_modules/wolfy87-eventemitter/EventEmitter.min",
-            "analytics": "../node_modules/webmaker-analytics/analytics",
+            "analytics": "editor/scrips/analytics",
             "moment": "../node_modules/moment/min/moment-with-locales.min",
             "gallery": "homepage/scripts/gallery"
           },
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -46,7 +46,7 @@ module.exports = function( grunt ) {
             "PathCache": "editor/scripts/path-cache",
             "constants": "editor/scripts/constants",
             "EventEmitter": "../node_modules/wolfy87-eventemitter/EventEmitter.min",
-            "analytics": "editor/scrips/analytics",
+            "analytics": "editor/scripts/analytics",
             "moment": "../node_modules/moment/min/moment-with-locales.min",
             "gallery": "homepage/scripts/gallery"
           },
--- a/public/editor/scripts/analytics.js
+++ b/public/editor/scripts/analytics.js
@@ -1,3 +1,6 @@
+// Based on https://github.com/mozilla/webmaker-analytics
+// Licensed under the MPL 2.0 License: https://github.com/mozilla/webmaker-analytics/blob/master/LICENSE
+
 (function(global, factory) {
   if (typeof define === 'function' && define.amd) {
     define(factory);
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

**Comment by flukeout** [comment]
Also, could use some pointers on that Travis failure - things are working fine locally but I'm sure I forgot a step.

File: public/editor/scripts/analytics.js [hunk 1]
```diff
@@ -0,0 +1,188 @@
+(function(global, factory) {
```
humphd: Add a comment above this that it's based on https://github.com/mozilla/webmaker-analytics and we're using it under MPL2.0 https://github.com/mozilla/webmaker-analytics/blob/master/LICENSE

public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
```diff
@@ -309,9 +301,11 @@ define(function(require) {
     // Preview Mode Toggle
     $("#preview-pane-nav-desktop").click(function() {
       activatePreviewMode("desktop");
+      analytics.event({ category : "EditorUI", action : "Preview Mode Toggle", label : "Desktop" });
```
humphd: You slip into `EditorUI` here from `Editor UI`, which brings up a point I've wondered reading through this entire file: why don't you move these strings to some module and use properties instead, for example `AnalyticsCategory.EditorUI` or something?  Then you won't corrupt your data by making typos like this.

humphd:
R+ from me.

flukeout:
@humphd what about the Travis failure?

```
>> Error: ENOENT: no such file or directory, open
>> '/home/travis/build/mozilla/thimble.mozilla.org/dist/editor/scrips/analytics.
>> js'
>> In module tree:
>>     editor/scripts/main
>>       bramble-editor
>>         fc/bramble-ui-bridge
>>           fc/project-rename
>> Warning: RequireJS failed. Use --force to continue.
```

Gruntfile.js
```diff
@@ -46,7 +46,7 @@ module.exports = function( grunt ) {
             "PathCache": "editor/scripts/path-cache",
             "constants": "editor/scripts/constants",
             "EventEmitter": "../node_modules/wolfy87-eventemitter/EventEmitter.min",
-            "analytics": "../node_modules/webmaker-analytics/analytics",
+            "analytics": "editor/scrips/analytics",
```
humphd: s/scrips/scripts/

flukeout: I should have put that into a constant.

PR #1958, opened 2017-04-06
Fix #1956: Added hover state to Homepage Gallery Remix Button

### LINKED ISSUE #1956: Add hover state to Homepage Gallery Remix button.
Doesn't have one currently...

image

omytryniuk:
Here is the fix for the hover state. 
I have just used the same technique that was used for hovering tag buttons on the homepage (by changing the opacity of buttons). 

Here is how it looks: 

hover

Fixes #1956


flukeout:
Awesome, thank you!

omytryniuk:
@flukeout You are welcome!

PR #1965, opened 2017-04-07
Fix #1962: Scrollable Language Menu

### LINKED ISSUE #1962: Language menu is too tall, runs off page.
The language dropdown has a ton of languages in it now, and it's getting too long for my screen.

**We could either**

* Give it a max-heigh or something and have it scroll
* Rearrange into 3 columns
* Something else...

cgsingh:
Is this okay? I have it extend just above the green portion of the page.

9bc42e0dbc3f62b4f766d4f791564a6c

### DIFF
--- a/public/resources/stylesheets/userbar.less
+++ b/public/resources/stylesheets/userbar.less
@@ -385,6 +385,8 @@ a.navbar-login:hover {
     text-align: left;
     color: black;
     position: absolute;
+    overflow: auto;
+    max-height: ~"calc(100vh - 85px)";
     min-width: 185px;
     top: 43px;
     left: 0;


flukeout:
I'll take a look - thanks for submitting!

flukeout:
Okay, so - one of the problems with this approach is that it's a fixed height. If the window is shorter than that height, it still will get cut off. Also, on a normal screen, it only uses about half of the available height, so it's not that efficient. However, there's a way we can get the best of both worlds, with using the ``vh`` measurement unit. 

Try something like ``max-height: ~"calc(100vh - 85px)";`` This is using the ``vh`` unit, which stands for ``vertical-height`` (as a percentage of available height) and the ``calc()`` rule, which lets you do some math. So in our case, we'd be setting the ``max-width`` to 100% of the available height minus 85px.

cgsingh:
@flukeout Not entirely sure why I didn't take that into account. Thank you!

flukeout:
No worries, it's an obscure unit that ``vh``.

flukeout:
Hey @cgsingh - how is this coming along, were you able to make the changes? I'd love to land this improvement! Cheers!

cgsingh:
Oh, uh, I actually pushed the changes the day you suggested them. I guess I should have mentioned that. @flukeout

humphd:
@cgsingh yup, best to @ mention people when you do that.  We get too much email from github to notice new commits going up on a bug.

flukeout:
Oh yeah, sorry about that @cgsingh I see it now above your last comment - oops. Will test now, thank you!

flukeout:
OK, tested and it works great - thank you @cgsingh!

PR #1970, opened 2017-04-08
Fix #1745: add UI to allow Brackets to treat SVG as XML or Image

### LINKED ISSUE #1745: Allow option to edit .svg files instead of showing a preview.
One of the cool aspects of SVG files is that you can change them by modifying their code in a text editor. However, in Thimble, we default to the "Image Preview" mode always for .svg files. Let's add a toggle to allow editing ``.svg`` files?

image

humphd:
I've rebased the work that @peiying16 did in https://github.com/mozilla/thimble.mozilla.org/pull/1891, and fixed the backend Brackets code in https://github.com/mozilla/brackets/pull/701.

@peiying16, @Pomax: please test/review this, doing `npm run build` in Brackets before you start the server.  @peiying16 make sure I didn't break anything with this rebase, I did it quickly.


Pomax:
this works beautifully

flukeout:
I'll test as well.

flukeout:
@humph - A couple of things I've noticed.

1. After clicking around the filetree a bunch and turning this feature on and off, getting some ``Possible memory leak`` errors in the console.

image

2. When I have two ``.svg`` files and this edit option is enabled i start seeing console errors and can only edit one of the images.

image

humphd:
@flukeout thanks for this.  I saw some of the same, and fixed a bunch of leaks, but I'll test some more.  This patch is basically opening/closing files in order to have open in the right mode when the pref changes, so it is a good stress test of a lot of the code paths that don't get run too often.

flukeout:
Okay, keep me posted I'll re-test when ready.

humphd:
Brackets portion of this has landed, just need to look into the leaks you found, which I'll try to do soon.

humphd:
@flukeout I tried to reproduce your errors above, and I can't.  Are you able to give me some STR?  If you can't, I think we should land this, and file bugs on it later if it persists.

flukeout:
I'll try again and let you know.

peiying16:
@humphd  I just click on the files hundred of times. You can try in this way.


image

humphd:
I still can't make this happen, and I've tried everything you are suggesting.  Given that you two can both reproduce this, please help me narrow this down.  I wonder if we're seeing a bug that is actually related to something other than what's in this PR.

I'm wondering specifically if it's related to some file(s) you have in your project that I don't.  Can you test a bit with different files? When I only use `.svg` files with the default project, none of this happens.

Also, can you expand those leak warnings so I can see the stack?  What is happening that's triggering them?

flukeout:
Hey @humphd humph, here's the project I'm using to test this...

[svg-tests.zip](https://github.com/mozilla/thimble.mozilla.org/files/917172/svg-tests.zip)

Here's how I can reproduce the "can't edit other .svg file" error (the second in my previous comment).

* In your project, have just a default .svg
* Drag in one of the svgs I provided
* Turn on "edit svg" feature
* Reload the editor
* Drag in the other svg
* Now click between the two svgs

On my end, this reliably makes it so that only one of the svgs can be edited, the code for the other 
one never shows up, and there are bunch of console errors.

star


image

humphd:
I did a PR on Brackets to try and fix this: https://github.com/mozilla/brackets/pull/713

peiying16:
For the memory leak, I think it only happens on text file such as HTML, js, CSS.  One of the error is:
image Does it mean you register the listener again and again or you need to destroy it when choosing another file? I'm not sure.

humphd:
This event is registered here:

* https://github.com/adobe/brackets/blob/f028b3ecdf2017498196f39ddbbe4a2364366c92/src/LiveDevelopment/MultiBrowserImpl/documents/LiveHTMLDocument.js#L71

I added code to remove it when the live doc is closed:

* https://github.com/mozilla/brackets/commit/baf0249ab3003ab688b8db59cfa0ee91e19bca1f#diff-c03aea40ccd499bf943db9e42de5b357R142

I'm not clear why I can't get this to trigger in my build (i.e., the error).

humphd:
@peiying16 I wonder if it's these: https://github.com/mozilla/brackets/blob/master/src/LiveDevelopment/LiveDevMultiBrowser.js#L547-L584

Can you do some testing to see if this is where those leak warnings are originating?

humphd:
I think I have a fix for this in https://github.com/mozilla/brackets/pull/715.

humphd:
Nice job, you two, you found multiple bugs testing this, neither of which are related to this bug.  The best way to find bugs is to change code.

humphd:
I've also filed https://github.com/mozilla/brackets/issues/716 to get rid of another exception I see in the screenshots above.

flukeout:
I'll retest after @gideonthomas lands related Brackets PRs

humphd:
Now that all the bugs this found in brackets are landed, I'm merging this too.  Congrats, @peiying16, we did it!

peiying16:
Good job!!

PR #1971, opened 2017-04-09
Fix#1921: Extended the Gruntfile to include eslint 

badrmodoukh:
@Pomax I installed eslint for grunt and modified the Gruntfile.js file to include the eslint task.
The only rule I can think of to enable and check for is the indent rule. I added it in the .eslintrc.json file and specified that the indent needs to be 2 spaces. If there are any other rules I need to add please let me know.
I also specified the ecmaVersion for the parser to be 6 because if I don't it will give many parsing errors like the keyword `let` is reserved and `const` is reserved.  
Currently when I run `npm test` I get 7 errors because some files don't use 2 spaces. This is how the errors look: 
screen shot 2017-04-09 at 2 59 35 pm

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,15 @@
+{
+  "env": {
+    "node": true,
+    "browser": true
+  },
+
+  "rules": {
+    "indent": ["error", 2]
+  },
+
+  "parserOptions": {
+    "ecmaVersion": 6
+  }
+  
+}
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -131,10 +131,21 @@ module.exports = function( grunt ) {
           ]
         }
       }
+    },
+    eslint: {
+      src: [
+        "public/homepage/**/*.js",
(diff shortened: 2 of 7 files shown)


badrmodoukh:
I fixed the above errors

humphd:
cc @dsych.  Let's build on this PR to also add prettier fix like we are doing in Brackets.  https://github.com/mozilla/brackets/pull/703

flukeout:
Can we merge this @humphd - seems like the problems were addressed.

humphd:
Tested this again today, and it needs a few things:

* This needs to get rebased onto Master, where I found new errors
* The jshint stuff needs to get removed, since eslint does a better job
* We need a set of rules
* We need to fix the errors it produces.

gideonthomas:
Since we're switching to webpack, we would need to run eslint there. Feel free to take on that issue once the webpack code is merged in. Closing, in favor of that.

PR #1987, opened 2017-04-12
Fixes #1815 - Feature: Delete multiple projects

### LINKED ISSUE #1815: Request: Ability to delete multiple projects at once
taking inspiration from imgur and google's android client, it would be nice if there was a "batch delete" option, where clicking the "trash" icon marks an element as "to be deleted" with a new button "DELETE THIS" popping up at the top of the page, so you can click more trashcans and select more than 1 project to delete in the same batch, with the "are you sure?" dialog not popping up until you press the button up top.

kramsamujh:
@flukeout Made a new branch with just my changes. I closed the other PR I had open for this issue.


public/editor/scripts/project-list.js
```diff
@@ -81,44 +81,88 @@ require(["jquery", "constants", "analytics", "moment"], function($, Constants, a
...
   $("#project-list").prepend(favoriteProjectsElementList);
 
+  var projectsToDelete = [];
+
   $(".project-delete").click(function() {
-    // TODO: we can do better than this, but let's at least make it harder to lose data.
-    if(!window.confirm("{{ deleteProjectConfirmationText }}")) {
-      return false;
+    if($(".delete-button").length === 0){
+      var deleteBtn = document.createElement("div");
```
Pomax: we have jQuery available here, let's use that. This changes this line and the next two into the following single line:
```
var deleteBtn = $("<div class='delete-button'>Delete Projects</div>");
```
However, I'm pretty sure that "Delete Projects" should be a locale string, so we'll need to be more careful here:
```
var deleteBtn = $("<div class='delete-button'></div>").text("{{ DeleteProjects }}");
```
The reason we need to do this is because user-generated-content can contain exploit code, and you should only ever add that as true text (via `.textContent` or jQuery's `text`), otherwise any script code that might have been maliciously slipped in will be able to run in our site, and that would be bad =)

public/editor/scripts/project-list.js
```diff
@@ -81,44 +81,88 @@ require(["jquery", "constants", "analytics", "moment"], function($, Constants, a
...
+      deleteBtn.appendChild(document.createTextNode("Delete Projects"));
+      deleteBtn.className = "delete-button";
+      deleteBtn.onclick = function() {
```
Pomax: do not use `on...` handlers in modern JS. Instead use `addEventListener` in plain JS, but since we have jQuery, if we use the above-mentioned code we can instead do:
```
deleteButton.on("click", function(evt) { .... });
```

public/editor/scripts/project-list.js
```diff
@@ -81,44 +81,88 @@ require(["jquery", "constants", "analytics", "moment"], function($, Constants, a
...
+        // TODO: we can do better than this, but let's at least make it harder to lose data.
+        if(!window.confirm("{{ deleteProjectConfirmationText }}")) {
+          return false;
+        }
+        
+        Array.prototype.forEach.call(projectsToDelete, function(project){
```
Pomax: Based on 84 this is already an array, so we don't need to so prototype indirection. Also, if you're going to delete things in an array it is always safer to do so using reverse iteration, so taking those things together, and the knowledge that we want to end up with all the projects-to-delete having been deleted, let's rewrite this to something more concise:
```
while (projectsToDelete.length > 0) {
  deleteProject($(projectsToDelete.shift()), projectsToDelete.length === 0);
}
```
That's super short, and lets us define the "what should happen" code as its own function, which is better than defining it on the fly only for the duration of the iteration:
```
function deleteProject(project, lastProjectToDelete) {
  var projectId = project.data("id");
  var projectElementId = project.attr("id");
  // we're going to delete this element so event handlers should be garbage collected automatically
 
  analytics.event("DeleteProject");

  deleteProjectFromDB(projectId, lastProjectToDelete, onDeleteSuccess, onDeleteError);
}
```
This relies on three functions, `deleteProjectFromDB` which does the request part, handing off success to `onDeleteSuccess` and handing off errors to `onDeleteError`, making sure to pass `lastProjectToDelete` along as well, as it will let the code know whether it just deleted the last-project-in-the-list.

public/editor/scripts/project-list.js
```diff
@@ -81,44 +81,88 @@ require(["jquery", "constants", "analytics", "moment"], function($, Constants, a
...
+            if(request.status !== 204) {
+              console.error("[Thimble error] sending delete request for project ", projectId, request.status);
+            }
+          });
+          request.fail(function(jqXHR, status, err) {
+            err = err || new Error("unknown network error");
+            console.error(err);
+          });
+
+          project.hide({
```
Pomax: This should happen in `onDeleteSuccess`, because a delete error means the project was not deleted.

public/editor/scripts/project-list.js
```diff
@@ -81,44 +81,88 @@ require(["jquery", "constants", "analytics", "moment"], function($, Constants, a
...
+            duration: 250,
+            easing: "linear",
+            done: function() {
+              project.remove();
+            }
+          });
+        });
+        projectsToDelete = [];
```
Pomax: due to the `shift()` call in the above review comment, this will automatically be true once iteration is done.

public/editor/scripts/project-list.js
```diff
@@ -81,44 +81,88 @@ require(["jquery", "constants", "analytics", "moment"], function($, Constants, a
...
-      },
-      type: "DELETE",
-      url: "/" + locale + "/projects/" + projectId,
-      timeout: Constants.AJAX_DEFAULT_TIMEOUT_MS
-    });
-    request.done(function() {
-      if(request.status !== 204) {
-        console.error("[Thimble error] sending delete request for project ", projectId, request.status);
+    var projectSelector = "#" + project.attr("id");
+    var deleteIndex = projectsToDelete.findIndex(function(x) {return x.attr("id") === project.attr("id");});
```
Pomax: don't "save time by writing short code", that's what minifiers are for. Just write functions like these out with words that people can immediately understand:
```
var deleteIndex = projectsToDelete.findIndex(function(element) {
  return element.attr("id") === project.attr("id");
});
```
However, can we use `projectsToDelete.indexOf(project)` instead?

public/editor/scripts/project-list.js
```diff
@@ -81,44 +81,88 @@ require(["jquery", "constants", "analytics", "moment"], function($, Constants, a
...
+    var disabledLink = {"cursor":"default", "text-decoration":"none"};
+    var enabledLink = {"cursor":"pointer", "text-decoration":"underline"};
+    
+    if(deleteIndex === -1) {
+      projectsToDelete.push(project);
+      $(projectSelector + " .edit-link").bind("click", false).css({"cursor":"default"});
```
Pomax: instead of `.bind("click", false)`, use `.off("click")`.

public/editor/scripts/project-list.js
```diff
@@ -81,44 +81,88 @@ require(["jquery", "constants", "analytics", "moment"], function($, Constants, a
...
+      if($(projectSelector + " .published-link").length !== 0) {
+        $(projectSelector + " .published-link").bind("click", false).css(disabledLink);
```
Pomax: same notes here, there should not be a reason to use jQuery's `bind` in modern JS

public/editor/scripts/project-list.js
```diff
@@ -81,44 +81,88 @@ require(["jquery", "constants", "analytics", "moment"], function($, Constants, a
...
-      easing: "linear",
-      done: function() {
-        project.remove();
+      project.css("background", "rgba(255,0,0,.08)");
+      $(this).closest(".project-delete").html("Cancel");
+    } else {
+      projectsToDelete.splice(deleteIndex, 1);
+      $(projectSelector + " .edit-link").unbind("click", false).css({"cursor":"pointer"});
+      if($(projectSelector + " .published-link").length !== 0) {
+        $(projectSelector + " .published-link").unbind("click", false).css(enabledLink);
```
Pomax: what did you intend for `unbind("click", false)` to do here?

public/editor/scripts/project-list.js
```diff
@@ -81,44 +81,88 @@ require(["jquery", "constants", "analytics", "moment"], function($, Constants, a
...
+        $(projectSelector + " .remix-link").unbind("click", false).css(enabledLink);
       }
-    });
+      project.css("background", "");
+      $(this).closest(".project-delete").html("<div class='icon-garbage-can'></div> {{prjListDeleteProjectBtn}}");
```
Pomax: `${this}.closest` is setting off some alarm bells: we should know which element we're putting this HTML code into. However, the bigger alarm bell is that this explicitly bakes a security exploit into our code: if a localization intentionally contains script code as part of the `prjListDeleteProjectBtn` translation then that code *will run*. Instead, Create the div, add it, then create a span with the locale string as `text`, *not* as `html`, and then add that.

public/editor/scripts/project-list.js
gideonthomas: @kramsamujh, can't we just have the html in the nunjucks template and have it be `display: none` and toggle that? I feel like that would be cleaner than using js to inject html.

public/editor/scripts/project-list.js
gideonthomas: +1 to using `indexOf` since `findIndex` is not supported in IE at all.

flukeout:
Hi @kramsamujh is there an email I can reach you at? I'm collecting information for a contributors page on the Thimble website and I'd like to include you. Can you please email me at ``luke@mozillafoundation.org``? Thanks!

public/editor/scripts/project-list.js
kramsamujh: I tried to use `indexOf` here originally but it does not seem to work when looking for objects in an object array. Since `findIndex` is not supported, the next best solution that would be compatible seems to be writing a simple loop to find the index manually. I used this implementation in my latest changes but if there is a better solution I'd be glad to change it.

public/editor/scripts/project-list.js
kramsamujh: I was not able to use `.off("click")` since the links are all anchor tags and I don't define a `click` behavior so using `.off("click")` did not seem to effect them at all. I found that the common approach to disabling the href links was to set a `click` behavior to use `.preventDefault()`. This `click` behavior is then turned off to activate the link again as can be seen in my latest changes.

public/editor/scripts/project-list.js
kramsamujh: My idea behind the binding and unbinding was to disable the href links when a project was marked as "to be deleted". I was having trouble using `.off("click")` as noted in my comment above and this solution seemed to work however in my latest changes I was able to use `.off("click")` and `.preventDefault()` to achieve the same functionality.

kramsamujh:
@Pomax @gideonthomas Made the first set of changes you guys requested.

flukeout:
Re-testing now, thanks @kramsamujh

flukeout:
Also, could you please use the Red Delete button styling for the Delete button as found in the publish dialog?

image

It's the ``#publish-button-unpublish`` button and the styling can be found in ``publish.less``.

Thank you!

kramsamujh:
@flukeout I changed the button to use the red styling.

public/editor/scripts/project-list.js
```diff
@@ -81,19 +81,19 @@ require(["jquery", "constants", "analytics", "moment"], function($, Constants, a
...
-    }
+  var projectsToDelete = [];
 
-    var project = $(this).closest(".project");
+  function deleteProject(project){
     var projectId = project.attr("data-project-id");
     var projectElementId = project.attr("id");
     $("#" + projectElementId + " > .project-title").off("click");
 
-    analytics.event({ category : analytics.eventCategories.PROJECT_ACTIONS, action : "Delete Project" });
```
flukeout: Please put back this analytics event. It's the more recent version.

public/editor/scripts/project-list.js
```diff
@@ -119,6 +122,81 @@ require(["jquery", "constants", "analytics", "moment"], function($, Constants, a
...
+      if($(projectSelector + " .published-link").length !== 0) {
+        $(projectSelector + " .published-link").on("click", function(link) {
+          link.preventDefault();
+        }).css(disabledLink);
+        $(projectSelector + " .remix-link").on("click", function(link) {
+          link.preventDefault();
+        }).css(disabledLink);
+      }
+
+      project.css("background", "rgba(255,0,0,.08)");
```
flukeout: Please apply a class to this element and change the background with a CSS rule instead of applying styles directly in JavaScript

flukeout:
Hey @kramsamujh - please see my comments above. Two changes requested from me:

* You changed the analytics event call to an older version, please use the more recent one.
* Please apply the styling to the element via a CSS rule and not directly in JavaScript, you can add a class name to it like ".pending-delete"

Thank you.

public/editor/scripts/project-list.js
kramsamujh: Woops I think i might have messed this up when I was trying to rebase my branch.

kramsamujh:
@flukeout Made the changes you requested.

flukeout:
Testing now.

flukeout:
Ok @kramsamujh this is getting so close!! But check this out...

image

This overlap is caused because each button is positioned with ``absolute`` positioning. Why is this a problem? Because in other languages the label of the green button (and delete) will change, so we can't rely on a very specific position.

Here's how to fix that:

* First, in the markup, put both buttons above the <h2> element with the title
* Don't use ``position: absolute;`` anymore. Instead use...
  * ``float:right;`` for both buttons.
  * This will make them to go to the right, and they'll adapt to different length labels
* Lastly, use ``position:relative`` on the Delete button wrapper so that the icon positioning still works

Thanks!!

kramsamujh:
@flukeout Made some more changes to fix the problem with the positioning of the buttons.

gideonthomas:
@kramsamujh, you have changes in here that were already made in mozilla/master (for e.g. the analytics require config), can you rebase onto master, and do not have those changes show up in your commits please?

gideonthomas:
@kramsamujh before you rebase, I have opened up a PR against your branch https://github.com/kramsamujh/thimble.mozilla.org/pull/2 with a couple of fixes and l10n changes. Can you merge that in or make those changes and then rebase your branch onto master?

gideonthomas:
@kramsamujh any updates on this?

kramsamujh:
@gideonthomas I merged your changes and rebased with master. Terribly sorry for the long delay I must have missed the changes you requested before.

public/editor/scripts/project-list.js
```diff
@@ -102,21 +95,66 @@ require(["jquery", "constants", "analytics", "moment"], function($, Constants, a
       url: "/" + locale + "/projects/" + projectId,
       timeout: Constants.AJAX_DEFAULT_TIMEOUT_MS
     });
-    request.done(function() {
+    request.done(function(){
       if(request.status !== 204) {
-        console.error("[Thimble error] sending delete request for project ", projectId, request.status);
+        console.error("[Thimble error] Failed to delete project ", projectId, " with status: ", status);
```
gideonthomas: looks like I missed this. Can you change `status` back to `request.status`?

flukeout:
In the meantime, testing this out.

flukeout:
Hey @kramsamujh - this is good to go as soon as you address @gideonthomas's request. Nice work! Let's get it merged soon.

kramsamujh:
@gideonthomas @flukeout I made the small change that was requested. I also had to change the import of colors.less to use the variables.less file in the project-list. I just realized this was changed in master.

gideonthomas:
Very nice work @kramsamujh! This is a pretty cool feature that a lot of people have been waiting for. Thanks for implementing this!

Issue #1952, opened 2017-04-04
Publishing projects without index.html fails to locate it.


dsych:
When publishing project without index.html file, the resulting link leads to a broken project. 

![no_index](https://cloud.githubusercontent.com/assets/16144679/24671349/39ae0432-193f-11e7-948d-3c850add8572.png)

While the resulting page has a suggestion on potential cause of the problem, however the existence of index.html should be checked at the publish time, and feedback should provided. Maybe block the publish button and provide feedback here:

![no_index_2](https://cloud.githubusercontent.com/assets/16144679/24671534/c70bb626-193f-11e7-8cf0-696fc3e7813b.png)

dsych:
This is related https://github.com/mozilla/brackets/pull/620

flukeout:
Hey @dsych - we've been down this road before and ended up with the solution to add the help text on that page. Before that change, we got quite a few support emails about "Projects not working" & the change seems to have helped a lot.

Your suggestion is one of the few we discussed, and I think it would still be a good idea. You can see some background discussion about it in this issue.

https://github.com/mozilla/thimble.mozilla.org/issues/1569#issuecomment-259006400

I'm not sure if totally blocking publishing is the right idea, there could be some valid cases where you wouldn't need an index page. We could just add a warning or something like that.

dsych:
@flukeout in https://github.com/mozilla/thimble.mozilla.org/issues/1569#issuecomment-259006400 , your first idea with the pop-up could solve the problem before the publishing occurs. Moreover, now that I am working on the API that will allow us to check project for the index.html file, implementing the pop-up message should not be an issue.

dsych:
Maybe we can just reuse the same publishing window, however when a project is lacking the index.html file, the "Publish" button will be of different colour, say orange. There will also be a highlighted reasoning to the left of the cancel.

MargarytaChepiga:
Can I work on this bug? @humphd

dsych:
@MargarytaChepiga @humphd this is part of my previous release.

humphd:
@MargarytaChepiga I think @dsych is tackling this, based on his earlier work.

dsych:
@flukeout did you have a chance to decide what is the better way to warn user about the absence of index.html?

flukeout:
Hey @dsych - not yet, BUT - please go ahead and start on this. We'll assume that we need two things:

* A class on the publish dialog, something like ".no-index"
[...]

PR #1994, opened 2017-04-13
[WIP] Publishing project without index.html #1952

dsych:
On click of Publish button with missing index.html, the publish event is not dispatched  to prevent publishing and message in displayed. 

cc @flukeout


Pomax:
@dsych make sure to run `grunt` locally so that you don't end up with a PR that fails testing

dsych:
@Pomax @flukeout I had to store a local reference to bramble in order to use its API in publish. "bramble_" name is kind of out of sync with naming convention of other local variables, however I could not make jshint pass the build otherwise. If I used this.bramble or publisher.bramble it kept failing the build stating that local bramble was declared, but never used. If you know a workaround, I would be glad to implement it.

Other than that, this PR is ready for review.

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -76,6 +78,8 @@ define(function(require) {
       publisher.showUnpublishedChangesPrompt();
     });
 
+    bramble_ = bramble;
```
humphd: Can you not just pass `bramble` when you call `Publisher.publish`?

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -129,6 +133,15 @@ define(function(require) {
     if (publisher.publishing) {
       return;
     }
+
+    if(!bramble_.hasIndexFile()){
+        dialog.buttons.index_message.css("display","block");
+        return;
+    }
+    else{
```
humphd: `} else {`

dsych:
@humphd fair enough, please review the changes.

> https://github.com/mozilla/thimble.mozilla.org/pull/1994/files/9d602c0a5320194d853b69e7766010fddb0936ab#r111757226

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -30,7 +30,8 @@ define(function(require) {
         publish: $("#publish-button-publish"),
         update: $("#publish-button-update"),
         unpublish: $("#publish-button-unpublish"),
-        parent: $("#publish-buttons")
+        parent: $("#publish-buttons"),
+        index_message: $("#no-index")
```
humphd: Is `index_message` your addition, or was it already in our tree?  In JS, we favour `thisStyle` vs. `this_style` (the latter is more common in Python).

public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -122,13 +123,21 @@ define(function(require) {
...
     var dialog = publisher.dialog;
 
     if (publisher.publishing) {
       return;
     }
+
+    if(!bramble.hasIndexFile()){
+        dialog.buttons.index_message.css("display","block");
+        return;
+    } else {
```
humphd: You never need/want an `else` after you have a return.  It makes you indent code that doesn't actually need to be.  Always do:

```js
if(...) {
  ...
  return;
}

// no else, keep going here with "else" case
```

views/editor/publish.html
```diff
@@ -37,6 +37,7 @@
   </div>
 
   <div id="publish-buttons">
+      <div id="no-index"> {{ gettext("noIndexFound") }} </div>
```
humphd: Indent issue here.

dsych:
@humphd I applied the fixes, please review.

flukeout:
Testing it out now.

flukeout:
Hey @dsych - this works for me. A couple of changes please...

* Please hide the error message when the publish dialog is closed.
* Please update the string to
  * ``Please add an <strong>"index.html"</strong> file before Publishing your Project.``
* Please update the ``#no-index`` styling to the following...

```
  color: #ca481e;
  float: left;
  position: relative;
  top: -1px;
  display: none;
  width: 60%;
  font-size: .9em;
```
It should appear like this...

image

Cheers! I think this will help - had another support email today about a missing index.html page. We could also link that error message to a help article down the road.

dsych:
@flukeout made fixes as you requested, please review.

humphd:
Either in this bug, or a follow-up, should that warning not be made into a button, so you can take action as a user as soon as you see it?  I don't want to derail or slow this down, but something to consider.

flukeout:
If we could add a link that fires the "add HTML file" method (same as the file menu dropdown), I think that would be great. What do you think @dsych

dsych:
@flukeout @humphd I think that while this button/link would be a really handy utility for an experienced programmer, implementing such feature will have a negative effect on the learning experience. Thimble is a learning platform, which means that it should not fix your problems, but instead educate you. So I believe it will be better if we just add a "Learn more." link after the message, that will take you the page with a full explanation of why index.html file is important. 

no_index_popup

We already have a page like this : 

no_index_details

We can just expand the message that is already there. 
Finally, adding the index.html file is just a matter of two mouse clicks and a couple key strokes. However, going thought this process once, will insure that students will be less likely to repeat the same mistake.

gideonthomas:
@dsych, I think you bring up some good points that also indicate that we need to evaluate the need for that feature even further. 

Consequently, I would ask that as @humphd mentions, you file a follow-up bug (or find an existing one that already addresses the automatic generation of the index.html file) and add your thoughts there and link to this PR. 

I suggest we scope this PR to the fix you have already made of simply providing feedback during publish (possibly with the "Learn More" part linking to something based on @flukeout's discretion).

dsych:
@flukeout So what do you think about the "Learn more" feature?

flukeout:
Sorry for the slow reply, I was traveling. Let's move forward with the "Learn More" link for now, we can link it to a page on the repo Wiki that explains the steps to adding an ``index.html`` file. The instructions on the "404" page for projects without an index don't seem good enough, I got a support email about it earlier today.

dsych:
@flukeout [here](https://github.com/dsych/thimble.mozilla.org/wiki) is a sample of a wiki page that could be linked to the "Learn more". I think, the end version would probably be better of on the CDOT wiki.

flukeout:
Testing this out now.

flukeout:
Closing this, but have built upon this work and opened a new PR here -
 https://github.com/mozilla/thimble.mozilla.org/pull/2238

Thanks very much for your effort and work on this valuable feature @dsych 👍 👍

PR #1997, opened 2017-04-14
Fixing line Height is too small - Issue #1988

### LINKED ISSUE #1988: Line height is too small
screen shot 2017-04-12 at 2 03 53 pm

In the publish dialog, the bottom of the `g` in the url is cut off because the line height is set to the same value as the font-size we're using. I think we probably just need to either remove the line-height or set it to something bigger. If we want to maintain the same space in the blue box, we can reduce the padding.

cc @flukeout

morrme:
Modified padding and removed line-height for the publish-link.

Fixes #1988

### DIFF
--- a/public/editor/stylesheets/publish.less
+++ b/public/editor/stylesheets/publish.less
@@ -61,14 +61,13 @@
     background: rgba(49, 134, 193,.1);
     border: solid 1px rgba(49, 134, 193,.2);
     border-radius: 2px;
-    padding: 12px 12px 10px 12px;
+    padding: 10px 12px 10px 12px;
     margin-bottom: 12px;
 
     a {
       overflow: hidden;
       text-overflow: ellipsis;
       display: inline-block;
-      line-height: 1em;
       color: #3186c1;
       text-decoration: none;
       font-weight: 600;
--- a/public/editor/stylesheets/publish.less
+++ b/public/editor/stylesheets/publish.less
@@ -67,7 +67,7 @@
     a {
       overflow: hidden;
       text-overflow: ellipsis;
-      display: inline-block;
+      display: block;
       color: #3186c1;
       text-decoration: none;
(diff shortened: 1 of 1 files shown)


morrme:
@gideonthomas @flukeout For your review 🤓

flukeout:
Testing it out now @morrme

flukeout:
Hey @morrme thanks for this. The blue box still has a wider space below the link...

image

Let's fix this up. We can change the ``#link_publish_link`` (lol @ that ID) to have ``display: block'' instead of ``display:inline-block``. If you could update your PR with that change, then we'll be set, thank you!

morrme:
@flukeout OK, making the change now. Should I remove the `padding` property as well?

flukeout:
@morrme do you mean on the parent element of the link? No let's leave that. The link element itself doesn't have any padding, right?

morrme:
@flukeout thanks! done.

flukeout:
Ahhhh @morrme - I'm an idiot. I remember now why ``display: inline-block`` is better.

* For one, try to highlight to copy paste the link now, it's harder because the link spans the whole blue box and you start dragging it.
* Second, when you hover anywhere over the blue box, the link gets an underline and you can click it, which is weird.

Sorry about that!

So what we should do it...

* make it ``inline-block`` again
* change the padding of the parent element to ``padding: 10px 12px 5px 12px;`` so that the blue box has even padding around the link. Should look like this still...
* Also, then lets slap a ``max-width: 100%;`` on the link element, so that if the URL is longer than the allowed space, it will properly add the ``...`` at the end and cut it off.



Thanks!!

morrme:
@flukeout Not at all! In making that attempt, I learned something new about the nuances between the two attributes.  :smiley:

flukeout:
Awesome, thank you @morrme - merging!

PR #2001, opened 2017-04-17
Show tutorial by default when available at project load. Fixes #1776

### LINKED ISSUE #1776: When remixing, if a tutorial is present, show it in the preview by default.
The projects that we'll feature on the homepage all have tutorials. However, when you remix one of them, the tutorial isn't displayed prominently, and there's a good chance it might get missed.

If possible, when remixing, can we show the tutorial in the preview pane if it's available? This means that in the header, the Tutorial option should be enabled like this...

image

In the editor view, however, we should still show whatever the default file logic is (first HTML file in the list I think?)

babhishek21:
As discussed in #1776. 

Does two things:
+ Fixes a typo that was preventing the `bramble:tutorialVisibilityChange` event from being interpreted correctly in the Thimble client.
+ Shows the Tutorial by default if available. If not, shows Index HTML file. Otherwise falls back to showing the first HTML file available in file list. This preserves behaviour from previous code.

Note that [Bramble #722](https://github.com/mozilla/brackets/pull/722) must land before this PR for it to work.

### DIFF
--- a/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
+++ b/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
@@ -290,7 +290,7 @@ define(function(require) {
 
     // Programmatic change to tutorial vs. regular preview mode from Bramble
     bramble.on("tutorialVisibilityChange", function(data) {
-      if(data.visibility) {
+      if(data.visible) {
         setTutorialPreview();
       } else {
         setNormalPreview();
--- a/public/editor/scripts/project/project.js
+++ b/public/editor/scripts/project/project.js
@@ -244,15 +244,21 @@ define(function(require) {
               return callback(err);
             }
 
-            // Look for an HTML file to open, ideally index.html
-            var indexPos = 0;
+            // Look for an HTML file to open
+            // Preferably `tutorial.html` (if available) or `index.html`
+            // Otherwise the first html file available
+            var indexPos = 0, tutorialPos = -1;
             found.forEach(function(path, idx) {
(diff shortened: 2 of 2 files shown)


public/editor/scripts/project/project.js
```diff
@@ -244,15 +244,21 @@ define(function(require) {
               return callback(err);
             }
 
-            // Look for an HTML file to open, ideally index.html
-            var indexPos = 0;
+            // Look for an HTML file to open
```
humphd: There's another code change happening in https://github.com/mozilla/thimble.mozilla.org/pull/1822 you should be aware of, and integrate with, once it lands.

public/editor/scripts/editor/js/fc/bramble-ui-bridge.js
```diff
@@ -290,7 +290,7 @@ define(function(require) {
 
     // Programmatic change to tutorial vs. regular preview mode from Bramble
     bramble.on("tutorialVisibilityChange", function(data) {
-      if(data.visibility) {
+      if(data.visible) {
```
humphd: Interesting, nice catch.  Looks like I documented this incorrect in https://github.com/mozilla/brackets/commit/b31163e3a7 (the README needs a fix in Brackets, can you do that too?)

public/editor/scripts/project/project.js
babhishek21: I guess I'll wait for #1822 to finish merging, pull in locally, rebase and then move ahead from there.

PR #2007, opened 2017-04-19
Fix #2006- Changing light/dark themes -Uncaught TypeError

### LINKED ISSUE #2006: Uncaught TypeError: Cannot read property 'EDITOR_UI' of undefined - analytics.eventCategories
When changing dark/light themes you get an error.
other

bramble-menus.js:195 Uncaught TypeError: Cannot read property 'EDITOR_UI' of undefined
      at HTMLDivElement.toggleTheme (bramble-menus.js:195)
      at HTMLDivElement.dispatch (jquery.js:4435)
      at HTMLDivElement.r.handle (jquery.js:4121)

edyedy123:
I fixed this by commenting out:
0e502bacbadafc306daf3d9c033135ca

Now it looks like:
aaa

I commented the lines out in both en-GB and en_US because I know that one of them over writes the other but I wasn't sure witch one. Also the changes were made to lines 195 and 198 in both files, i'm not sure why it +ed all the lines.
Fixes #2006

### DIFF
--- a/client/en_GB/editor/scripts/editor/js/fc/bramble-menus.js
+++ b/client/en_GB/editor/scripts/editor/js/fc/bramble-menus.js
@@ -1,0 +1,324 @@
+define(function(require) {
+
+  var $ = require("jquery");
+  var PopupMenu = require("fc/bramble-popupmenu");
+  var analytics = require("analytics");
+
+  function setupUserMenu() {
+    PopupMenu.create("#navbar-logged-in .dropdown-toggle", "#navbar-logged-in .dropdown-content");
+  }
+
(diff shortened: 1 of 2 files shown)


humphd:
The files you changed are *both* generated from the versions in `public/...`, so whatever change you make needs to happen there vs. in `client/...`.

The `analytics` module is here: https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/analytics.js

It looks like this is all correct to me.  As I mention in the bug you filed, I can't reproduce this.  Can you confirm that your tree is up to date?

PR #2008, opened 2017-04-20
Issue 1991

### LINKED ISSUE #1991: Add some DNS, Resource prefetch and preload meta info to Thimble
Some great speedups discussed in https://csswizardry.com/2013/01/front-end-performance-for-web-designers-and-front-end-developers/#section:dns-prefetching and https://css-tricks.com/prefetching-preloading-prebrowsing/.

In https://github.com/mozilla/thimble.mozilla.org/issues/1524 we're undoing my hacky way of preloading these big Bramble resources.  We can use this proper way instead.

Mordax:
Cleaning up the preloading of Bramble by using prefetch. Fixes #1991. 

(Excuse the slight fumbling of this, I thought I could pull within the same issue number)

*Realized my error and in the process of fixing it*

### DIFF
--- a/public/homepage/scripts/main.js
+++ b/public/homepage/scripts/main.js
@@ -30,15 +30,6 @@ require.config({
   }
 });
 
-// While the user is reading this page, start to cache Bramble's biggest files
-function preloadBramble($) {
-//Using prefetch for easier cacheing and preloading of resources
-  <link rel="prefetch" href="/dist/styles/brackets.min.css">
-  <link rel="prefetch" href="/dist/bramble.js">
-  <link rel="prefetch" href="/dist/main.js">
-  <link rel="prefetch" href="/dist/thirdparty/thirdparty.min.js">
-}
-
 function setupNewProjectLinks($, analytics) {
   var authenticated = $("#navbar-login").hasClass("signed-in");
   var newProjectButton = $("#new-project-button");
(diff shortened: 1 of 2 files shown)


public/homepage/scripts/main.js
```diff
@@ -32,17 +32,11 @@ require.config({
...
-    brambleHost + "/dist/styles/brackets.min.css",
-    brambleHost + "/dist/bramble.js",
-    brambleHost + "/dist/main.js",
-    brambleHost + "/dist/thirdparty/thirdparty.min.js"
-  ].forEach(function(url) {
-    // Load and cache files as plain text (don't parse) and ignore results.
-    $.ajax({url: url, dataType: "text"});
-  });
+//Using prefetch for easier cacheing and preloading of resources
+  <link rel="prefetch" href="/dist/styles/brackets.min.css">
```
humphd: This is HTML, and you're putting it in a JavaScript file.  You can't mix these.  This would have to go in the HTML for the homepage, see https://github.com/mozilla/thimble.mozilla.org/blob/master/views/homepage/index.html.

Next issue: you need to fix the URLs you are loading, as these are going to try and load from the Thimble server.

gideonthomas:
Note: The purpose of this seems to conflict with: https://github.com/mozilla/thimble.mozilla.org/pull/2013. Are we completely stopping prefetching Bramble or just changing the way we do it?
cc @humphd

humphd:
Changing the way we do it to be more modern, less of a hack.

Mordax:
I realized my HTML in the Javascript mistake right after I pushed the changes - my bad! 

Hopefully this works - the reason I took a while to fix this was I wanted to check if the loading was actually speeding up and I ran into some weird Vagrant error making insanely long Windows path names.

humphd:
I'm finishing this in https://github.com/mozilla/thimble.mozilla.org/pull/2234

PR #2011, opened 2017-04-20
Fix #1917 - remove hard coded files of stay-calm project

### LINKED ISSUE #1917: Get rid of Stay Calm project from the codebase
For some reason the ``stay-calm`` project is hard-coded into our codebase. I don't think we need it there...

image

Is that correct @humphd? If we don't - mind labeling this Good Bug, ty.

ZenanZha:
Regarding to the [Issue1917](https://github.com/mozilla/thimble.mozilla.org/issues/1917), the folder stay-calm needs to be removed from the project because we never use the file anywhere.
@gideonthomas [here](https://github.com/mozilla/thimble.mozilla.org/issues/1917#issuecomment-293433747) and I tested it. (I build the project based on the old version of thimble, not bramble)

Before and After:
capture


flukeout:
@gideonthomas This good to merge? Seems like it's doing the right things and I guess the commits will just get squashed anyway yeah?

gideonthomas:
hmm interesting! Testing this out, thimble hangs on load

humphd:
Interesting.  I remember us having some bug where we had code that looked for files under `default/` on startup (server code) and it was blowing up when it hit a file not of the form it expects.  I can't remember where that is, but I wonder if this same code path is connected to this.

gideonthomas:
I'll try to debug

gideonthomas:
Hmm actually, @humphd would you be able to test this? I think my dev env isn't working. I'm not able to load the gallery on the homepage on master.

humphd:
This isn't hanging for me.  Can I confirm that I just need to start a "Blank Project" and it should just load as we would expect?  It does.

flukeout:
Testing now.

flukeout:
Just tested this and it works for me as well @humphd - can we merge this?

PR #2013, opened 2017-04-21
Fix #1524:  Removing previous patch of mozillathimblelivepreview script preloading in homepage  

### LINKED ISSUE #1524: Remove mozillathimblelivepreview script preloading in homepage
Despite the fact that neither of those things should hold up the app loading and starting up a UI.

evantsyferov:
Working on fixing the second part of issue [1524](https://github.com/mozilla/thimble.mozilla.org/issues/1524)

### DIFF
--- a/public/homepage/scripts/main.js
+++ b/public/homepage/scripts/main.js
@@ -1,5 +1,3 @@
-/* global requirejs */
-
 require.config({
   waitSeconds: 120,
   baseUrl: "/{{ locale }}/homepage/scripts",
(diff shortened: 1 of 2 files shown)


public/homepage/scripts/main.js
```diff
@@ -1,13 +1,3 @@
-/* global requirejs */
```
humphd: This comment can't be removed, it's an instruction to our linters about a global variable.

public/homepage/scripts/main.js
```diff
@@ -1,4 +1,4 @@
...
+// global requirejs
```
humphd: What you had before was right, please undo this change.

evantsyferov: I tried submitting the code with  /* global requirejs */ but Travis fails the push and stays it is an error since requirejs is declared and not used.

gideonthomas: @evantsyferov then remove the entire comment.

humphd:
I'm finishing this in https://github.com/mozilla/thimble.mozilla.org/pull/2234

PR #2029, opened 2017-04-22
[PR]Fixed Issue-2014: added protocol for publish dialog for local dev

bkim49:
Fixed the issue where it did not show hypertext protocol in the publish dialog.

### DIFF
--- a/public/editor/scripts/editor/js/fc/publisher.js
+++ b/public/editor/scripts/editor/js/fc/publisher.js
@@ -273,8 +273,8 @@ define(function(require) {
 
     // Expose the published state with the updated link
     published.link
-      .attr("href", publishUrl)
-      .text(publishUrl);
+      .attr("href", "https://" + publishUrl)
+      .text("https://" + publishUrl);
     published.changed.addClass("hide");
 
     // Re-attach the unpublish handler and remove


public/editor/scripts/editor/js/fc/publisher.js
```diff
@@ -273,8 +273,8 @@ define(function(require) {
 
     // Expose the published state with the updated link
     published.link
-      .attr("href", publishUrl)
-      .text(publishUrl);
+      .attr("href", "https://" + publishUrl)
+      .text("https://" + publishUrl);
```
humphd: Two questions:

* Why does `publishUrl` not already have this?  It seems to me like the right fix is to add it to that string, vs. appending here, which feels brittle, especially if we ever do add it in the future.
* If you *do* fix it this way, I think it's wrong to force `https://` on dev, since it will be running on `http://`.  Why not use the value of `window.location.protocol` to match what we're loading the app with at runtime?

bkim49: @humphd I will see what I can do with window.location.protocol. 
Thank you for your suggestion! 👍

gideonthomas:
@bkim49 I don't think this is the right fix. I think the fix needs to happen in two places:

First, in the https://github.com/mozilla/publish.webmaker.org make the change here: https://github.com/mozilla/publish.webmaker.org/blob/master/env.dist#L28. Change that to `http://localhost:8001`

Second, make the same change here: https://github.com/mozilla/thimble.mozilla.org/blob/master/ecosystem.json#L32

humphd:
I'm going to fix this in publish instead: https://github.com/mozilla/publish.webmaker.org/pull/250.  Closing.

PR #2032, opened 2017-04-22
Add tutorial about how to create a starting kit

ZenanZha:
I just finish a Starting Kit, it was not a easy job...
I think adding these two file (one template, one tutorial) would greatly help others to understand what to do, when they are creating new Starting Kits.

### DIFF
--- a/default/how_to_make_a_starting_kit/How(double click me).html
+++ b/default/how_to_make_a_starting_kit/How(double click me).html
@@ -1,0 +1,329 @@
+<html>
+  <head>
+    <!-- Edit the title here-->
+    <title>Creating Starting Kit Tutorial</title>
+  </head>
+
+  <body>
+    <div class="tutorial">
+      <h1>Creating Starting Kit Tutorial</h1>
+      <p>I just create a Starting Kit, and I believe a good tutorial will help a lot of people to create more Starting Kit much easier.
+      </p>
+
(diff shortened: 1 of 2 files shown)


ZenanZha:
Should I pull the file as .md file instead of .html?

default/how_to_make_a_starting_kit/How(double click me).html
```diff
@@ -0,0 +1,329 @@
...
+          <p class = "tip">
+            The template "tutorial.html" is given in the same folder as this file. :)
+          </p>
+        </li>
+
+        <!-- Step 3 -->
+        <li>
+          <h6>Create a Teaching Kit and Publish it.</h6>
+          <p>
+            This one is opectional.(because I saw a lot of people didn't do that...)
```
Rajat-dhyani: typo error : opectional --> optional

ZenanZha: Fixed, thank you.

humphd:
This probably belongs in our DOCS vs. as an HTML file where you have it, and yes that likely means as a markdown file.  @flukeout could say more.

flukeout:
Agreed with that @humphd - I think this would make a great addition to our Wiki pages. @ZenanZha could you take the content you've created and add it as a wiki page? Thanks!

PR #2048, opened 2017-05-09
The project list now scrolls when it should.

### LINKED ISSUE #2044: My Projects page doesn't scroll, cuts off projects list.
image

Looks like there is a ``overflow: hidden`` rule in editor.css that is being applied to this page also.

flukeout:
To test, load up the ``/projects`` view ("Your Projects") and test that when the browser window is too short to accommodate all of your projects, the page will scroll.

Fixes #2044


humphd:
Tested this locally, works well. r+ from me.

Issue #2043, opened 2017-05-08
[dev] Unable to upload image files to project.


flukeout:
@humphd @gideonthomas can you guys try to see if it works locally for you?

I can't drag files in, nor upload them via the "Upload file" dialog. Getting this error...

![image](https://cloud.githubusercontent.com/assets/25212/25827440/1e655a5a-3400-11e7-8ef7-802f0b9fde13.png)

humphd:
Which browser?

humphd:
Also, can you send me the file you're using that it's rejecting?  I'll test locally.

flukeout:
Chrome and it seems like literally any image file.

flukeout:
![cat-1](https://cloud.githubusercontent.com/assets/25212/25827504/643f201a-3400-11e7-91b8-28e6710920bc.jpg)

This one doesn't work, and is pretty cute so there's that ^

humphd:
I have no issue getting this to work on latest `master`.  What else could be happening here?  Which commit are you on?

humphd:
I'm on e0747c612fc062932f849d998888119ad5a8d025 and every way I try to do this (drag-and-drop to file tree, Upload dialog drag-and-drop, Upload dialog From Computer...) all work.

flukeout:
Okay, thanks for checking. I'll see if I can provide any additional detail.

flukeout:
@humphd I'm running latest master on brackets, just hitting it directly via ``http://localhost:8000/src/hosted``. I get this error in both Firefox and Chrome when I try to drag something in. Can you try latest master?

flukeout:
I get no console errors when it fails, although the console has this after the editor loads

![image](https://cloud.githubusercontent.com/assets/25212/25828306/5a081832-3404-11e7-8496-c2390996ad1f.png)

gideonthomas:
@flukeout `npm run build && npm start` try that

flukeout:
@gideonthomas COME ON MAN what do you take me for over here

flukeout:
That is to say, I tried that and it didn't work. Can you try it locally and let me know? I think I noticed this a few days ago, but it's only become urgent now.

humphd:
I'm able to reproduce this on staging, so maybe it's something that's happening with the built code vs. what's in src:

[...]

PR #2050, opened 2017-05-09
Fix #1930, #2043: Added an alert to prompt the user to refresh Thimble.

### LINKED ISSUE #1930: Add "Updates Available" UX for new service worker cached content
In Brackets, and eventually in Thimble (#1459), we cache and serve static resources so that offline use cases work.  When a service worker has new content, we really should provide the user with some kind of UI that indicates they need to refresh the browser in order to load the new code.  It's a lot like having a "Updates Installed, Restart Required..." UX for a native app.

Depending on what the change is that's been updated in the cache, the update may or may not be critical for the code to continue to work.  An example is when we make changes to `dist/bramble.js` and Thimble needs to call some API function in Bramble that isn't available in the old version.  Running Brackets and Thimble separately means that they can drift out of compatibility until you refresh and force the cached code to get loaded.

Right now we don't do anything, and it worries me.  I'd like to add some kind of UI for the user to tell them that they need to refresh before they continue to use the app.  Once we have this, I can wire up the various bits of Brackets and Thimble to use it.

@flukeout this is probably worth doing before we update prod with all our new changes.

flukeout:
Should look like this if all goes well...

image

Fixes #1930

cc @humphd

### DIFF
--- a/public/editor/scripts/main.js
+++ b/public/editor/scripts/main.js
@@ -53,7 +53,7 @@ require(["jquery", "bowser"], function($, bowser) {
   Bramble.once("error", onError);
 
   $("button.refresh-browser").on("click",function(){
-    window.location.reload();
+    window.location.reload(true);
   });
 
   Bramble.once("updatesAvailable", function() {
--- a/public/editor/scripts/main.js
+++ b/public/editor/scripts/main.js
@@ -38,6 +38,19 @@ require(["jquery", "bowser"], function($, bowser) {
     });
   }
 
+  Bramble.once("updatesAvailable", function() {
+    showRefreshAlert();
+  });
+
+  function showRefreshAlert(){
+    console.log("Thimble has updates - please refresh your browser to get the latest changes.");
+    $("body").addClass("has-alert-bar");
+    $("#serviceworker-warning").removeClass("hide");
+    $("#serviceworker-warning button").on("click",function(){
+      window.location.reload();
+    });
+  }
+
   function onError(err) {
     console.error("[Bramble Error]", err);
     $("#spinner-container").addClass("loading-error");
--- a/public/editor/scripts/main.js
+++ b/public/editor/scripts/main.js
@@ -52,6 +52,10 @@ require(["jquery", "bowser"], function($, bowser) {
 
   Bramble.once("error", onError);
 
+  $("button.refresh-browser").on("click",function(){
+    window.location.reload();
+  });
+
   Bramble.once("updatesAvailable", function() {
     showRefreshAlert();
   });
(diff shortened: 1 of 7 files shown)


public/editor/scripts/main.js
```diff
@@ -38,6 +38,19 @@ require(["jquery", "bowser"], function($, bowser) {
     });
   }
 
+  Bramble.once("updatesAvailable", function() {
```
humphd: Move all this new code down below line 65 where we check to make sure `window.Bramble` exists (it might not if we can't load `bramble.js`

views/editor/warnings.html
```diff
@@ -0,0 +1,17 @@
...
+    <h1>{{ gettext("errorUnsupportedBrowser") }}</h1>
+    <p>
+      {{ gettext("errorUnsupportedBrowserSuggestion") | instantiate | safe }}
+    </p>
+    <div title="{{ gettext("errorUnsupportedBrowserIgnoreTitle") }}" class="let-me-in">{{ gettext("errorUnsupportedBrowserIgnore") }}</div>
+  </div>
+</div>
+
+<div class="top-alert-bar hide" id="serviceworker-warning">
+  {{ gettext("refreshForUpdates") | safe }}
```
humphd: I think you should do a separate string vs. trying to reuse this logic around having that event show this.  If they hit this error screen, they need to try a refresh, basically.  We should give them a button to press that says "Refresh" to download updates and try again.

flukeout:
@humphd Added a Refresh button to the black screen of doom as well. R?

views/editor/bramble.html
```diff
@@ -76,6 +77,36 @@
             -webkit-animation: none;
```
humphd: Not part of this patch, but can you flip the order of this `animation` and `-webkit-animation` so the non-prefixed is first?

public/editor/scripts/main.js
```diff
@@ -52,6 +52,20 @@ require(["jquery", "bowser"], function($, bowser) {
 
   Bramble.once("error", onError);
 
+  $("button.refresh-browser").on("click",function(){
+    window.location.reload();
```
humphd: Make this do a hard-refresh by passing `true` to `reload(true);`

views/editor/bramble.html
```diff
@@ -130,22 +161,19 @@
         <div class="thimble-spinner"></div>
         <div class="loading-message message">{{ gettext("loadingProject") }}</div>
         <div class="error-message message">
-            <p>{{ gettext("errorLoadingProject") | safe }}</p>
+            <!-- <p>{{ gettext("errorLoadingProject") | safe }}</p> -->
```
humphd: What is this comment doing here?  Can we remove?

flukeout:
@humphd - addressed your feedback. R?

PR #2061, opened 2017-05-11
"Won't load" help message on green loading screen.

flukeout:
This patch adds the following message to the loading screen...

It should appear after 10 seconds if the project hasn't loaded. For testing, you can make it appear faster by modifying ``var errorMessageTimeoutMS = 10000;`` [in this file](https://github.com/mozilla/thimble.mozilla.org/compare/master...flukeout:loading-screen-message?expand=1#diff-270f212d36b8a2447fd4f97c986a0b06R55)

cc @humphd

### DIFF
--- a/locales/en-US/messages.properties
+++ b/locales/en-US/messages.properties
@@ -39,6 +39,9 @@ errorUnsupportedBrowserSuggestion=Thimble works in all of the following browsers
 errorUnsupportedBrowserIgnore=Let me try anyway!
 errorLoadingProjectBramble=For error details open your browser's Developer Tools console.
 refreshForUpdates=<strong>Thimble has Updates</strong> - Please refresh your browser to get the latest changes. <button class="refresh-browser">Refresh</button>
+projectNotLoadingMessage=If your project doesn't load, <br>please reload the page to try again.
+tryAgainButtonLabel=Try Again
+emailForHelp=If that doesn't help, <a href="mailto:thimble@mozillafoundation.org">let us know</a>.
 
 # Userbar
 renameProjectSaveBtn=Save
--- a/locales/en-US/messages.properties
+++ b/locales/en-US/messages.properties
@@ -39,7 +39,7 @@ errorUnsupportedBrowserSuggestion=Thimble works in all of the following browsers
 errorUnsupportedBrowserIgnore=Let me try anyway!
 errorLoadingProjectBramble=For error details open your browser's Developer Tools console.
 refreshForUpdates=<strong>Thimble has Updates</strong> - Please refresh your browser to get the latest changes. <button class="refresh-browser">Refresh</button>
-projectNotLoadingMessage=If your project doesn't load, <br>please reload the page to try again.
+projectNotLoadingMessage=Your project seems to be taking a long time to load.<br/> If it doesn't load, please reload the page to try again.
 tryAgainButtonLabel=Try Again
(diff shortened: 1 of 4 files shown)


locales/en-US/messages.properties
```diff
@@ -39,6 +39,9 @@ errorUnsupportedBrowserSuggestion=Thimble works in all of the following browsers
 errorUnsupportedBrowserIgnore=Let me try anyway!
 errorLoadingProjectBramble=For error details open your browser's Developer Tools console.
 refreshForUpdates=<strong>Thimble has Updates</strong> - Please refresh your browser to get the latest changes. <button class="refresh-browser">Refresh</button>
+projectNotLoadingMessage=If your project doesn't load, <br>please reload the page to try again.
```
humphd: What about `Your project seems to be taking a long time to load. If it doesn't load, please reload the page to try again.`

public/editor/scripts/main.js
```diff
@@ -52,6 +52,16 @@ require(["jquery", "bowser"], function($, bowser) {
 
   Bramble.once("error", onError);
 
+  var errorMessageTimeoutMS = 10000;
```
humphd: I think 10s is too short.  I'd do 15s.  We don't want this coming up on slow computers/networks.

public/editor/scripts/main.js
```diff
@@ -52,6 +52,16 @@ require(["jquery", "bowser"], function($, bowser) {
...
+
+  setTimeout(function(){
+    showLoadingErrorMessage();
+  }, errorMessageTimeoutMS);
```
humphd: Can you add analytics here so we record that this happened?  Would be good to know how often we hit it.

public/editor/scripts/main.js
```diff
@@ -52,6 +52,16 @@ require(["jquery", "bowser"], function($, bowser) {
...
+
+  function showLoadingErrorMessage(){
+    $("#spinner-container .taking-too-long").addClass("visible");
+  }
+
   $("button.refresh-browser").on("click",function(){
     window.location.reload(true);
```
humphd: Same thing here, can we add analytics so we get some insight into how often this is happening?

flukeout:
@humphd - made all of the changes you requested, R?

public/editor/scripts/main.js
```diff
@@ -52,8 +52,20 @@ require(["jquery", "bowser"], function($, bowser) {
...
+  }, errorMessageTimeoutMS);
+
+  function showLoadingErrorMessage(){
+    $("#spinner-container .taking-too-long").addClass("visible");
+    analytics.event({ category : analytics.eventCategories.TROUBLESHOOTING, action : "Not loading message shown" });
+  }
+
   $("button.refresh-browser").on("click",function(){
     window.location.reload(true);
+    analytics.event({ category : analytics.eventCategories.TROUBLESHOOTING, action : "Refresh button clicked" });
```
humphd: Move this above the `reload()` call.

flukeout: ha yeah that makes sense

flukeout:
@humphd made all requested changes here - R?

humphd:
This looks good, R+.

On May 11, 2017 4:52 PM, Luke Pacholski <notifications@github.com> wrote:

@humphd<https://github.com/humphd> made all requested changes here - R?

PR #2094, opened 2017-05-14
Homepage nav items should open their localized pages

gideonthomas:
They were always going to `en-US` otherwise, even if you selected a different language on the homepage.


flukeout:
Testing now

flukeout:
@gideonthomas R+

PR #2096, opened 2017-05-14
Fixing MDN and mozilla.org links for all locales

TheoChevalier:
A few explanations on the changes:

Using Thimble in French, `https://developer.mozilla.org/fr/Learn/JavaScript` does not redirect to `https://developer.mozilla.org/fr/Apprendre/JavaScript`, so the link to JavaScript on Thimble is dead when the MDN URL is localized.
Using `https://developer.mozilla.org/docs/Learn/JavaScript` lets MDN properly redirect to `https://developer.mozilla.org/fr/Apprendre/JavaScript` in French.
I’ve also made all MDN links point to the learning area for consistency, instead of only JavaScript.

For MDN and mozilla.org links (and any link to an external website), we should not hardcode the locale using `{{ locale }}`, but let the other website handle the language negotiation directly. We should only hardcode it for internal navigation links, where we know we support the current locale.

Take for instance Tajik (tg) locale, we are generating this link `https://www.mozilla.org/tg/firefox/new/`, which leads to a 404 error since there is no Tajik translation on mozilla.org.

I did a massive search and replace in the locale folder, because I don’t want to ask people to retranslate this string once again. I also updated the string IDs just in case. Those are old strings, so I don’t expect any locale to translate them before this PR is merged, but who knows.


views/editor/warnings.html
```diff
@@ -4,7 +4,7 @@
   <div class="message-box">
     <h1>{{ gettext("errorUnsupportedBrowser") }}</h1>
     <p>
-      {{ gettext("errorUnsupportedBrowserSuggestion") | instantiate | safe }}
+      {{ gettext("errorUnsupportedBrowserSuggestion1") | instantiate | safe }}
```
gideonthomas: since we don't have any more template parameters inside the strings (previously was `{{ locale }}`), we should remove the `instantiate` filter here so that it's just `{{ gettext("errorUnsupportedBrowserSuggestion1") | safe }}`

views/homepage/index.html
```diff
@@ -10,7 +10,7 @@
       <div class="left-side">
         <h1>{{ gettext("thimbleTagLine") | safe }}</h1>
         <p>
-          {{ gettext("thimbleDescription") | instantiate | safe }}
+          {{ gettext("thimbleDescription1") | instantiate | safe }}
```
gideonthomas: same here, remove `instantiate`

TheoChevalier:
Ah, thanks! Fixed

gideonthomas:
Awesome! Thanks for this fix on @TheoChevalier!

PR #2100, opened 2017-05-15
Fix #2099: Wire click handler for Reload Now button before it's needed

### LINKED ISSUE #2099: Loading Error page "Reload Now" button doesn't work
screen shot 2017-05-15 at 11 19 06 am

I'm doing a bunch of testing locally right now, and hit this new error page (I don't have Brackets running locally, so it couldn't load `dist/bramble.js`).  I tried clicking the`Reload Now` button and it doesn't do anything, and there are no error messages on the console.

humphd:
This moves the click handler for the Reload Now button earlier, before it's needed, and before we crash, such that it works when any error case is encountered.

### DIFF
--- a/public/editor/scripts/main.js
+++ b/public/editor/scripts/main.js
@@ -38,6 +38,11 @@ require(["jquery", "bowser", "analytics"], function($, bowser, analytics) {
     });
   }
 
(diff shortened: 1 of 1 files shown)


flukeout:
R+

PR #2145, opened 2017-05-16
Fix :2084 Added Play Pause feature

### LINKED ISSUE #2084: Play/Pause videos in Features page on scroll for better performance
The new Cool Features page (see https://bramble.mofostaging.net/en-US/features) is awesome, and does a good job showing how things work, while still keeping download sizes small.  I've tested it on a bunch of machines/browsers/OSes, and especially on Edge, I notice lag in the framerates.  We have a lot of videos (and therefore media threads) running on this page, even when the `video` elements are visible in the viewport.

Let's see if we can add code to call `play()` and `pause()` when the videos scroll into/out-of view, so they don't run all the time.

Rajat-dhyani:
Fixes #2084

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -51,6 +51,7 @@ module.exports = function( grunt ) {
             "moment": "../node_modules/moment/min/moment-with-locales.min",
             "gallery": "homepage/scripts/gallery",
             "getinvolved": "homepage/scripts/getinvolved",
+            "features": "homepage/scripts/features",
           },
           shim: {
             "jquery": {
--- a/public/homepage/scripts/features.js
+++ b/public/homepage/scripts/features.js
@@ -1,0 +1,20 @@
+define(["jquery"], function($) {
+
+  var features = {
+    init: function(){
+      this.feature = $(".feature");
+
+    	this.feature.mouseenter(function(){
+        $('video', this).get(0).play(); 
+      });
+      	
+      this.feature.mouseleave(function(){
+        $('video', this).get(0).pause(); 
+      });
+  	}
+  };
+  
+  return features;
+});
+
+
--- a/public/homepage/scripts/features.js
+++ b/public/homepage/scripts/features.js
@@ -4,17 +4,15 @@ define(["jquery"], function($) {
     init: function(){
       this.feature = $(".feature");
 
-    	this.feature.mouseenter(function(){
+      this.feature.mouseenter(function(){
         $('video', this).get(0).play(); 
       });
-      	
+
       this.feature.mouseleave(function(){
         $('video', this).get(0).pause(); 
       });
-  	}
+    }
   };
-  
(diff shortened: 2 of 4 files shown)


humphd:
@Rajat-dhyani before we do anything with this PR, we have to fix your git workflow.  You have 12 commits showing in this PR, but only 1 is part of this fix.  Here's how you should work on Thimble code:

* When you want to fix a bug, you start by updating your `master` branch to be the same as what Mozilla has.  Assuming your remote for Mozilla's repo is called `upstream` you do this: `git checkout master && git pull upstream master`.  Your `master` is now at the same place as Mozilla's.
* Now you create a new branch for this bug: `git checkout -b issue-2084`
* Now you do your work, `git add ...`, and `git commit ...` your work.

To fix what you have done here, do the following:

```
git fetch upstream
git checkout -B master upstream/master
git checkout -B issue-2084 master
git cherry-pick 24c2ca63b7a372e63acb4fc2780d2409cd56b50d 
git push origin issue-2084 -f
```

Now you should have a single commit for this work.

You don't want to do merges of `master` with your branches.  When we land your PRs it will mean that your code gets merged, and the next time you do `git pull upstream master` you'll get the latest code.  You should almost never need to merge.

Rajat-dhyani:
I do have a separate branch for this issue but may be issue but this the series i did not know how to remove the series of commits.
Thank you for your help @humphd 👍

Rajat-dhyani:
Did the changes @humphd  it is good to go now ....

humphd:
Great, this looks way better, nice work.

Now, re: how this code should work, what if we do something like this https://codepen.io/gil--/pen/bNxZWg?  I don't think we want to show `controls` on each of these.  @flukeout, thoughts?

Rajat-dhyani:
@humphd i already mentioned the issue  [here](https://github.com/mozilla/thimble.mozilla.org/issues/2084#issuecomment-301819148)

humphd:
Another process tip: once you open a PR for something, move all new discussion to the PR, or it becomes hard to follow the thread.

Rajat-dhyani:
okay sorry for that sir @humphd  👍

flukeout:
Okay, will test this out!

Rajat-dhyani:
@flukeout sir just added play and pause on hover. You can check this one out now.
c.c @humphd

humphd:
We should also support touch/click so these will work on mobile, where we don't have hover events.

Rajat-dhyani:
@humphd done sir :)

flukeout:
Hi @Rajat-dhyani - I've built on your work in a new PR:

* https://github.com/mozilla/thimble.mozilla.org/pull/2197

I'm going to close this one now, but would appreciate you looking over the changes I made.

Cheers and thanks for getting this started! 👍 👍

PR #2148, opened 2017-05-16
Expire session on 500s, only keep cookies for 1 week

humphd:
Changes our `mozillaThimble` cookie to expire after 1 week vs. never, and forces sessions to get destroyed when we hit 500s, in the hopes of avoiding the case where user cookies are blocking a request from proceeding.


humphd:
Follow-up to this in #2150, since our current way of setting `maxAge` on the `cookie` option is wrong.

PR #2151, opened 2017-05-16
Adding reload button to non-404 errors

flukeout:
cc @humphd


flukeout:
cc @humphd OK, this should work now. Please review.

locales/en-US/messages.properties
```diff
@@ -444,6 +444,8 @@ prjListCancelDeleteProjectBtnTitle=Cancel deletion of this project
 # Errors
 errorDetailsHeader=Error Details
 errorSomethingWentWrong=Oops, something went wrong!
+errorNotSureWhatWrong=We're not sure what went wrong, but why don't you...
+errorKeepSeingPage=If you keep seeing this page, please let us know...
```
Pomax: `errorKeepSeeingPage`?

views/error.html
```diff
@@ -31,8 +30,16 @@
...
-            {{ gettext("errorSomethingWentWrongDetails") }}
+            {{ gettext("errorNotSureWhatWrong") }}
           </h2>
+          <p class="button-wrapper">
+            <button class="try-again action-button">
+              {{ gettext("tryAgainButtonLabel") }}
+            </button>
+          </p>
+          <p>
+            {{ gettext("errorKeepSeingPage") }}
```
Pomax: `errorKeepSeeingPage`?

public/error/scripts/main.js
```diff
@@ -0,0 +1,7 @@
+var tryAgainButton = document.querySelector(".try-again") || null;
```
humphd: You don't need `|| null;` since it will return `null` if it doesn't find the element.

server/lib/http-error.js
```diff
@@ -21,7 +21,8 @@ class HttpError {
     const userFriendlyError = {
       statusMessage,
       status: statusCode,
-      message: error.userMessage || error
+      message: error.userMessage || error,
+      localeInfo: localeInfo
```
humphd: You can also just do `localInfo` like line 22 if you want, but this is fine too.

views/error.html
```diff
@@ -60,6 +67,10 @@
       </div>
     </section>
 
+      <script id="homepage-script"
+          src="/scripts/vendor/require.min.js"
+          data-main="/{{ localeInfo.locale }}/error/scripts/main.js">
+      </script>
```
humphd: You aren't loading a `requirejs` module, so I think this can just be:

```
<script src="/{{ localeInfo.locale }}/error/scripts/main.js"></script>
```

views/error.html
flukeout: Yea, fixing!

locales/en-US/messages.properties
flukeout: Oui!

flukeout:
Fixed things up @humphd - R?

humphd:
@cadecairos, @gideonthomas can we prod push this fix too?

cadecairos:
@Pomax pushed this to production 45 minutes ago, so people should be seeing it now.

PR #2159, opened 2017-05-17
Pass all necessary variables to error.html template for 404s

humphd:
I notice that we aren't passing all the necessary variables to the `error.html` template when we show a 404 error vs. what we do on 500s.

### DIFF
--- a/server/lib/http-error.js
+++ b/server/lib/http-error.js
@@ -2,28 +2,34 @@
 
 const Nunjucks = require("nunjucks");
 
+function toUserFriendlyError(error, request, statusCode) {
+  //jshint unused:vars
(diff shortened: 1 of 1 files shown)


flukeout:
Will test this out now.

flukeout:
What's a good way to test this @humphd - it doesn't look like we're making use of the data anywhere on the page.

humphd:
Locale info wasn't being passed, for example.

flukeout:
Re-testing now

flukeout:
Okay, locale is being passed now, so that works. Good to merge @humphd

humphd:
@cadecairos this would be good to have on prod, too.

PR #2167, opened 2017-05-17
Temporary token assert code to figure out what's breaking on prod

humphd:
This adds the ability to assert that a user's session token exists, is a string, and is of the proper form.  If those aren't true, we log it.  I am not logging the actual token, only the characters that are outside what we expect.

cc @cadecairos, @Pomax

### DIFF
--- a/server/lib/middleware.js
+++ b/server/lib/middleware.js
@@ -85,16 +85,18 @@ module.exports = function middlewareConstructor(config) {
       let token = req.user.token = cryptr.decrypt(req.session.token);
 
       if(ASSERT_TOKEN) {
-        // We expect a token of the form '16d4c4d2fa4d6e4aa6b1b5c6a115ad44fd271dd98204f2008bf5efbba5a56dec'
+        // We expect a 64 character HEX string for the token, for example:
+        // '16d4c4d2fa4d6e4aa6b1b5c6a115ad44fd271dd98204f2008bf5efbba5a56dec'
         let tokenType = typeof token;
         if(!token) {
           console.log("ASSERT_TOKEN FAILED: Expected token to exist");
         } else {
           if(tokenType !== "string") {
             console.log("ASSERT_TOKEN FAILED: Expected token type to be String, instead got: " + tokenType);
-          }
-          if(!/^[a-z0-9]+$/.test(token)) {
-            console.log("ASSERT_TOKEN FAILED: Expected token to only have chars a-z, 0-9. Also got: '" +  token.replace(/[a-z0-9]/g, ' ') + "'");
+          } else {
+            if(!/^[a-z0-9]{64}$/.test(token)) {
+              console.log("ASSERT_TOKEN FAILED: Expected token to only have chars a-z, 0-9. Also got: '" +  token.replace(/[a-z0-9]/g, ' ') + "'");
+            }
           }
         }
       }
--- a/server/lib/middleware.js
+++ b/server/lib/middleware.js
@@ -87,9 +87,16 @@ module.exports = function middlewareConstructor(config) {
       if(ASSERT_TOKEN) {
         // We expect a token of the form '16d4c4d2fa4d6e4aa6b1b5c6a115ad44fd271dd98204f2008bf5efbba5a56dec'
         let tokenType = typeof token;
-        console.assert(token, "Expected token to exist");
-        console.assert(tokenType === "string", "Expected token type to be String, got %s", tokenType);
-        console.assert(/^[a-z0-9]+$/.test(token), "Expected token to only have chars a-z, 0-9. Also got %s", token.replace(/[a-z0-9]/g, ' '));
+        if(!token) {
+          console.log("ASSERT_TOKEN FAILED: Expected token to exist");
+        } else {
+          if(tokenType !== "string") {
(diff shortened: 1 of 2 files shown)


server/lib/middleware.js
```diff
@@ -79,7 +82,15 @@ module.exports = function middlewareConstructor(config) {
...
       req.user = req.session.user;
-      req.user.token = cryptr.decrypt(req.session.token);
+      let token = req.user.token = cryptr.decrypt(req.session.token);
+
+      if(ASSERT_TOKEN) {
+        // We expect a token of the form '16d4c4d2fa4d6e4aa6b1b5c6a115ad44fd271dd98204f2008bf5efbba5a56dec'
+        let tokenType = typeof token;
+        console.assert(token, "Expected token to exist");
+        console.assert(tokenType === "string", "Expected token type to be String, got %s", tokenType);
+        console.assert(/^[a-z0-9]+$/.test(token), "Expected token to only have chars a-z, 0-9. Also got %s", token.replace(/[a-z0-9]/g, ' '));
```
cadecairos: this will throw on `token.replace` in the event token isn't a String.

for example:
```
let token = {};
console.assert(/^[a-z0-9]+$/.test(token), "Expected token to only have chars a-z, 0-9. Also got %s", token.replace(/[a-z0-9]/g, ' '));
// TypeError: token.replace is not a function
```

humphd:
Moved away from `console.assert` since it throws an unhandled exception.  Also added a null check around the replace.

I hard-coded a bad token, and my logs looked like this (note the fact that it goes across 2 lines, because I included a `\n`:

```
timestamp=2017-05-17-19:20:32+0000 app=thimble id=4 type=out message=ASSERT_TOKEN FAILED: Expected token to only have chars a-z, 0-9. Also got: '
timestamp=2017-05-17-19:20:32+0000 app=thimble id=4 type=out message=       !'
```

server/lib/middleware.js
```diff
@@ -79,7 +82,22 @@ module.exports = function middlewareConstructor(config) {
 
       // Decrypt oauth token
       req.user = req.session.user;
-      req.user.token = cryptr.decrypt(req.session.token);
+      let token = req.user.token = cryptr.decrypt(req.session.token);
+
+      if(ASSERT_TOKEN) {
+        // We expect a token of the form '16d4c4d2fa4d6e4aa6b1b5c6a115ad44fd271dd98204f2008bf5efbba5a56dec'
```
Pomax: can we change this to "we expect an X character long, hexadecimal character string"? And directly related to that, if we know the exact length we expect, we can assert on token length as well

server/lib/middleware.js
```diff
@@ -79,7 +82,22 @@ module.exports = function middlewareConstructor(config) {
 
       // Decrypt oauth token
       req.user = req.session.user;
-      req.user.token = cryptr.decrypt(req.session.token);
+      let token = req.user.token = cryptr.decrypt(req.session.token);
```
Pomax: can we change this to "we expect an X character long, hexadecimal character string"? And directly related to that, if we know the exact length we expect, we can assert on token length as well

server/lib/middleware.js
```diff
@@ -79,7 +82,22 @@ module.exports = function middlewareConstructor(config) {
...
+        if(!token) {
+          console.log("ASSERT_TOKEN FAILED: Expected token to exist");
+        } else {
+          if(tokenType !== "string") {
+            console.log("ASSERT_TOKEN FAILED: Expected token type to be String, instead got: " + tokenType);
+          }
+          if(!/^[a-z0-9]+$/.test(token)) {
+            console.log("ASSERT_TOKEN FAILED: Expected token to only have chars a-z, 0-9. Also got: '" +  token.replace(/[a-z0-9]/g, ' ') + "'");
+          }
+        }
```
Pomax: we should be able to test for token length here

server/lib/middleware.js
```diff
@@ -79,7 +82,22 @@ module.exports = function middlewareConstructor(config) {
...
+        // We expect a token of the form '16d4c4d2fa4d6e4aa6b1b5c6a115ad44fd271dd98204f2008bf5efbba5a56dec'
+        let tokenType = typeof token;
+        if(!token) {
+          console.log("ASSERT_TOKEN FAILED: Expected token to exist");
+        } else {
+          if(tokenType !== "string") {
+            console.log("ASSERT_TOKEN FAILED: Expected token type to be String, instead got: " + tokenType);
+          }
+          if(!/^[a-z0-9]+$/.test(token)) {
+            console.log("ASSERT_TOKEN FAILED: Expected token to only have chars a-z, 0-9. Also got: '" +  token.replace(/[a-z0-9]/g, ' ') + "'");
```
cadecairos: this will still throw an exception if token.replace isn't a function. should this be an `else if` block?

humphd:
I think I've fixed everything you two want.  See if you agree

humphd:
@cadecairos what do you want to do about putting this out on staging + prod, since it also requires the env to be updated with `exports ASSERT_TOKEN=true` in order to run?

cadecairos:
I can update the environment on heroku from my command line

humphd:
OK, this is in.  @cadecairos please try this and see what you can get from the logs on prod.

PR #2168, opened 2017-05-17
Adjusted the github issue count logic

### LINKED ISSUE #2135: get involved -> good first bug count is wrong?
I don't know if this value is hardcoded atm, but it's showing me 30 good first bugs, when in actuality there are 49 (as I write this issue)

flukeout:
Since the API returns a max of 30 results. I've changed it to say "30+" when it equals 30 or more. This will be wrong in the case of there being 30 issues, but otherwise OK.

Fixes #2135

### DIFF
--- a/public/homepage/scripts/getinvolved.js
+++ b/public/homepage/scripts/getinvolved.js
@@ -24,7 +24,7 @@ define(["jquery"], function($) {
 
       if(issueCount > 1) {
         if(issueCount >= 30) {
-          issueCount = issueCount + "+";
+          issueCount = "30+";
         }
         this.issueCountEl.text(issueCount);
         this.issueCountEl.show();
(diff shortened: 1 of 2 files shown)


flukeout:
cc @TheoChevalier - what do you think about hardcoding a "30+" value for an indicator showing that there are more than 30 items?

image

TheoChevalier:
@flukeout I would say it’s not optimal, but okay-ish in this case. In a not so different context, "+" is not kept everywhere: https://transvision.mozfr.org/string/?entity=mail/chrome/messenger/multimessageview.properties:atLeastNumIgnored&repo=release#en-US

But at the same time, there’s not that many alternative options here, given the character constraint. Maybe ">30" or "+30"

flukeout:
Okay, I'll merge and update later if things seem weird.

PR #2170, opened 2017-05-17
test cryptr fallback  with eas-256

Pomax:
This adds a fallback instance of Cryptr, using the 1.0.0 `eas-256` algorithm

### DIFF
--- a/server/lib/middleware.js
+++ b/server/lib/middleware.js
@@ -21,6 +21,7 @@ const ASSERT_TOKEN = env.get("ASSERT_TOKEN");
 
 module.exports = function middlewareConstructor(config) {
   let cryptr = new Cryptr(env.get("SESSION_SECRET"));
+  let cryptrFallback = new Cryptr(env.get("SESSION_SECRET"), "aes-256");
 
   return {
     /**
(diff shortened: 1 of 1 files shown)


server/lib/middleware.js
```diff
@@ -21,6 +21,7 @@ const ASSERT_TOKEN = env.get("ASSERT_TOKEN");
 
 module.exports = function middlewareConstructor(config) {
   let cryptr = new Cryptr(env.get("SESSION_SECRET"));
+  let cryptrFallback = new Cryptr(env.get("SESSION_SECRET"), "aes-256");
```
cadecairos: `'aes256'`

server/lib/middleware.js
```diff
@@ -85,19 +86,35 @@ module.exports = function middlewareConstructor(config) {
...
+          if(!/^[a-z0-9]{64}$/.test(token)) {
+            console.log("ASSERT_TOKEN FAILED: Expected token to only have chars a-z, 0-9. Also got: '" +  token.replace(/[a-z0-9]/g, ' ') + "'");
+            return false;
+          }
+          
+          return true;
+        }
+
+        if (!assert(token)) {
+          token = req.user.token = cryptrFallback.decrypt(req.session.token);
```
cadecairos: we should set `req.session.token` to the result too.

server/lib/middleware.js
```diff
@@ -85,19 +86,36 @@ module.exports = function middlewareConstructor(config) {
...
+          if(!/^[a-z0-9]{64}$/.test(token)) {
+            console.log("ASSERT_TOKEN FAILED: Expected token to only have chars a-z, 0-9. Also got: '" +  token.replace(/[a-z0-9]/g, ' ') + "'");
+            return false;
+          }
+          
+          return true;
+        }
+
+        if (!assert(token)) {
+          console.log("ASSERT_TOKEN FAILED: retrying decryption using eas-256 rather than eas-256-ctr");
```
cadecairos: typo: 'aes'

server/lib/middleware.js
```diff
@@ -85,19 +86,36 @@ module.exports = function middlewareConstructor(config) {
...
+          token = req.user.token = cryptrFallback.decrypt(req.session.token);
+          assert(token);
         }
```
humphd: At this point, if we still don't have a valid token, I think we should consider calling `next()` with an error that indicates that the user's session has expired, and they need to login again.  This will trigger our `error.html` template to render, and at least give some clue that there is something wrong.

server/lib/middleware.js
humphd: Let's add a comment here with a link to https://github.com/MauriceButler/cryptr/compare/fabae97a61119d69f03fc189f7c95dda826c96b7...master#diff-168726dbe96b3ce427e7fedce31bb0bcR9 so it's clear why we're doing this.

server/lib/middleware.js
```diff
@@ -85,19 +86,36 @@ module.exports = function middlewareConstructor(config) {
...
+            return false;
           }
+          
+          if(!/^[a-z0-9]{64}$/.test(token)) {
+            console.log("ASSERT_TOKEN FAILED: Expected token to only have chars a-z, 0-9. Also got: '" +  token.replace(/[a-z0-9]/g, ' ') + "'");
+            return false;
+          }
+          
+          return true;
+        }
```
cadecairos: semicolon plz

server/lib/middleware.js
```diff
@@ -85,18 +90,37 @@ module.exports = function middlewareConstructor(config) {
...
+          }
+          
+          return true;
+        };
+
+        if (!assert(token)) {
+          console.log("ASSERT_TOKEN FAILED: retrying decryption using aes-256 rather than aes-256-ctr");
+          token = req.user.token = cryptrFallback.decrypt(req.session.token);
+          if (!assert(token)) {
+            return next(new Error("Session token cannot be decrypted"));
```
humphd: Can you update this string to say, `"Session token cannot be decrypted. Please sign out and sign in again."` so that there is some info on what to do?

gideonthomas: you probably want to use https://github.com/mozilla/thimble.mozilla.org/blob/master/server/lib/http-error.js#L65-L74 instead of `new Error` so that it logs and displays correctly.

Issue #2162, opened 2017-05-17
Consider moving Export Zip to the bottom of the file tree.


flukeout:
It's a cool feature, and doesn't realllly belong in the menu that we have added it to. We could move it out of the menu and add an export/download button at the bottom of the file tree.

PR #2174, opened 2017-05-18
Duplicate Export to .zip feature to avoid confusing users, fixes #2115

### LINKED ISSUE #2115: QA > File menu doesn't close after clicking "Download Files"
On Mac Firefox, this here menu stays open after clicking download project.

image

humphd:
This undoes some of what we did in https://github.com/mozilla/thimble.mozilla.org/commit/adf9a5f79f7828e2b025cc673c412dff823d2928 in order to duplicate the "Export .zip" feature, since users are confused by us moving it, and thinking we killed it off.  I think it's better to have it in both places.  I've also fixed the bug where we don't close the popup menu when we download the zip #2115.

Later in #2162 we might do something different again.  This is a stop-gap until then.

### DIFF
--- a/public/editor/scripts/editor/js/fc/bramble-menus.js
+++ b/public/editor/scripts/editor/js/fc/bramble-menus.js
@@ -315,6 +315,7 @@ define(function(require) {
     var $addJs = $("#filetree-pane-nav-add-js");
     var $addUpload = $("#filetree-pane-nav-add-upload");
     var $addTutorial = $("#filetree-pane-nav-add-tutorial");
+    var $downloadZip = $("#filetree-pane-nav-export-project-zip");
 
     // Add File button and popup menu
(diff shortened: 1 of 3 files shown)


flukeout:
Testing now.

flukeout:
I don't think we have the ``exportProjectLink`` string in Thimble anymore. It's moved to Brackets. I'm not able to get the translation to show up after running the localization scripts in any case...

image

humphd:
How is that possible, since the menu we moved it to is in Thimble?  I just used the same string here.

flukeout:
Oh damn yeah you are right, I got totally confused. Maybe the key is wrong, let me double check.

flukeout:
It's called ``downloadProjectLink`` now

humphd:
Good catch, pushed a new commit. r?

flukeout:
Okay, label is now localized and the feature works. Good to Merge @humphd

PR #2175, opened 2017-05-18
Sets target as self for userbar links, Fixes #2123

### LINKED ISSUE #2123: QA > Top nav links in Your Projects page shouldn't spawn a new tab.
image

We should just open them in the same tab.

hkirat:
Fixes #2123

### DIFF
--- a/views/editor/userbar.html
+++ b/views/editor/userbar.html
@@ -30,9 +30,9 @@
           </div>
           <div class="dropdown-content">
             <ul>
-              <li><a href="https://github.com/mozilla/thimble.mozilla.org/wiki/Thimble-Troubleshooting">{{ gettext("helpTroubleshooting") }}</a></li>
-              <li><a href="https://github.com/mozilla/thimble.mozilla.org/wiki/Using-Thimble-FAQ">{{ gettext("helpFAQ") }}</a></li>
-              <li><a href="https://github.com/mozilla/thimble.mozilla.org/wiki/About-Thimble">{{ gettext("helpFeatures") }}</a></li>
+              <li><a target = "_blank" href="https://github.com/mozilla/thimble.mozilla.org/wiki/Thimble-Troubleshooting">{{ gettext("helpTroubleshooting") }}</a></li>
+              <li><a target = "_blank" href="https://github.com/mozilla/thimble.mozilla.org/wiki/Using-Thimble-FAQ">{{ gettext("helpFAQ") }}</a></li>
+              <li><a target = "_blank" href="https://github.com/mozilla/thimble.mozilla.org/wiki/About-Thimble">{{ gettext("helpFeatures") }}</a></li>
             </ul>
           </div>
         </div>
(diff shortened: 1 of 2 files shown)


flukeout:
Testing now.

Pomax:
This kind of feels like the wrong fix: the reason these are opening as new tabs is because https://thimble.mozilla.org/en-US/projects/ has a `<base target="_blank">` in the head element, which we should probably remove.

flukeout:
Oh snap good catch @Pomax, removing that seems like the better fix. Would you mind changing the PR to try this new method @hkirat - it seems cleaner.

hkirat:
Yes that is right @Pomax. It looks like it inherits this from layout.html [here](https://github.com/mozilla/thimble.mozilla.org/blob/master/views/editor/layout.html#L13).
 
I think this might be related to a problem in the structure of the views folder that I mentioned earlier [here](https://github.com/mozilla/thimble.mozilla.org/issues/2044#issuecomment-300135021).

I think target has been set to blank in layout.html for a reason (probably links are expected to open in a new tab when you are in editor more). layout.html is used by other html files as well so removing this line might change the way links respond on some other page (editor page specifically)

Would overriding this line be a good fix?

flukeout:
I'm guessing that the ``_blank`` is there in case someone clicks either ``Help`` or the Logo in which case we didn't want them to leave the editor. All of the other clickable stuff in the header doesn't really care about ``_blank``. 

My bet is that it's safe to remove the line @Pomax pointed out, and then just assign ``_blank`` to the things we know we want, like the help link & logo.

hkirat:
Hey
sorry for slacking this one.
Was going through the code.
If we remove the line that sets target  to _blank globally, we wiill have to modify [this](https://github.com/mozilla/thimble.mozilla.org/blob/master/locales/en-US/messages.properties#L60) line for each language. 
I think that was the reason that this line was added in the first place. 
Would that be the right way to proceed @flukeout?

hkirat:
Hey sorry got a little confused on this one.
Did the final changes.

views/editor/userbar.html
```diff
@@ -4,7 +4,7 @@
 
     <div id="navbar-left" class="container-flex">
       <div class="navbar-logo-container small-logo">
-        <a href="/{{ locale }}" class="logo" title="{{ gettext("backToHomepage") }}">
+        <a target="_blank" href="/{{ locale }}" class="logo" title="{{ gettext("backToHomepage") }}">
```
Pomax: I don't think we even need a target here... what would be the value of opening in a new tab/window for the logo? @flukeout?

flukeout:
Testing now.

flukeout:
Hey @hkirat - I'd like to request some changes on this PR.

* First, can you rebase it on the latest master branch?
* Remove the ``_blank`` from the logo link
* Add ``_blank`` to the items in the new Help dropdown (you'll see it after rebasing)

Then I things will feel right. We open new tabs for the help items, since someone might want help on what they're working on, and for everything else, we load it in the same tab.

Sound good?

hkirat:
ping @flukeout

flukeout:
Testing now - sorry for the delay @hkirat

flukeout:
Great, works. Thanks @hkirat

PR #2178, opened 2017-05-18
Return proper object type for bramble.getAutoCloseTags() shim, don't crash

humphd:
Fix for crash reported by user:

image

I was returning the wrong type in the BrambleShim code.  The real answer is for the user to update to the new Bramble API, but this will stop us from crashing, and let the "Reload Now" flow work.


flukeout:
Testing now!

flukeout:
R+ let's do it

humphd:
This needs to land on prod, cc @cadecairos

PR #2182, opened 2017-05-18
Broken links in footer for some locales

TheoChevalier:
Follow up of #2096, missed those links in the footer


flukeout:
Thanks @TheoChevalier - testing now.

flukeout:
Okay yeah works, I guess we don't want to assume that each of the locales we have work on other sites.

flukeout:
Thanks a lot @TheoChevalier 👍

PR #2198, opened 2017-05-23
The Black Screen of Doom now fires off a Google Analytics event we can keep track of.

### LINKED ISSUE #2190: Add unique Google Analytics event for Black Screen of Death.
I think we only track the Green Screen of Death via analytics, would be good to check in on this one too.

flukeout:
Also updated the Green Screen error description.

Fixes #2190

### DIFF
--- a/public/editor/scripts/main.js
+++ b/public/editor/scripts/main.js
@@ -47,7 +47,7 @@ require(["jquery", "bowser", "analytics"], function($, bowser, analytics) {
   function onError(err) {
     console.error("[Bramble Error]", err);
     $("#spinner-container").addClass("loading-error");
-    analytics.event({ category : analytics.eventCategories.TROUBLESHOOTING, action : "Editor loading error (black Screen)" });
+    analytics.event({ category : analytics.eventCategories.TROUBLESHOOTING, action : "Editor loading error (Black Screen)" });
   }
 
   // If Bramble fails to load (some browser loading issues cause it to fail),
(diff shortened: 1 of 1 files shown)


public/editor/scripts/main.js
```diff
@@ -47,6 +47,7 @@ require(["jquery", "bowser", "analytics"], function($, bowser, analytics) {
   function onError(err) {
     console.error("[Bramble Error]", err);
     $("#spinner-container").addClass("loading-error");
+    analytics.event({ category : analytics.eventCategories.TROUBLESHOOTING, action : "Editor loading error (black Screen)" });
```
humphd: To be consistent, let's change to `Black screen` (capital `B`) since you do `Green screen` below.

Also, what do you think about including the `err` in your action?  @gideonthomas is this a good or bad idea?  It would at least give us a fast look at what's happening

flukeout: I thought about that as well. Is it always a string, or is it some kind of object? I could maybe stringify it and send the first 150 chars or something. I'll see if there are any limits.

humphd: You could try `err.toString()` and just see how it goes.  The worst thing that will happen is we'll crash, but the editor is already crashed, so should be fine.

flukeout:
@humphd Updated the capitalization and also logging the error copy in the event. The GA limit is 500 chars and it will automatically be truncated if it's longer, so I think we're OK. R?

PR #2200, opened 2017-05-24
Fix #2072 - make it possible to add new folders from Thimble

### LINKED ISSUE #2072: Make it possible to add a new Folder from the File Tree page icon
Via Twitter, a request to add the ability to create new Folders in addition to files from the File Tree page icon.  I'll add a patch to Brackets to make this possible, then do the UI here.

humphd:
This adds a new menu option above the file tree to enable adding folders.

I also wrote https://github.com/mozilla/brackets/pull/767 on the Brackets side to fix an issue in the naming of our commands, and we should land that first.


flukeout:
Testing now.

PR #2204, opened 2017-05-24
Expand GA analytics to include timing and exception tracking

### LINKED ISSUE #2212: Project Loading Slowly GA event fires even though a project loaded quickly.
Oh boy. I was reviewing @humphd's analytics PR and logging all the GA calls and lo and behold I saw this thing pop up in the console even though my project loaded just fine (in less than 5 seconds). 

image

Tested this in Chrome and FF. Happens on both signed out and signed in projects.

humphd:
I notice that we can include timing and exception analytics in GA along with event categories.  I've expanded the API and added some new tracking code to help us see what's happening.

Fixes #2212

### DIFF
--- a/public/editor/scripts/analytics.js
+++ b/public/editor/scripts/analytics.js
@@ -20,6 +20,11 @@
   eventCategories.PROJECT_ACTIONS = "Project Actions";
   eventCategories.TROUBLESHOOTING = "Troubleshooting";
 
+  // Strings for timing category names
+  var timingCategories = {};
+  timingCategories.BRAMBLE = "Bramble";
+  timingCategories.THIMBLE = "Thimble";
+
   var _redacted = "REDACTED (Potential Email Address)";
 
   /**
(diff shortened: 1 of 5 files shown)


public/editor/scripts/analytics.js
```diff
@@ -65,14 +70,71 @@
...
+      } else {
+        warn("Expected `value` arg to be a Number.");
+        return;
+      }
+    }
+
+    if(options.label) {
+      eventOptions.timingLabel = trim(options.label);
+    }
+  }
```
flukeout: The ``timing()`` function ends without ever calling ``gaSend``.

flukeout:
The documentation for exception shows the event structure as...

image

Yet when I console, log out the event object you're creating it's...

```
Object {
  hitType: "exception", 
  exDescription: "Unable to load Bramble editor in this browser", 
  exFatal: true
}
```

Will this alternate way work, or do we need to structure it like the docs say?

flukeout:
Also thanks for leaving links to the docs in the comments, that's a handy practice.

humphd:
OK, I've fixed the issues you found (you can indeed call the functions by passing the object vs. expanding the args out--both ways work).  I've also folded your timeout changes into this, and refactored our startup so we can cancel the timeout later. @gideonthomas you might want to take a look at this, too.

@flukeout can you test again, please?

flukeout:
Testing again now.

flukeout:
Just tested this and exceptions, timings and events seem to work well. The only difference I noticed is that you removed the ``eventLabel`` of the Black Screen troubleshooting event. We used to toString() the ``err`` and pass it along. Other than that looks good on my end.

flukeout:
cc @humphd

public/editor/scripts/analytics.js
```diff
@@ -65,14 +70,71 @@
...
+        return;
+      }
+      // Force value to int
+      eventOptions["timingValue"] = options.value|0;
+    } else {
+      // If now value is given, assume we want to try to measure time since page load
+      if(window.performance) {
+        eventOptions["timingValue"] = Math.round(window.performance.now());
+      } else {
+        warn("Expected `value` arg to be a Number.");
```
gideonthomas: wrong message?

gideonthomas:
This also fixes https://github.com/mozilla/thimble.mozilla.org/issues/2212

Issue #1667, opened 2017-01-24
Saving indicator is not showing, even though changes are saved.


flukeout:
In Chrome, when editing a project, changes are saved even though the Saving indicator never shows up...

![saving](https://cloud.githubusercontent.com/assets/25212/22270926/e519eeec-e246-11e6-8966-a21cd5f93159.gif)

This change survived refreshing the browser, and also closing the tab and re-opening the project.

humphd:
The issue here is that we have two kinds of saving:
* saving the files to "disk" in IndexedDB in the browser so they will be available locally and survive a refresh
* sending the files to the server to be saved there.  You have to be logged in for this to happen.

I'm not sure how much we want to expose the user to this difference.

flukeout:
I'm not seeing a "Saving.." indicator in either of the cases above, one of which I'm guessing sends the files to the server as I'm closing the browser tab and opening the project from the project list.

When do we currently attempt to show the "Saving..." indicator? I was under the impression it was only for server saves.

Taguiar:
I wold like to fix this bug

flukeout:
Sounds good @Taguiar

gideonthomas:
@flukeout yes that indicator should only show up if it saves to the server.

I am hesitant to leave this as a "good first bug" since this might involve touching/investigating complex code in our file syncing logic. @humphd what do you think?

Pomax:
As long as the code path is easy to trace and we're available for questions, I don't think this is too complicated a bug to work on. I'd be in favour of the mantra "see if it's too much work, switch if it is": if more than a day is required to get *into* the bug, then this is a bad bug, but ruling it out before someone starts to look into it is probably acting too early... someone might be really good at bug hunting and love this challenge!

humphd:
The code for this (at least the code that shows/hides things) is here:

https://github.com/mozilla/thimble.mozilla.org/blob/fa3dde088dc016a562520e62a9df1f5558ba8f70/public/editor/scripts/editor/js/fc/filesystem-sync.js#L37

The code that's triggering those events is here:

https://github.com/mozilla/thimble.mozilla.org/blob/fa3dde088dc016a562520e62a9df1f5558ba8f70/public/editor/scripts/editor/js/fc/sync-manager.js#L468-L484

Taguiar:
I found the bug really fast, but the problem I am having is thaT I can log in from my local computer.
What you think @humphd

Taguiar:
[...]

PR #2209, opened 2017-05-25
Fix #1909, #1667: Add GUI display when saving logic is triggered

### LINKED ISSUE #1909: Prompt user when file gets saved to the browser

humphd:
Builds on work started in https://github.com/mozilla/thimble.mozilla.org/pull/1932 by @Taguiar.  This adds saving indicators for users when they aren't authenticated, and also enhances what we show for authenticated users.

If you are anonymous, it will look like this:

ofline-saving-indicators

If you are logged in, it looks like this (notice that it lets you know about offline and server saving):

auth-saving-indicators


flukeout:
Cool, some questions @humphd...

* Looks like after making a change in my code, there is about a 5 second timeout before saving those files locally, right?
  * Something about this feels odd to me. I know we can save basically instantly so it's weird to see "Saving..." when I know it's just hanging out and waiting to make a save.
  * It just makes it feel slow and like I want to wait for it to finish or something, needless anticipation.
  * I'll think on some other label or solution, but curious what you think?
* ``All files saved offline`` to me suggests my files are somehow available when I'm "offline" which isn't really the case.
  * We're really talking about all the files being saved locally here, right?
  * I almost think we don't need this label / state indicator at all. It seems like it would be pretty rare that you'll refresh the browser within 5 seconds of a change, and that's all we're really trying to prevent, right?

Will give this a bit more thought and chime in again shortly.

humphd:
Yup, let's perfect this.  First, consider the state of the art:

Google Docs:

dirty-saved

Medium's editor:

medium-example

The `Saving...` I have is really `There are unsaved changes, I'm going to save them soon...`.  That's what Medium's `Unsaved Changes` means; but Google does what I did, `Saving...`.  I'm not sure what's right.  I agree that we should make it feel fast.

`All files saved offline` is going to mean exactly that, once I land the last service worker patch: you can refresh and you'll have them in *this* browser.  Maybe there is better wording?  I'm trying to say:

* offline - saved to indexeddb in this browser, so you can safely close this tab or refresh
* saved to server - synced to Thimble, so you can move to another computer, login, and keep working

Maybe we don't need to indicate both, but I think it helps people understand what is going on?

The reason I moved to always show this was because Google Docs does, so it seemed "normal" to do so.

flukeout:
Yeah, totally understand the points you're making. 

**When signed out**
* Here we only do local saving, and I think the 5 seconds feels slow. Looks like gdocs does every 2 seconds. Can we try that and indicate ``All files saved to browser`` once ``Saving...`` is done?

**When signed in**
* Will it hurt us to do server saving every 2 seconds?
* Additionally, it looks like we fire the event 5 seconds after the first change, even if other changes are made after. Can we defer the save until 2 seconds after the last changes has been made given some reasonable time we wait for between keystrokes?
  * That could reduce our overall save count, because now we're going every 5 seconds no matter what if someone is typing, but we could go just once after 20 seconds of continuous typing.

Pomax:
Is this still a PR we can expect to see landing, or should we close this?

Taguiar:
I thought you have already accepted it. I will continue for sure.

On Jan 18, 2018 1:45 PM, Mike Kamermans <notifications@github.com> wrote:

Is this still a PR we can expect to see landing, or should we close this?

Pomax:
@humphd was this passed by in another PR and can it be closed?

humphd:
Yes, it can be closed.

PR #2211, opened 2017-05-25
Fix #2163 - Missing locale variable breaks links on 404 page

### LINKED ISSUE #2163: 404 Page > Can't navigate back to homepage via clicking Logo
It just reloads the 404 page again because it tries to add the locale to the URL twice...

image

The logo is linked as ``<a href="/{{ locale }}" class="logo">`` - Should we just remove the locale from here and let the browser re-assign it on the next page load?

gideonthomas:
This should fix the logo link and also cleans up the code a bit.


flukeout:
Testing

PR #2215, opened 2017-05-26
Removes false positive "slow project loading" google event.

### LINKED ISSUE #2212: Project Loading Slowly GA event fires even though a project loaded quickly.

flukeout:
We gave a project 15 seconds to load before showing an error message, but even if the project loaded, we sent off a "slow loading" event to GA causing mass confusion and panic. This clears the timeout that triggered the event when a project loads.

In addition, I dropped the slow loading timeout to ``10s`` so that the message will show up a bit faster. I think this is the right move since our average load time seems to be about 5 seconds, so reaching twice the average might be a good time to step in.

Fixes #2212

### DIFF
--- a/public/editor/scripts/main.js
+++ b/public/editor/scripts/main.js
@@ -62,11 +62,11 @@ require(["jquery", "bowser", "analytics"], function($, bowser, analytics) {
 
   Bramble.once("error", onError);
 
-  var errorMessageTimeoutMS = 15000;
+  var slowLoadingTimeoutMS = 15000;
 
-  setTimeout(function(){
+  var slowLoadingTimeout = setTimeout(function(){
     showLoadingErrorMessage();
-  }, errorMessageTimeoutMS);
+  }, slowLoadingTimeoutMS);
 
   function showLoadingErrorMessage(){
     $("#spinner-container .taking-too-long").addClass("visible");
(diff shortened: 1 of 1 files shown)


public/editor/scripts/main.js
```diff
@@ -89,6 +89,7 @@ require(["jquery", "bowser", "analytics"], function($, bowser, analytics) {
     var appUrl = thimbleScript.getAttribute("data-app-url");
     var projectDetails = thimbleScript.getAttribute("data-project-details");
     var editorUrl = thimbleScript.getAttribute("data-editor-url");
+    clearTimeout(slowLoadingTimeout);
```
humphd: I wonder if this is the right place to do this, since this doesn't represent a point where the editor is loaded or usable.  Some other options might include:

* Once Bramble fires the `ready` event here https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/editor/js/bramble-editor.js#L34
* Once the editor is usable, and we're about to remove the spinner/loader here https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/editor/js/fc/bramble-ui-bridge.js#L446

Doing that would require us to move this `timeout` to a separate module, so you can `require` it in a few places and start/stop it.

flukeout: Yeah, that's pretty much why I put it here, just to keep all the related functionality in one place. You're right about ``bramble-ui-bridge`` though. Seems like it belongs in there and was also my initial, pre-lazy assessment.

humphd:
I'm going to roll this into my changes in https://github.com/mozilla/thimble.mozilla.org/pull/2204

PR #2228, opened 2017-05-30
Adds GA event tracking when published project is remixed.

### LINKED ISSUE #2224: Add GA Event when a published project is remixed.
Right now we track pageviews only on published projects, which is pretty handy, but it would be nice to know which projects actually get remixed the most. This is easy enough to add with GA, and could either track the project ID or the project title (or both!)

flukeout:
This will track the ID of any project that is remixed via the Details Bar in Google Analytics.
Tested and shows up like this...

image

Fixes #2224

I'm not sure how to fix the last two remaining lints @gideonthomas

### DIFF
--- a/public/resources/remix/index.js
+++ b/public/resources/remix/index.js
@@ -32,7 +32,7 @@
       if(projectMetaEl) {
         var projectID = projectMetaEl.getAttribute("content") || false;
         if(projectID && typeof ga === "function") {
-          ga("send", {
+          window.ga("send", {
             hitType: "event",
             eventCategory: "Remix Bar",
             eventAction: "Project Remixed",
--- a/public/resources/remix/index.js
+++ b/public/resources/remix/index.js
@@ -31,7 +31,7 @@
       var projectMetaEl = document.head.querySelector("[name=data-remix-projectId]");
       if(projectMetaEl) {
         var projectID = projectMetaEl.getAttribute("content") || false;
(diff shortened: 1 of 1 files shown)


public/resources/remix/index.js
```diff
@@ -19,14 +19,29 @@
...
 
     detailsBar.setAttribute("style", "");
     detailsBar.classList.add("mouse-mode");
 
+    detailsBar.querySelector(".details-bar-remix-button").addEventListener("click",function(event){
+      var projectMetaEl = document.head.querySelector("[name=data-remix-projectId]");
+      if(projectMetaEl) {
+        var projectID = projectMetaEl.getAttribute("content") || false;
+        if(projectID) {
+          ga('send', {
```
humphd: We use `"` vs. `'` throughout this file, let's use `"` here.

humphd: I would also add a check to make sure `ga` is loaded, so we don't crash here because the GA script was slow to load:

```js
if(projectID && typeof ga === "function") {
...
```

flukeout:
@humphd - made the changes you requested (consistent quoting style and checking if ``ga`` is a function). R?

humphd:
@flukeout You're failing lint.  Can you fix and ping me again when it's passing?

flukeout:
@humphd how do i fix this?

```
     35 |          ga("send", {
                   ^ 'ga' is not defined.
```

PR #2234, opened 2017-05-31
Fix #1991, #1524: prefetch bramble resources using HTML5 <link>s vs. script

### LINKED ISSUE #1991: Add some DNS, Resource prefetch and preload meta info to Thimble

humphd:
Finishing work started by @evantsyferov in #2013 and @Mordax in #2008.  This removes the hacked preloading of Bramble's large assets, and uses [<link> prefetching](https://developer.mozilla.org/en-US/docs/Web/HTTP/Link_prefetching_FAQ) instead.

The effect is the same, but this is a lot cleaner, and it also lets us get the `index.html` cached early for the editor too.

### DIFF
--- a/public/homepage/scripts/main.js
+++ b/public/homepage/scripts/main.js
@@ -1,13 +1,4 @@
 /* global requirejs */
-
-// Because we pre-load require scripts needed by the editor, we need to
-// eat error messages related to fetching but not using those modules.
-requirejs.onError = function(err) {
-  if(err.requireType !== "mismatch") {
-    throw err;
-  }
-};
-
 require.config({
   waitSeconds: 120,
   baseUrl: "/{{ locale }}/homepage/scripts",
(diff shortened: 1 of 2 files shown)


flukeout:
With respect to testing @humphd - what am I looking for, should I try out? Is the goal to grab stuff that the editor needs while someone is browsing the homepage?

humphd:
Correct.  If you open your dev tools Network tab, you should be able to see those files get loaded when you're on the homepage.  The browser will use these `prefetch` links as hints, and when it's not busy loading other stuff, it will load them into cache.  We do it to reduce the time it takes to start the editor, since some of those files are quite big.  Previously I was doing this with some custom JS and XHR'ing the files.  This is a better, standards compliant way to achieve the same thing.

flukeout:
Okay testing now.

flukeout:
Hey @humphd looking at the network tab, I see the following...

image

However, when I load up the editor, ``bramble.js`` is still being loaded ...

image

The path for ``bramble.js`` is strange though, has two slashes, could that be why?

image

humphd:
screen shot 2017-06-02 at 7 49 15 pm

Looks like you have `Disable Cache` turned on, so it will always fetch again.

Regarding the double-slash, it's due to how the `env` is doing the URL (trailing `/`) and then we add it again later.  It's fine.

jquery.min.js should get loaded twice, once in Thimble, once in Bramble (different origin).

PR #2238, opened 2017-06-01
Prevents publishing without an index.html and displays error.

### LINKED ISSUE #1952: Publishing projects without index.html fails to locate it.

flukeout:
If you try to publish without having an ``index.html`` page, it does the following:

_no-indexerrr

Link in the error message goes to the [answer in the wiki](https://github.com/mozilla/thimble.mozilla.org/wiki/Using-Thimble-FAQ#why-do-i-need-an-indexhtml-file-when-publishing-my-project).

Currently, the link is embedded into the localization string, should I extract it @TheoChevalier ?

Builds on work started in #1994 by @dsych.

Fixes #1952

### DIFF
--- a/locales/en-US/messages.properties
+++ b/locales/en-US/messages.properties
@@ -219,7 +219,7 @@ publishUnpublishingIndicator=Unpublishing…
 publishProjectDescription=Project Description
 publishCancelBtn=Cancel
 publishDeleteWarning=Are you sure? The current link will stop working and if you publish again, the link will be different.
-noIndexFound=Please add an <strong>"index.html"</strong> file before Publishing your Project.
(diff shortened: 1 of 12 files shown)


TheoChevalier:
> Currently, the link is embedded into the localization string, should I extract it @TheoChevalier ?

@flukeout Definitely safer if you do, but clearly not a blocker if you don’t.

flukeout:
Thanks, @TheoChevalier - I moved the URL into a variable [here](https://github.com/mozilla/thimble.mozilla.org/pull/2238/files#diff-c9a9094a335b05ca97f4236da1e41d8aR38), mind taking a quick look?

TheoChevalier:
> Thanks, @TheoChevalier - I moved the URL into a variable here, mind taking a quick look?

LGTM, thank you 👍

flukeout:
Thanks @TheoChevalier ! Merging and thanks again to @dsych for doing the bulk of the work! 👍

PR #2246, opened 2017-06-01
Fix #2245: Don't use path.join() for URL part concat

### LINKED ISSUE #2245: OAuth redirect uses windows-style path delimiters when running node on Windows
When I run thimble.mozilla.org natively on windows, the oauth redirect is incorrectly formed, due to using `path.join()` - i.e. `http://localhost:3500/en-US/%5Cen-US%5Ceditor`

This is where I believe the problem lies: https://github.com/mozilla/thimble.mozilla.org/blob/92888feb0fa297bed9c473f933fbf2d60b71f032/server/routes/auth/oauth2-callback.js#L24

humphd:


server/routes/auth/oauth2-callback.js
```diff
@@ -151,7 +149,7 @@ module.exports = function(config, req, res, next) {
         );
       }
 
-      res.redirect(301, path.join("/", locale, "/editor"));
+      res.redirect(301, "/" + locale + "/editor");
```
gideonthomas: would you also be able to change this to a 307 instead of a 301?

humphd:
Changed to 307 from 301, going to merge.

Issue #2257, opened 2017-06-02
[Bramble Error] error loading Bramble Error: Bramble.mount() while already mounted


humphd:
I'm reworking the way we load projects for an upcoming PR, and sometimes I see this on Chrome:

```
startup.js:27 [Bramble Error] error loading Bramble Error: Bramble.mount() while already mounted, or attempting to mount.
    at u._.mount (bramble.js:4)
    at bramble.js:4
    at bramble.js:3
    at bramble.js:1
    at bramble.js:3
    at o.<anonymous> (bramble.js:3)
    at r (bramble.js:3)
    at i (bramble.js:2)
    at s (bramble.js:2)
    at IDBRequest.r.onsuccess (bramble.js:3)
```

This comes from https://github.com/mozilla/brackets/blob/master/src/bramble/client/main.js#L476.  I wonder if we have a timing error.

PR #2262, opened 2017-06-05
Allow loading projects as either a single tarball (default) or expanded files

humphd:
This alters the Thimble client and server to allow for a project to be loaded either as a tarball, or as a set of separate files.  I'm doing this behind an opt-in flag, because I want to experiment with it.  I'm curious to see what will happen if we remove the tarball creation from publish.  It is not fully optimized (e.g., I would need to modify publish to send `Buffer` vs. JSON), but that shouldn't matter too much.

To make this work, I've changed the project load order to load the metadata first.  This gives the file IDs if I need them for loading an expanded set of files.

To use it, add `?loadfiles=1` to the Thimble URL.  If you do nothing, it should load the project as a tarball like normal.

### DIFF
--- a/public/editor/scripts/project/metadata.js
+++ b/public/editor/scripts/project/metadata.js
@@ -286,7 +286,7 @@ define(function(require) {
 
   function install(config, callback) {
     if(!config.user) {
-      loadAnonymous(config, callback);
+      return loadAnonymous(config, callback);
     }
 
     setMetadata(config, callback);
--- a/public/editor/scripts/project/metadata.js
+++ b/public/editor/scripts/project/metadata.js
@@ -238,7 +238,7 @@ console.log("config.data", config.data);
 
   // Downloads project metadata (project id, file paths + publish ids).
   function download(config, callback) {
-    if(config.update) {
+    if(!config.user || config.update) {
       // There is no metadata to fetch from the server if this project is
       // being upgraded from an anonymous to a persisted project (we push
       // metadata to the server instead of downloading it)
--- a/public/editor/scripts/project/metadata.js
+++ b/public/editor/scripts/project/metadata.js
@@ -236,7 +236,12 @@ define(function(require) {
     });
   }
 
-  function fetchMetadata(config, callback) {
+  // Downloads project metadata (project id, file paths + publish ids).
+  function download(config, callback) {
+    if(!config.user) {
+      return callback(null, null);
+    }
+
     var url = config.host + "/projects";
     if (config.id) {
       url += "/" + config.id;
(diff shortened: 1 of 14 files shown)


public/editor/scripts/project/metadata.js
```diff
@@ -279,24 +284,12 @@ define(function(require) {
...
-          data: data,
-          id: config.id,
-          root: config.root,
-          user: config.user,
-          title: config.title
-        }, callback);
-      });
+  function install(config, callback) {
+    if(!config.user) {
+      loadAnonymous(config, callback);
```
humphd: um, yup!  Good catch.

humphd:
I noticed that error, too, and filed it at https://github.com/mozilla/thimble.mozilla.org/issues/2257.  I think it might be that missing `return` you found.

I've fixed everything you noticed, want to test again?

gideonthomas:
I still see the failure with the `upgrade` workflow I mention above.

humphd:
OK, thanks, I'll debug it locally.  I was hoping that would be all it was.

humphd:
I tried to fix this quickly tonight, but I think I'm going to have to spend a bunch of time on it to get it right.  The problem is that we used to create the project dir when we grabbed the tarball, but now I need it sooner so I can start writing file attributes.  I'll try and do that in the next few days.

humphd:
@gideonthomas I think this is working well now, can you re-test please?  You can change your `env` for the load type to change it from `"tarball"` to `"files"`.

public/editor/scripts/project/metadata.js
```diff
@@ -226,7 +226,7 @@ define(function(require) {
       // [{ id: 1, path: "/index.html", project_id: 3 }, ... ]
       if (config.data) {
         project.id = config.id;
-
+console.log("config.data", config.data);
```
gideonthomas: remove

server/routes/main/editor.js
```diff
@@ -6,9 +6,13 @@ var querystring = require("querystring");
 var defaultProjectNameKey = require("../../../constants").DEFAULT_PROJECT_NAME_KEY;
 var utils = require("../utils");
 var HttpError = require("../../lib/http-error");
-
+var env = require("../../lib/environment");
 var snippets = require("../../lib/snippets");
 
+// One of "tarball" or "files" to specify how projects should get loaded.
+// If not specified, the client code will default to "tarball".
+var projectLoadStrategy = env.get("PROJECT_LOAD_STRATEGY");
```
gideonthomas: can we do this in server/routes/config.js instead and pass in `config.projectLoadStrategy` below?

gideonthomas:
r+

PR #2280, opened 2017-06-07
Fix #2275: add CSP for https://mozillathimblelivepreview.net

### LINKED ISSUE #2275: [Firefox] Fix CSP blocking precaching of editor
Firefox is reporting that our precaching isn't working due to CSP:

screen shot 2017-06-07 at 12 15 14 pm

We probably need to add the domain for the editor.

humphd:


server/security.js
```diff
@@ -6,7 +6,10 @@ let csurf = require("csurf");
 const ONE_YEAR = 31536000000;
 
 let defaultCSPDirectives = {
-  defaultSrc: [ "'self'" ],
+  defaultSrc: [
```
gideonthomas: You don't want to add it here. You want to add it here: https://github.com/mozilla/thimble.mozilla.org/blob/master/server/index.js#L82-L87

humphd:
OK @gideonthomas, moved it as requested.  This right?

Issue #2241, opened 2017-06-01
Fix remix script injecting


lkisac:
Remix script tag not added when file is missing `<head>` tag.

Cross-reference #1804

PR #2287, opened 2017-06-12
inject remix script for html with missing <head> tags

lkisac:
Addresses issue #2241.

Uses [jsdom](https://www.npmjs.com/package/jsdom) to build DOM from html string and query for missing `<head>` tags.  If they do not exist, adds empty `<head>` and `<title>` tags:

i.e. https://thimbleprojects.org/jgmac1106/223453/

```html
<!DOCTYPE html>
<header>
  <script src="https://use.fontawesome.com/e83e9b1e77.js"></script>
  <!-- Latest compiled and minified CSS --> 
  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTs
  <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.1/jquery.min.js"></script>
  <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.min.js"></script>
  <!-- Latest compiled and minified JavaScript --> 
  <link href="style.css" rel="stylesheet">
  <script src="https://thimble.mozilla.org/resources/remix/index.js" type="text/javascript"></script>
</header>
<html>
  <body>
   ...
  </body>
</html>
```
replaced with:

```html
<!DOCTYPE html>
<header>
  <script src="https://use.fontawesome.com/e83e9b1e77.js"></script>
  <!-- Latest compiled and minified CSS --> 
  <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTs
  <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.1.1/jquery.min.js"></script>
  <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.min.js"></script>
  <!-- Latest compiled and minified JavaScript --> 
  <link href="style.css" rel="stylesheet">
  <script src="https://thimble.mozilla.org/resources/remix/index.js" type="text/javascript"></script>
</header>
<html>
  <head>
    <title>Untitled</title>
    <script src="http://localhost:3500/resources/remix/index.js" type="text/javascript"></script>
  </head>
  <body>
   ...
  </body>
</html>

```

### DIFF
--- a/services/publish.webmaker.org/lib/remix.js
+++ b/services/publish.webmaker.org/lib/remix.js
@@ -2,6 +2,7 @@
 
 const Hoek = require(`hoek`);
 const Url = require(`url`);
+const jsdom = require(`jsdom`);
 
 const REMIX_SCRIPT = process.env.REMIX_SCRIPT;
 
@@ -9,6 +10,7 @@ Hoek.assert(REMIX_SCRIPT, `Must define location of the remix script`);
 
 const remixUrl = Url.parse(REMIX_SCRIPT);
 const slashes = remixUrl.slashes ? `//` : ``;
+const { JSDOM } = jsdom;
 
 function injectMetadata(html, metadata) {
   let metaTags = ``;
@@ -22,10 +24,27 @@ function injectMetadata(html, metadata) {
 }
 
 function injectRemixScript(html) {
-  return html.replace(
-    /<\/head/,
-    `<script src="${REMIX_SCRIPT}" type="text/javascript"></script>\n$&`
-  );
+  // Check if <head> tag exists
+  var retInject = ``;
+  const dom = new JSDOM(html);
+
+  // If they exist, inject remix script before closing </head> tag
+  if (dom.window.document.querySelector(`head`)) {
+    console.info(`has head tags`);
+    retInject = html.replace(
+      /<\/head/,
+      `  <script src="${REMIX_SCRIPT}" type="text/javascript"></script>\n  $&`);
+    console.info(retInject);
+  // If they don't exist, add empty <head><\/head> tags and inject remix script
+  } else {
+    console.info(`missing head tags`);
+    retInject = html.replace(
+      /<html>/,
+      `<html>\n  <head>\n    <title>Untitled<\/title>\n    <script src="${REMIX_SCRIPT}" type="text/javascript"></script>\n  <\/head>`);
+    console.info(retInject);
+  }
+
+  return retInject;
 }
 
 // Inject the Remix script into the given HTML string
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "helmet": "^3.5.0",
     "jit-grunt": "^0.10.0",
     "jquery": "^3.2.1",
+    "jsdom": "^11.0.0",
     "less-middleware": "2.2.0",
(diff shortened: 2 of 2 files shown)


services/publish.webmaker.org/lib/remix.js
```diff
@@ -22,10 +24,27 @@ function injectMetadata(html, metadata) {
...
-  return html.replace(
-    /<\/head/,
-    `<script src="${REMIX_SCRIPT}" type="text/javascript"></script>\n$&`
-  );
+  // Check if <head> tag exists
+  var retInject = ``;
+  const dom = new JSDOM(html);
+
+  // If they exist, inject remix script before closing </head> tag
+  if (dom.window.document.querySelector(`head`)) {
```
Pomax: Note that JSDOM is a spec-compliant DOM parser, which means that no matter what code you throw into it, it generates a proper DOM, following the `<html><head/><body/></html>` pattern. As such, we no longer need to check for a head element, and we don't have to use strings: we can now solve this with pure DOM operations:

```
function injectRemixScript(html) {
  // We should check if this constructor can every throw. If it can,
  // we need to make sure we catch that.
  const dom = new JSDOM(html);

  // Here, as well, we should check the jsdom API to make sure that
  // there are no edge cases that can lead to a successful call to
  // the JSDOM constructor, but errors trying to access "window"
  const document = dom.window.document;

  // because we're using a spec-compliant DOM parser,
  // we are now guaranteed to have a `<head>` element.
  const head = document.head;

  // We can inject the remix script using standard DOM operations now:
  const remixScript = document.createElement("script");
  remixScript.src = REMIX_SCRIPT;
  head.appendChild(remixScript);

  // and we can then get the DOM-compliant code using the 
  // "outerHTML" property, which is supported by all modern
  // browsers: http://caniuse.com/#search=outerHTML
  const updatedHTML = document.querySelector("html").outerHTML;
  return updatedHTML;
}
```

lkisac: @Pomax Will Thimble be updating its node version to 6+?  If so, I can use this new jsdom api's features

humphd: That's https://github.com/mozilla/thimble.mozilla.org/pull/1714, which we should finish at some point, cc @gideonthomas.

humphd: See https://github.com/mozilla/thimble.mozilla.org/issues/2312.  We should be able to require a newer node.js for Thimble, and isolate what we use for publish in Vagrant.

gideonthomas:
Thanks for your work on this! We're going to be changing the infrastructure around how data is stored which will affect how remixes occur. Hence, we would need to revisit this issue then.

lkisac:
Sounds good @gideonthomas!   I will keep an eye out for the changes

PR #2360, opened 2017-07-12
Update Node

gideonthomas:


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- '4.6'
+- '4.8'
```
Pomax: 4.8.4 -- just 4.8 is not good enough

PR #2377, opened 2017-07-17
Updated localization strings for Image Viewer

flukeout:


locales/en-US/editor.properties
```diff
@@ -145,7 +145,9 @@ DND_SUCCESS_UNZIP=Successfully unzipped <b>{0}</b>.
 DND_SUCCESS_UNTAR=Successfully untarred <b>{0}</b>.
 
 # Image Viewer
-IMAGE_DIMENSIONS={0} (width) &times; {1} (height) pixels
+IMAGE_FILE_TITLE=Image File
+IMAGE_DIMENSIONS={0} &times; {1} pixels
```
humphd: This is a change, so I think we need a new string key.  cc @TheoChevalier

flukeout:
Okay, will await Theo's instructions.

locales/en-US/editor.properties
TheoChevalier: Yes please, it needs to be retranslated

flukeout:
@humphd updated the string key name

locales/en-US/editor.properties
```diff
@@ -145,7 +145,9 @@ DND_SUCCESS_UNZIP=Successfully unzipped <b>{0}</b>.
 DND_SUCCESS_UNTAR=Successfully untarred <b>{0}</b>.
 
 # Image Viewer
-IMAGE_DIMENSIONS={0} (width) &times; {1} (height) pixels
+IMAGE_FILE_TITLE=Image File
+IMAGE_DIMENSIONS_1={0} &times; {1} pixels
```
humphd: I would use `IMAGE_PIXEL_INFO` vs. a versioned one, but it's up to you.  Make sure whatever you use gets changed in Brackets too.

humphd:
@gideonthomas make sure you're cool with this approach (e.g., removing the old, versioning the new) and then this is good to go from my end.

locales/en-US/editor.properties
gideonthomas: I think the l10n guidelines support your suggestion - create a meaningful name vs. just tacking on numbers.

gideonthomas:
Yeah I think this looks good

flukeout:
Okay cool, I'm adding an "Audio Viewer" so I'll update this with those strings as well.

PR #2389, opened 2017-07-20
Add strings needed by Brackets for https://github.com/mozilla/brackets/pull/847

humphd:
These strings are needed for the change in https://github.com/mozilla/brackets/pull/847

### DIFF
--- a/locales/en-US/editor.properties
+++ b/locales/en-US/editor.properties
@@ -26,6 +26,7 @@ ERROR_PROJECT_SIZE_EXCEEDED=The total size of the files in your project has exce
 
 # File Open/Save Error strings
 
+FILE_EXISTS_HEADER=The file already exists.
 # The {0} here will be replaced by an actual error message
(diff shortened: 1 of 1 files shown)


locales/en-US/editor.properties
```diff
@@ -101,6 +102,10 @@ SAVE_AND_OVERWRITE=Overwrite
 DELETE=Delete
 BUTTON_YES=Yes
 BUTTON_NO=No
+# "Use Imported" as in "Use the imported version of this file vs. the existing one"
+USE_IMPORTED=Use Imported
+# "Use Existing" as in "Use the existing version of this file vs. the imported one"
```
gideonthomas: nit: "Keep Existing" vs. "Use Existing" and "Keep the existing" vs. "Use the existing"

flukeout:
Should it say, ``A file with this name already exists.`` or are we actually checking to see if it's the exact same file? Also, when does this pop up? Whenever you drag in or upload a new file? 

We can just say "Keep existing" and "Use new" and ditch "Importing" as a verb.

Issue #2331, opened 2017-06-21
"Unzip Completed Successfully" dialog only shows up when 1 of 3 available ways to import is used.


flukeout:
This dialog here...

![image](https://user-images.githubusercontent.com/25212/27409630-4bff5394-5697-11e7-9ac3-9df82596e56f.png)

Only shows up if you use the **Upload Files** dialog and select **From your Computer...".

If you drag it in directly or drag it into the Upload Files dialog, the notice doesn't show up.

Let's show the notice no matter the method used.

cc @humphd

humphd:
Can we pair this with https://github.com/mozilla/thimble.mozilla.org/issues/2285?  I really want to move that forward so we properly signal that we're doing work in long running processes like import, resize, unzip, untar, etc.

PR #2390, opened 2017-07-20
Remove unneeded strings for unzip/untar

humphd:
For https://github.com/mozilla/thimble.mozilla.org/issues/2331 I'm removing these strings, since we don't know the name of the file when we import and unzip.


gideonthomas:
Do we still need the TITLE strings above those?

humphd:
Yes, I'm only going to show the title.

gideonthomas:
sounds good

PR #2393, opened 2017-07-21
Fix #2250: Change load error message

### LINKED ISSUE #2250: Change "Oops, something went wrong" page title.
Currently when the editor doesn't start we show this...

image

However, I think we should just say something like **Thimble could not start** so that it makes it more obvious to the user what's going on when they are reaching out to us via email.

MhdTlb:


MhdTlb:
Fix for #2250

locales/en-US/messages.properties
```diff
@@ -20,21 +20,22 @@ termsOfUse=Terms of Use
 privacyPolicy=Privacy Policy
 thimbleFooter=Thimble is part of the Mozilla Learning Network. <br/>Learn to teach like mozilla at <a href="http://learning.mozilla.org">learning.mozilla.org</a>.
 pageTitleGetInvolved=Thimble by Mozilla - Get Involved
+############
```
gideonthomas: why was this change made?

locales/en-US/messages.properties
```diff
@@ -20,21 +20,22 @@ termsOfUse=Terms of Use
...
-## Editor ##
-############
 
 # Loading the editor
 loadingProject=Loading Project
 loadingProjectSupportSuggestion=If the project doesn't load, <a title="Thimble help page" href="https://github.com/mozilla/thimble.mozilla.org/wiki/Project-not-Loading-Help">get help here</a>.
-somethingWentWrong=Thimble could not start.
+somethingWentWrong=Oops, something went wrong!
 reloadThimbleMessage=Please reload your browser to try loading Thimble again. <button class="refresh-browser">Reload Now</button>
+errorCouldNotStart=Thimble could not start.
```
gideonthomas: This string should be "Oops, something went wrong!"

locales/en-US/messages.properties
```diff
@@ -20,21 +20,22 @@ termsOfUse=Terms of Use
...
 
-############
-## Editor ##
-############
 
 # Loading the editor
 loadingProject=Loading Project
 loadingProjectSupportSuggestion=If the project doesn't load, <a title="Thimble help page" href="https://github.com/mozilla/thimble.mozilla.org/wiki/Project-not-Loading-Help">get help here</a>.
-somethingWentWrong=Thimble could not start.
+somethingWentWrong=Oops, something went wrong!
```
gideonthomas: this key should not exist anymore

gideonthomas:
@flukeout does the text look right?

PR #2400, opened 2017-07-22
Fix a hardcoded string in MozFest banner

TheoChevalier:
One of our localizers noticed the error, this should fix it


gideonthomas:
Thanks for catching this!

PR #2410, opened 2017-07-31
Webpack editor

gideonthomas:
Relies on https://github.com/mozilla/thimble.mozilla.org/pull/2404 landing first.

Changes made:
- Remove the requirejs build step from the Gruntfile
- Remove the require.js script from the HTML as well as delete the static script we had in `vendor`
- Switched to CommonJS and use the `shared` module from previous patches.

Haven't made any logical code changes.


gideonthomas:
@cadecairos sorry, I should have mentioned this in the description - you need to run `npm install` first because I made `bowser` a node module instead of the copy hack we had before.

PR #2421, opened 2017-08-03
Update node to 8

gideonthomas:
This patch updates the node version to 8 and also removes Thimble out of Vagrant to make dev easier (yay no more breakable watch scripts).

You'll need to run this with Vagrant as per the Readme instructions. Would be good to test out on Windows as well.


cadecairos:
Why not update to the active LTS branch (6)?

humphd:
I tend to agree, let's use LTS unless we have to use 8

PR #2422, opened 2017-08-03
Move thimble out of vagrant and use Node 6

gideonthomas:
This patch updates the node version to 8 and also removes Thimble out of Vagrant to make dev easier (yay no more breakable watch scripts).

You'll need to run this with Vagrant as per the Readme instructions. Would be good to test out on Windows as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ public/stylesheets/userbar.css
 dist/
 client/
 mozilla.webmaker.org
+package-lock.json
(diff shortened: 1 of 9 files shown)


gideonthomas:
@humphd and @cadecairos I chose node 8 because it's becoming LTS in October, which isn't that far away.

cadecairos:
even still, I would prefer we use an active LTS branch, rather than bank on nothing being shipped to 8.x in the next two months that breaks us.

gideonthomas:
fixed

PR #2428, opened 2017-08-09
Strings for new binary viewer

flukeout:
Added localization strings for the Binary Viewer, which is an in-progress PR...

* https://github.com/mozilla/brackets/pull/851

### DIFF
--- a/locales/en-US/editor.properties
+++ b/locales/en-US/editor.properties
@@ -195,7 +195,8 @@ AUDIO_OPTION_DISABLE_AUDIO=Start audio player muted
 
 # Binary Viewer
 BINARY_FILE_TITLE=Non-editable File
-BINARY_FILE_CANNOT_EDIT=I'm not sure how to edit this file!
+BINARY_FILE_CANNOT_EDIT=Thimble is not sure how to edit this file!
+# Title for binary file actions that user can take
 BINARY_FILE_OPTIONS_TITLE=But you can...
 BINARY_FILE_DOWNLOAD=Download it
 BINARY_FILE_OPEN=Open it in a new tab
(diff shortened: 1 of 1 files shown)


locales/en-US/editor.properties
```diff
@@ -193,6 +193,14 @@ AUDIO_OPTION_AUTOPLAY=Start playing the audio automatically
 AUDIO_OPTION_LOOPING=Restart the audio when it is over
 AUDIO_OPTION_DISABLE_AUDIO=Start audio player muted
 
+# Binary Viewer
+BINARY_FILE_TITLE=Non-editable File
+BINARY_FILE_CANNOT_EDIT=I'm not sure how to edit this file!
```
humphd: I don't love the use of "I'm" here.  What about the third-person "Thimble doesn't know how to edit this file!"?

locales/en-US/editor.properties
```diff
@@ -193,6 +193,14 @@ AUDIO_OPTION_AUTOPLAY=Start playing the audio automatically
...
+BINARY_FILE_OPTIONS_TITLE=But you can...
```
humphd: I think you should put a comment above this so it's clear how this will connect to the options beneath.

gideonthomas: How is this displayed in the UI? Are the options below shown in a dropdown? If it is then it's fine. Otherwise if it is being attached to this string, you need a placeholder here for where the rest of the string will appear.

flukeout: The options are shown directly below this line. I'll add a comment.

flukeout:
@humphd I made the requested changes.

Issue #2041, opened 2017-05-02
Add touch events for the remix bar


gideonthomas:
In https://github.com/mozilla/thimble.mozilla.org/pull/1825, we noticed that the touch events for the remix bar don't work (so we removed them).

We need to implement events for the remix bar on touch devices so that it is usable. The code that we removed was:
```js
var isTouchDevice = 'ontouchstart' in document.documentElement;
var detailsBar = $(".details-bar");
detailsBar.attr("style", "");

if(isTouchDevice) {
  detailsBar.addClass("touch-mode");
} else {
  detailsBar.addClass("mouse-mode");
}
```

We also don't use jquery any more so we would need to handle it using regular DOM manipulation.

cc @flukeout

aayushsanghavi:
@gideonthomas I'd like to work on this

gideonthomas:
sure thing @aayushsanghavi! @flukeout and @Pomax might be able to give a bit more information about what's required here.

Rajat-dhyani:
i looked at @Pomax  comment [here](https://github.com/mozilla/thimble.mozilla.org/pull/1825/files/a23d523979ab953f074ec8544c03e651bbaa2bfc#r108536379)
i think ontouchstart works for desktop i.e `return false` (on chrome) only if we are not viewing it in developer mode.
I have done some changes according to touch event [here](https://github.com/mozilla/thimble.mozilla.org/commit/7936e777bdcc2e58224b21959670ad1062f22251)
Though i am not sure if these are the things that you wanted to add here.

@aayushsanghavi  you can check these changes it might help you out if you are still working on this or otherwise i will open up the PR for the same.

cc: @gideonthomas @flukeout

aayushsanghavi:
Looks like you've been working on this @Rajat-dhyani. Its good that you filed a PR. I was actually out of town for a couple of weeks so I was unavailable all this while. I guess you can continue your work on this issue then. 👍

flukeout:
In looking into how this is handled, it seems like we may not be able to reliably detect when a browser is "touch-enabled". Chrome has changed it's implementation of the method we used to rely on. I think we should come up with a different solution here and maybe just rely on media queries and look for screen width, or have a design that just works in both circumstances (move away from ``hover`` or ``mouseenter`` approaches. I'll refile.

PR #2095, opened 2017-05-14
Fix:2041 Add touch events for the remix bar

Rajat-dhyani:
This adds touch event to remix bar.

 #2041

PR #2433, opened 2017-08-12
Fix:2041 Add touch events for the remix bar

Rajat-dhyani:
This adds touch event to remix bar.

 #2041

### DIFF
--- a/public/resources/remix/index.js
+++ b/public/resources/remix/index.js
@@ -25,7 +25,48 @@
     }
 
     detailsBar.setAttribute("style", "");
-    detailsBar.classList.add("mouse-mode");
+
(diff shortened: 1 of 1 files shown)


Rajat-dhyani:
@flukeout 
 i added the way to detect the touch event as was proposed by you [here](https://github.com/mozilla/thimble.mozilla.org/pull/2095#issuecomment-316532936) but it is not working.
Can you look into this ?

flukeout:
@Rajat-dhyani in looking into this again, it looks like our old method (which used to work) isn't reliable anymore. For example, Chrome on the desktop thinks that it is touch enabled, while Firefox doesn't. I think we'll need to redesign the remix bar to just have on standard style, or maybe change it's layout depending on the screen width.

I'll close this PR until we figure out a better, more reliable solution.

PR #2435, opened 2017-08-15
[WIP] Implement Social Sign-In

ryanwarsaw:
**Warning:** This is still is progress, don't merge!

### DIFF
--- a/env.dist
+++ b/env.dist
@@ -71,9 +71,9 @@ export AWS_SECRET_ACCESS_KEY="foo"
 # id.wm.o config
 #
 
-export OAUTH_CLIENT_ID="test"
-export OAUTH_CLIENT_SECRET="test"
-export OAUTH_AUTHORIZATION_URL="http://localhost:1234"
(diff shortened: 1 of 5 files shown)


ryanwarsaw:
**Update**

So this code actually compiles and runs now, but I want to do some better error handling before asking for a review, and considering pushing this to production. This does contain some breaking changes, the most prominent one being the changes to the `.env` file. 

The `passport-webmaker` strategy also reformats the response from the `id.webmaker.org` server, so there is potential for breakage there, but I've gone through the code pretty in-depth to try and find as many if not all of the places where this could cause problems. 

If you try to run this locally, make sure you re-populate your `id.webmaker.org` database with new dummy data, where you point the callback URL to `/login/webmaker/callback` rather than it's default which is `/callback` (since this is a non-existent route now).

ryanwarsaw:
Hey @gideonthomas I think this is ready for review, I've taken care of all the items I've outlined above, let me know if you've got any suggested changes, or need help getting this running.

This isn't a very easy PR to review, it'll probably take a few hours just a heads up.

Issue #2462, opened 2017-09-03
Default Project CSS has Multiple Font @include Statements Where Only One is Needed


emctague:
In the Thimble default project's `style.css`, the following two `@include` lines are present, to include both Open Sans and Merriweather:

```css
@import url('https://fonts.googleapis.com/css?family=Open+Sans:400,700');
@import url('https://fonts.googleapis.com/css?family=Merriweather:400,700');
```

This could be cut down to one, much simpler `@include` statement without impacting readability:

```css
@import url('https://fonts.googleapis.com/css?family=Open+Sans|Merriweather:400,700');
```

I'd be more than happy to submit a PR for this issue, because it's kinda bugging me. I've already fixed it in [my fork](https://github.com/emctague/thimble.mozilla.org/commit/9772981674f97500cf344c117f3b243627a5423b).

ryanwarsaw:
From a quick glance, I'm not sure if this is a change we'd want to make. I think it might end up confusing Thimbles further, since the special syntax being used to combine the imports may not
be inherently clear to the user, where as IMO it's more clear now with the two differing import(s).

I'd still submit the PR, and then it can be deliberated further on what we want to do with this.

gideonthomas:
Closing since we think it's clearer to keep the code the way it is.

PR #2463, opened 2017-09-03
Merge Default Project Font @include Statements into One

emctague:
I've merged the two, separate `@include` statements in the default project into a single line. I believe this will make the default CSS a little bit smaller and less intimidating for anyone new to CSS, since there's one fewer lines of code for a newbie to try and read and understand. @ryanwarsaw reports that he believes the syntax of the multi-font syntax in a google fonts URL will be too confusing, but I doubt that most people new to development will bother trying to interpret the include URL(s) at all.

Of course, from an optimization standpoint, it'll also ever-so-slightly cut down the number of HTTP requests needed to load the fonts and reduce file size - but this is probably a bit less important to the targeted user base.


gideonthomas:
Fixes #2462

gideonthomas:
Thanks for the contribution, but it looks like this might not be a direction we want to pursue.

PR #2472, opened 2017-09-08
Fixes#2259:completes the incomplete string on Get Involved page

### LINKED ISSUE #2259: Incomplete string on Get Involved page
On Get Involved page, one of Brian Yang's contributions (Added support for file icons in) is incomplete. I guess there should be one more word (Thimble; Editor; something else?).

apoorvasingh17:
Fixes #2259


apoorvasingh17:
@flukeout Hi. Please review my PR. Thanks!

flukeout:
Thanks @apoorvasingh17 - looks good. We just need to check about localization.

@TheoChevalier - will this be OK for localization? We had an incomplete string - will this trigger localizers to re-translate or should we rename the key as well?

TheoChevalier:
@flukeout the only way to get a string re-translated is to update the key, so yes we need to do that here

flukeout:
That's what I thought - thanks @TheoChevalier.

@apoorvasingh17 - can you please update the key in the string to something new like...

``peiying16Contrib1fix=Added support for file icons in the file list``

And then we have to change that key name in the Get Involved page as well. Let me know if you need help. Thanks!

apoorvasingh17:
@flukeout Updated the key. Please check. Thanks!

flukeout:
Thank you for your help & contribution @apoorvasingh17

PR #2473, opened 2017-09-09
Fixes #2469 Add scroll in Editor dropdown menu

### LINKED ISSUE #2469: Add scroll in Editor dropdown menu
The last options in the editor dropdown menu hide when the browser window is not maximized and therefore, these options can't be accessed.

editor_dropdown

editor_dropdown2

This is seen in Firefox and Chrome as well.
Maybe, adding a scroll to menu will work...

ayush9398:
Fixes #2469

### DIFF
--- a/public/editor/stylesheets/editor.less
+++ b/public/editor/stylesheets/editor.less
@@ -537,6 +537,8 @@ body {
     list-style-type: none;
     margin: 0;
     padding: 0;
+    overflow-y: scroll;
+    max-height: 252px;
 
     li {
       padding: 20px 100px 20px 20px;
--- a/public/editor/stylesheets/editor.less
+++ b/public/editor/stylesheets/editor.less
@@ -537,8 +537,8 @@ body {
     list-style-type: none;
     margin: 0;
     padding: 0;
-    overflow-y: scroll;
-    max-height: 252px;
+    overflow-y: auto;
+    max-height: calc(~"100vh - 150px");
 
     li {
       padding: 20px 100px 20px 20px;


flukeout:
Hi @ayush9398 thanks for submitting. Can you please update your PR description to include a reference to the issue you are fixing, in this format...

``Fixes #XXXX`` - where XXXX is the issue number. 

Then, when this PR is merged, that issue will automatically get closed. Thanks!

flukeout:
As far as the PR itself goes, we don't want to set a fixed height on the list of options. Doing this makes it so that we always have to scroll the list if the option we want to change is at the bottom. We only want a scrolling list when there is not enough space to show the whole dialog.

Here are the rules we should apply instead...

```
overflow-y: auto;
max-height: calc(~"100vh - 150px");
```

This makes is so that the scroll bar is shown ONLY when necessary. A value of ``scroll`` shows blank scrollbar even if when scrolling is not possible.

The second rule says, we want a height that is 100% of the height of the available window height - minus 150px. This will ensure that the dialog is always cut off just a bit before the bottom of the editor, like this...

image

Please make the following change in your PR, test it locally in both Chrome and Firefox (or any other browser you are using) and we can get it merged. Thanks!

ayush9398:
@flukeout I have made the changes....

flukeout:
Thanks @ayush9398 - will test right now.

flukeout:
Great, thanks for your work on this @ayush9398 - tested and it works great in both Firefox and Chrome.

PR #2477, opened 2017-09-11
Fixed styling of the Sign Out error page.

### LINKED ISSUE #2450: "Please Sign Out" page is broken.
Styles are all messed up.

image

Easiest way to see it is...
* Sign in
* Go to project list
* Clear your cache 
* Reload browser

flukeout:
This adds the correct styling to the Sign Out page. It can be viewed by following the steps to reproduce in #2450 and should appear now like this...

image

Fixes #2450

### DIFF
--- a/views/sign-out.html
+++ b/views/sign-out.html
@@ -7,7 +7,7 @@
     <link rel="stylesheet" href="/homepage/stylesheets/style.css">
     <link rel="stylesheet" href="/resources/stylesheets/error.css">
   </head>
-  <body class="homepage">
+  <body class="error-page">
 
     <nav class="navbar-thimble container-flex">
       <div id="navbar-left" class="container-flex">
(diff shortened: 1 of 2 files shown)


views/sign-out.html
```diff
@@ -27,7 +27,9 @@
         <div class="reload-message">
           <h1>{{ gettext("pleaseSignOut") }}</h1>
           <p>{{ gettext("pleaseSignOutMessage") }}</p>
-          <a href="{{ logoutURL }}" class="action-button" target="_self">{{ gettext("signOutButtonLabel") }}</a>
+          <p class="button-wrapper">
```
gideonthomas: nit: `<p>` vs. `<div>`?

flukeout: Have kept as ``p`` since the other page that uses the same markup has a ``p`` so will leave for now.

PR #2481, opened 2017-09-12
 Fixes #2378 : Added locale string for PDF Viewer Title 

### LINKED ISSUE #2378: Update and localize PDF viewer title to say "PDF File"
Instructions for doing this below...

You have to pass it into the iframe via a data attribute:

Make a new string, Strings.PDF_File or something in Thimble/Brackets
Add a data-title attribute to the pdf iframe template here https://github.com/mozilla/brackets/blob/master/src/extensions/extra/PDFView/htmlContent/pdf-view.html#L4-L5
Add it to the template rendering, and use Strings.PDF_FILE or whatever you added above.
Read the values off the <iframe ... data-*> attributes https://github.com/mozilla/brackets/blob/master/src/extensions/extra/PDFView/bramble-pdfjs.js#L5-L8 and then use them in the UI

Cryzek:
@flukeout 
Added locale string to "locales/en-US/editor.properties" for PDF Viewer title corresponding to changes done in brackets.

### DIFF
--- a/locales/en-US/editor.properties
+++ b/locales/en-US/editor.properties
@@ -202,6 +202,9 @@ BINARY_FILE_DOWNLOAD=Download it
 BINARY_FILE_OPEN=Open it in a new tab
 BINARY_FILE_TRY_EDIT=Try to edit it
 
+#PDF Viewer
+PDF_FILE_TITLE=PDF File
+
 ################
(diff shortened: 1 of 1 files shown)


flukeout:
Thanks very much @Cryzek - will merge this in so things are ready for the Brackets PR.

PR #2494, opened 2017-09-20
Enable CORS for all origins on /refresh route

gideonthomas:


server/routes/main/index.js
```diff
@@ -27,7 +29,7 @@ module.exports = {
     // Refresh Editor Page
     app.get(
       "/refresh",
-      middleware.enableCORS(config.editorHOST),
+      cors(), // Enable all origins to access this route
```
humphd: This is fine, but it introduces two different ways we are adding CORS.  Why not modify the `enableCORS()` function do basically call `cors()` when no domain is passed in?  Then we would always use it vs. only sometimes, and be a bit more consistent.

PR #2495, opened 2017-09-20
Fix 404 on /refresh route, force top-level frame to reload

humphd:
We have a 404 when trying to load the `/refresh` route, which this fixes.  Also, this does the reload on the top-level iframe vs. the current window, which we'll need in Brackets.

### DIFF
--- a/views/refresh.html
+++ b/views/refresh.html
@@ -45,6 +45,6 @@
       </div>
     </section>
 
-    <script src="/scripts/error.js"></script>
+    <script src="/resources/scripts/error.js"></script>
   </body>
 </html>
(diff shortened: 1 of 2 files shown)


views/refresh.html
```diff
@@ -45,6 +45,6 @@
       </div>
     </section>
 
-    <script src="/{{ localeDir }}/error/scripts/main.js"></script>
+    <script src="/scripts/error.js"></script>
```
gideonthomas: this should be `/resources/scripts/error.js`

humphd: Thanks, fixed.

PR #2502, opened 2017-09-22
Appveyor and env cleanup

gideonthomas:
This also removes a bunch of env vars we don't use anymore.

### DIFF
--- a/env.dist
+++ b/env.dist
@@ -5,97 +5,63 @@
 WEB_CONCURRENCY=1
 
 # development or production
-export NODE_ENV="development"
+NODE_ENV="development"
 
 # default thimble port is 3500
-export PORT=3500
+PORT=3500
 
 # hostname for this server according to your browser
 # if they don't match then Persona will not work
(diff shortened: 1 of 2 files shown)


env.dist
```diff
@@ -5,97 +5,63 @@
...
 
 # Is this server running behind an SSL-enabled load-balancer?
-export FORCE_SSL=false
+FORCE_SSL=false
 
 # i18n Language Support
-export L10N_SUPPORTED_LANGUAGES="[ "*" ]"
-export L10N_LOCALE_SRC="locales"
-export L10N_LOCALE_DEST="dist/locales"
+L10N_SUPPORTED_LANGUAGES="[ "*" ]"
```
humphd: I'm sure this is right, but the way the `"`s are embedded here is odd.

gideonthomas: yeah it's weird but the only reason we do this is because we can rely on habitat which parses it for us. Regular env parsing won't parse it correctly, we would need to manually `JSON.parse` it.

Issue #2116, opened 2017-05-15
Feature request > Allow toggle of Editor options by clicking the option label


flukeout:
In this menu here...

![image](https://cloud.githubusercontent.com/assets/25212/26074345/50d0afd8-3966-11e7-9b68-63ccb11f36e7.png)

Would be nice to be able to click the label the change the setting instead of having to click the toggle itself. When hovering on the label, we should highlight the toggle somehow to show that it will be toggled if you click.

feihaozi77:
Can I work this one? if no one is being assigned?

humphd:
@feihaozi77 it's all yours.

flukeout:
Sounds good @feihaozi77 - all yours. Let me know if you have questions. This will require you running both Thimble and Brackets locally.

feihaozi77:
can you provide a brief STR ?

flukeout:
Sure thing, check out this gif...

![toggle-click](https://user-images.githubusercontent.com/25212/30981343-ac956ad6-a449-11e7-961a-a62073a72611.gif)

Notice how only clicking the actual toggle changes the setting? When I click the label nothing happens, but it should! Clicking the label should have the same behavior as clicking the toggle UI.

feihaozi77:
@flukeout i was following the automated installation section in the readme file to set up the environment, but there are still some confusions in my mind, since you told me both thimble and brackets should be running locally, could you explain what is the relationship between those two tools? (I know brackets is a editor while thimble is like a online web based editor)

flukeout:
Take a look below...

<img width="948" alt="image" src="https://user-images.githubusercontent.com/25212/30986574-81aa1ee2-a459-11e7-9b97-87d7e6d46266.png">

In essence, everything above the red line is "Thimble" and everything below is "Brackets". The homepage also lives in the Thimble repo. Let me know if that helps. Thimble is like the wrapper piece that adds the UI for all of the editor settings, account management as well as the ability to publish etc.

feihaozi77:
@flukeout when i modify the main.js in dist/editor/scripts and restart the server, the main.js always restore back to the previous state which is without my modification. Do you know what is the cause of this?

humphd:
cc @gideonthomas, who can explain the webpack build stuff to you, and where things live in the src.

feihaozi77:
[...]

PR #2510, opened 2017-09-30
Added Text Toggle functionality for autocomplete  Issue #2116 

feihaozi77:


feihaozi77:
Yeah, Of course , I will fix this

feihaozi77:
@humphd I agree with your opinion on "#autocomplete-toggle-text is very similar to #autocomplete-toggle above it. If we find a bug in one of these, we'll have to remember to change both",   so I moved the click handler to the parent <li> element. 
However, for your second concern "Next, this pattern will have to be done for all the other menu items, not just toggling autocomplete", even I moved the handler to <li>, each of the <li> for the menu options still need to have unique identifiers to trigger functions to manipulate DOM on their corresponding sliders. 
I have made some changes to the code based on the previous change I made. Maybe you can take a look and please let me know If that's the ideal implementation for this issue. below is a screen cast for the changes
final_toggle

views/editor/nav-options.html
```diff
@@ -119,7 +119,7 @@ <h2 class="snippets-menu-title">{{ gettext("snippetMenuTitle") }}</h2>
                 <div title="{{ gettext("navIncreaseTextSizeIconTitle") }}" id="editor-pane-nav-increase-font" class="font-size-button"></div>
               </div>
             </li>
-            <li>
+            <li id='colortheme-toggle-li'>
```
humphd: Let's use a different naming convention here that's more about what it is vs. the type, since that could change (also note the use of double vs. single quotes):

```html
<li id="editor-option-colortheme">
...
<li id="editor-option-wraptext">
...
```

flukeout:
Okay, cool! I left some comments for you @feihaozi77

feihaozi77:
@flukeout I just fixed the cursor

feihaozi77:
@flukeout Please review my changes whenever you are available

flukeout:
@feihaozi77 - testing now!

flukeout:
@feihaozi77 - OK this works great, thanks a lot! 👍 

@humphd Once you're happy, go ahead and merge.

humphd:
@feihaozi77 great fix, thanks for doing it!  I've merged this, and it will go live next time we push to production.  In the meantime, you should be able to test this on our staging server within the hour:

https://bramble.mofostaging.net/en-US/

flukeout:
Yeah, nice work here @feihaozi77 - much appreciated.

PR #2511, opened 2017-10-02
Fixes #2102 - When switching locales, stay on the same page

### LINKED ISSUE #2102: When switching locales, stay on the same page.
Right now if you navigate to the features page, and then switch the locale, you end up on the homepage. We should keep users on the same page.

Cryzek:
Made changes suggested by @gideonthomas

### DIFF
--- a/server/routes/main/features.js
+++ b/server/routes/main/features.js
@@ -12,7 +12,7 @@ module.exports = function(config, req, res) {
     loginURL: config.appURL + "/" + locale + "/login",
     editorHOST: config.editorHOST,
     editorURL: config.editorURL,
-    URL_PATHNAME: "/" + qs,
+    URL_PATHNAME: "/features/" + qs,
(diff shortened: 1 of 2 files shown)


flukeout:
Testing this out now.

flukeout:
@Cryzek - is this change working for you? I'm running your PR branch locally and whenever I switch locales on the **Get Involved** or **Features** page, I end up on the homepage. Let me know. Maybe I'm missing some kind of step.

Cryzek:
Yes, it is working properly. 
work1

Could it be a caching problem? Are server-side pages cached?

Cryzek:
Thanks @gideonthomas @flukeout. :blush:

flukeout:
Re-testing now, will report back.

flukeout:
Well, it works as expected! Thanks so much @Cryzek - merging!

Issue #2509, opened 2017-09-29
Test for Brackets when starting Thimble


humphd:
Follow-up from https://github.com/mozilla/thimble.mozilla.org/issues/2355#issuecomment-333153037: let's have Thimble test to see if Brackets is running when the server starts, since it's a runtime dep that we assume is there.  This tends to trip-up new devs who forget (or don't know) to run them both.

Let's use https://github.com/sindresorhus/is-reachable.  The general idea here would be:

* install and save the `is-reachable` module: `npm install --save is-reachable`
* Alter the [server code](https://github.com/mozilla/thimble.mozilla.org/tree/master/server) to do a check to see if the [`BRAMBLE_URI` URL defined in the `env`](https://github.com/mozilla/thimble.mozilla.org/blob/7a6c2f8d492b5ade8c5a28e5bb93f37c20be2c3e/server/routes/config.js#L19).  @gideonthomas can tell you where he'd want that code to live specifically
* If the Brackets host isn't reachable, log a message indicating that the server should be running.

Assigning this to @MPierre9.  Ask questions here and we can help you through this.

gideonthomas:
you can add it somewhere at the bottom of this file: https://github.com/mozilla/thimble.mozilla.org/blob/master/server/index.js. And use `config.editorHOST` instead of testing the `env` for `BRAMBLE_URI`. You also don't need to check if it exists because it will throw an error earlier in the code if it is missing.

MPierre9:
I'll get started working on it, thanks @humphd and @gideonthomas.

MPierre9:
So I implemented what @humphd and @gideonthomas  suggested and have a working "prototype". 
![bramble bug fix](https://user-images.githubusercontent.com/27080760/31050337-6e4fe27e-a615-11e7-9854-f140abb02f6c.JPG)

The only odd thing I encountered is that when you run npm start and the build process bar is displayed the console.log() text will be repeated. 

![bramble bug 2](https://user-images.githubusercontent.com/27080760/31050347-be13014c-a615-11e7-8a25-15efbb0d6f80.JPG)


What do you guys think?

gideonthomas:
Huh, that is interesting...I'm not sure why that happens. Now that I think about it, it might make more sense to have this be a separate module. Here's my suggestion:
1. Take your code and move it to a new file inside the `scripts` folder called `verify-brackets.js`. Use `process.env.BRAMBLE_URI` instead this time.
[...]

PR #2514, opened 2017-10-02
Test for Brackets when starting Thimble #2509, Error cannot find module using require

MPierre9:
As suggested by @humphd I opened a PR for this issue which acts as a continuation of issue #2509. There seems to be an error when I do `const config = require('server/routes/config');`. The console says it can't find the module. 
require bug

The idea is to have the code test to see if the brackets server is running and output a notification. This will minimize the issue from happening to new developers (happened to me).


server/verify-brackets.js
```diff
@@ -0,0 +1,18 @@
+const config = require('server/routes/config');
+
+const isReachable = require('is-reachable');
+var brambleServer = config.editorHOST;
+
+console.log("Checking for Bramble Server. " + brambleServer);
+
+isReachable(brambleServer).then(reachable => {
+  
+  if(reachable)
```
humphd: In JS you want to always put the `{` on the same line as the statement.  There are some weird bugs you can get into with `return` and automatic semi-colon insertion, for example:

```js
return // actually returns `undefined`
{
  data: 1
};

// vs. 

return { // does what you want
  data: 1
};
```

server/verify-brackets.js
```diff
@@ -0,0 +1,18 @@
+const config = require('server/routes/config');
+
+const isReachable = require('is-reachable');
+var brambleServer = config.editorHOST;
+
+console.log("Checking for Bramble Server. " + brambleServer);
```
humphd: In the spirit of Unix, let's only print in the case of an error.  You can drop this.

server/verify-brackets.js
```diff
@@ -0,0 +1,18 @@
+const config = require('server/routes/config');
```
humphd: The way that node does its module loading is that you can pass it a module name, like `is-reachable` below, or you can pass it a path, which needs to start with `.` or `..` or the like.  You probably want this:

```js
const config = require('./routes/config');
```

server/verify-brackets.js
```diff
@@ -0,0 +1,18 @@
...
+
+isReachable(brambleServer).then(reachable => {
+  
+  if(reachable)
+  {
+    console.log("Bramble server is running.");    
+  }
+  else
+  {
+    console.log("Error: Bramble server is not running. Please end this process and run the Brackets server with: npm start");
```
humphd: Let's only print in the error case:

```js
if(!reachable) {
    console.warn("...");
}
```

server/verify-brackets.js
MPierre9: Oh I didn't even know that could happen with JS. Good to know.

MPierre9:
Thanks for the input and help @humphd I'll implement those changes and do another commit 👍

server/verify-brackets.js
MPierre9: Oh ok I was using the wrong path. I looked up the error and saw people using "./" but I was doing "./server/routes/config".

MPierre9:
Implemented the changes @humphd mentioned. It seems the Travis CI check doesn't like me outputting to the console. Here are some screenshots of the build working. 

working brackets 1


working brackets 2

gideonthomas:
@MPierre9, as per https://github.com/mozilla/thimble.mozilla.org/blob/master/CONTRIBUTING.md#submit-a-pull-request, run `npm run test:lint:fix` to fix your linting issues and commit the changes. That should make Travis happy.

scripts/verify-brackets.js
```diff
@@ -0,0 +1,11 @@
+const config = require("../server/routes/config");
+const isReachable = require("is-reachable");
+var brambleServer = config.editorHOST;
```
gideonthomas: nit: this should be a `const` since we don't change its value at any point.

MPierre9: Oh ok makes sense, Thanks!

scripts/verify-brackets.js
```diff
@@ -0,0 +1,14 @@
+const config = require("../server/routes/config");
+const isReachable = require("is-reachable");
+const brambleServer = config.editorHOST;
+const colors = require("colors");
+
+isReachable(brambleServer).then(reachable => {
+  if (!reachable) {
+    console.warn(
+      colors.yellow(
+        `Error: Bramble server is not running. Please end this process and run the Brackets server with: npm start`
```
gideonthomas: ah darn, sorry I missed this, could you change this to:
```
`Error: Brackets is not running. In a separate terminal window, run Brackets using npm start in the brackets folder.`
```

MPierre9: No worries, I'm on it 👍

gideonthomas:
Awesome work @MPierre9! Thanks for sticking with it.

MPierre9:
Thanks for all the help @gideonthomas :)

Issue #2520, opened 2017-10-05
Add is-reachable to package.json


gideonthomas:
cc @MPierre9 

the module you added should be in `package.json` as a `devDependency`. Would you be able to add that @MPierre9?

humphd:
@MPierre9 see https://docs.npmjs.com/files/package.json#devdependencies for info.

MPierre9:
@gideonthomas For sure I'll get started on it. Thanks for the extra info @humphd

PR #2521, opened 2017-10-05
Added is-reachable to package.json follow up to issue #2520

MPierre9:
As a follow up to issue #2520 and suggested by @gideonthomas I added is-reachable to the devDependencies section of package.json.


MPierre9:
@humphd That would be great thanks! Also would this do the trick setting to compare only the newest commit?
git commit

humphd:
We haven't covered this workflow yet, but here's how you solve this in the future (I can help you fix it with what you have now, too)

### Standard Workflow

First, you have to update your `master` branch to be the same as Mozilla's.  To do so, you need to have a remote for Mozilla:

```
git remote add upstream https://github.com/mozilla/thimble.mozilla.org.git
```

Now switch to your `master` branch, and pull in the changes from Mozilla.  I notice that you've been working on your `master`, so this won't work.  You have to leave `master` clean (we'll fix that below):

```
git checkout master
git pull upstream master
```

Now create and checkout a new branch for your bug:

```
git checkout -b issue-1234
````

You can use whatever name you want instead of `issue-1234`.  Now you can do your fixes, commit, and then push this branch up to GitHub:

```
git push origin issue-1234
```

Then you do a PR from that branch on your fork.

### Fixing Your Case

Your repo has a few issues.  First, you made changes to your `master` branch that will diverge from Mozilla's.  You only ever want to work on a new "topic" branch (e.g., a branch that fixes one bug, or addresses one topic).

Luckily with git, you can't really mess anything up: it keeps a record of everything.  Here's what you should do.  Start by resetting your `master` branch to Mozilla's (note the Capital `B`):

```
git checkout -B master upstream/master
git pull upstream master
```

Now your local `master` is in sync with Mozilla's.  Next we make a branch for this work:

```
git checkout -b is-reachable-fix
```

Next, we can pull in your fix from before:

```
git cherry-pick 048929d
```

Now you'll have just your final commit on this branch, and you can commit, push, and do a PR.  You can close this one and do a new one if you want.

PR #2522, opened 2017-10-05
Added is-reachable to package.json

MPierre9:
Huge thanks to @humphd for helping me to use git for this is-reachable-fix. This is a continuation of issue #2520 to add is-reachable to package.json.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,11 +92,11 @@
   "devDependencies": {
     "eslint": "^4.6.1",
     "eslint-plugin-prettier": "^2.2.0",
+    "is-reachable":"^2.3.3",
     "prettier": "^1.6.1",
     "shx": "^0.2.2",
     "stylelint": "^8.1.1",
-    "stylelint-config-standard": "^17.0.0",
-    "is-reachable":"^2.3.3"
+    "stylelint-config-standard": "^17.0.0"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


humphd:
I'll let @gideonthomas do the final review, but one more tip.  It looks like you did this by hand, which would seem to make sense.  However, there's a trick you can use, `npm install --save <package-name>` or `npm install --save-dev <package-name>`.  The first would add it to `dependencies` and the second to `devDependencies` in `package.json`, as well as install it.

Glad you were able to get the git stuff worked you, nice job.

MPierre9:
Thanks @humphd for the tips, good to know for the next time

package.json
```diff
@@ -95,7 +95,8 @@
     "prettier": "^1.6.1",
     "shx": "^0.2.2",
     "stylelint": "^8.1.1",
-    "stylelint-config-standard": "^17.0.0"
+    "stylelint-config-standard": "^17.0.0",
+    "is-reachable":"^2.3.3"
```
gideonthomas: can you move this to below `"eslint-plugin-prettier"` so that it's alphabetical and doesn't move somewhere else when we do `npm install --save`.

gideonthomas:
Thanks for doing this @MPierre9!

PR #2523, opened 2017-10-08
Update publish popup title when project is live - Issue 1918 

### LINKED ISSUE #1918: Change Publish popup title when project is Live
Currently, the popup title always says "Publish your Project", even if it is already published...

image

Let's change it to "Your Project is Online" when the project is published.

marcobeltempo:
Fixes #1918 
The original issue was that the publish popup title would always display **"Publish your Project"** regards if it was live or not.

@flukeout @humphd My solution was to reuse the existing `h1`  in "_views/editor/publish.html_" to assign the value depending on the current state of the project performed on the server side. 
Unpublished title = **"Publish your Project"**  
Published title = **" Your Project is Online"**   

1. created var `TEXT_PUBLISHED_HEADER = strings.get("publishedHeader");` holds the string value from 
_"locales/en-US/server.properties"_

2. added publishHeader to the Publisher() object in "\public\editor\scripts\ui\publisher.js". This gets the `<h1>` id in _"views/editor/publish.html"_
 
3.  `Publisher.prototype.updateDialog` located in  _"views/editor/publish.html"_ already checks the handler of the project status.

4. If the project is published,  the header gets assigned `TEXT_PUBLISHED_HEADER` from step 1

5. If the project is unpublished the header is set to default

thimble_issue1918_fix_capture

### DIFF
--- a/locales/en-US/server.properties
+++ b/locales/en-US/server.properties
@@ -212,7 +212,6 @@ snippetJSChangeStyleTitle=Change a CSS property of an element
 snippetJSChangeStyleComment=Select the element with id='alert'
 
 # Publishing
-publishHeader=Publish your Project
 publishShareLink=Here's a link you can share…
 publishChangesPrompt=You've made changes since you last published this project.
 publishProjectDescription=Project Description
--- a/public/editor/scripts/ui/publisher.js
+++ b/public/editor/scripts/ui/publisher.js
@@ -14,6 +14,8 @@ var TEXT_UNPUBLISH = strings.get("publishDeleteBtn");
 var TEXT_UNPUBLISHING = strings.get("publishUnpublishingIndicator");
 var TEXT_UPDATE_PUBLISH = strings.get("publishChangesBtn");
 var TEXT_UNPUBLISH_WARNING = strings.get("publishDeleteWarning");
+var TEXT_PUBLISH_HEADER = strings.get("publishHeader");
+var TEXT_PUBLISH_HEADER_ONLINE = strings.get("publishHeaderOnline");
 
 function unpublishedChangesPrompt() {
   var dialog = this.dialog;
@@ -34,6 +36,7 @@ function Publisher() {
       indexMessage: $("#no-index")
     },
     description: $("#publish-details > textarea.publish-description"),
+    publishHeader: $(".content > h1"),
     published: {
       link: $("#publish-link > a"),
       changed: $("#publish-changes"),
@@ -319,10 +322,12 @@ Publisher.prototype.updateDialog = function(publishUrl, allowUnpublish) {
   // "publish"/"cancel" buttons
   if (allowUnpublish) {
     this.dialog.buttons.parent.addClass("hide");
+    this.dialog.publishHeader.text(TEXT_PUBLISH_HEADER_ONLINE);
     unpublishBtn.off("click", unpublish).on("click", unpublish);
     published.container.removeClass("hide");
   } else {
     published.container.addClass("hide");
+    this.dialog.publishHeader.text(TEXT_PUBLISH_HEADER);
   }
 };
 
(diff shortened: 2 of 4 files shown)


locales/en-US/server.properties
```diff
@@ -213,6 +213,7 @@ snippetJSChangeStyleComment=Select the element with id='alert'
 
 # Publishing
 publishHeader=Publish your Project
+publisedhHeader=Your Project is Online
```
gideonthomas: since you're using this in the client files you should add this string to the client.properties file instead as per https://github.com/mozilla/thimble.mozilla.org/wiki/Localization#client-side-localization.
Also rename this to `publishHeaderOnline`

humphd:
@marcobeltempo before we go to land this, I need to show you how to rebase your work, since it looks like you've got a bunch of commits in here that aren't part of your patch (did you merge with master, maybe?).  It's not a problem, we can fix it, but FYI, you don't need to do that while you're developing.

marcobeltempo:
@humphd that would be great! I made a mess somewhere down the line and a bigger one trying to fix it.

humphd:
Yeah, it's easily fixed, don't worry.  For now, just keep getting your PR fixed.  Then I'll give you a hand.  Remind me please, I'll forget, but I need to cover this with all of you.

locales/en-US/client.properties
```diff
@@ -27,6 +27,7 @@ renameProjectSavingIndicator=Saving…
 publishPublishingIndicator=Publishing…
 publishUnpublishingIndicator=Unpublishing…
 publishDeleteWarning=Are you sure? The current link will stop working and if you publish again, the link will be different.
+publishHeaderOnline=Your Project is Online
```
gideonthomas: also move the `publishHeader` string from `server.properties` to `server-client-shared.properties` since we are going to use that string in both the server and the client.

public/editor/scripts/ui/publisher.js
```diff
@@ -14,6 +14,7 @@ var TEXT_UNPUBLISH = strings.get("publishDeleteBtn");
 var TEXT_UNPUBLISHING = strings.get("publishUnpublishingIndicator");
 var TEXT_UPDATE_PUBLISH = strings.get("publishChangesBtn");
 var TEXT_UNPUBLISH_WARNING = strings.get("publishDeleteWarning");
+var TEXT_PUBLISHED_HEADER = strings.get("publishHeaderOnline");
```
gideonthomas: let's call the `TEXT_PUBLISH_HEADER_ONLINE`
and add another variable here called `TEXT_PUBLISH_HEADER` that stores the `publishHeader` string

public/editor/scripts/ui/publisher.js
```diff
@@ -319,6 +321,7 @@ Publisher.prototype.updateDialog = function(publishUrl, allowUnpublish) {
   // "publish"/"cancel" buttons
   if (allowUnpublish) {
     this.dialog.buttons.parent.addClass("hide");
+    this.dialog.publishHeader.text(TEXT_PUBLISHED_HEADER);
```
gideonthomas: you also need to set the title in the else condition back to `TEXT_PUBLISH_HEADER`

views/editor/publish.html
```diff
@@ -41,4 +41,4 @@
     <div id="publish-button-publish">{{ gettext("publishBtn") }}</div>
   </div>
 </div>
-{% endblock %}
+{% endblock %}
```
gideonthomas: revert this change

marcobeltempo:
@gideonthomas Thank you for all your feedback! I've got it working again on my localhost. I'll commit shortly with the revisions

flukeout:
Looking forward to testing this. Please ping me when it's ready @marcobeltempo!

marcobeltempo:
Everything's ready for testing @flukeout

gideonthomas:
@marcobeltempo can you rebase this PR onto master? You seem to have some extra commits in here.

marcobeltempo:
@gideonthomas I rebased the PR onto master

gideonthomas:
This looks great! Thanks for the contribution @marcobeltempo!

marcobeltempo:
@gideonthomas Thanks again for all your help! 😃

Issue #2355, opened 2017-07-09
Unable to publish without index.html gives no feedback


gideonthomas:
Looks like a regression on both staging and prod!

STR:
1. Login and start a new project
2. Delete the index.html file
3. Click Publish

The publish dialog box vibrates but shows no explanation (which I think we had at the bottom of the box).

gideonthomas:
cc @flukeout

flukeout:
Hey @gideonthomas - do you mean click "Update Published" after deleting the index.html file?

gideonthomas:
yeah sorry, I meant the "Update Published Version"

humphd:
@flukeout @gideonthomas do you think this is a good bug for @MPierre9 to work on?  If so, can you provide some guidance on where to begin?

flukeout:
Yeah, I think this would be a good one!

MPierre9:
@humphd @flukeout Thanks! I'll get started :)

MPierre9:
@flukeout I got Thimble running on my local machine using the instructions on GitHub, however when I go to make new project and recreate the bug it hangs on "Creating new project" am I missing a step or installation instruction. Thanks

Also is there a slack channel I can ask questions?

humphd:
@MPierre9 hop into https://chat.mozillafoundation.org/mozilla/channels/thimble

flukeout:
That's interesting - when you say it hangs, what are you seeing? Can you post any errors you are seeing in your console? Users run into this once in a while, so it's handy that you are seeing it too so maybe we can debug together.

MPierre9:
It was stuck at loading project but I figured it out, I wasn't running the brackets server and thimble server at the same time *facepalm*

humphd:
@MPierre9 yeah, that's something a lot of people run into.  One thing we could do is have the Thimble node.js server check that the Brackets host is available on startup, and print some kind of useful error message if it isn't.  If you want to file and fix that bug, I'll help you figure out what to do.

MPierre9:
@humphd That sounds like a good idea I don't have a lot of experience with node.js but I am very familiar with other server side languages like PHP and am willing to jump in to fix that bug. Any pointers are appreciated :)

humphd:
@MPierre9 I filed https://github.com/mozilla/thimble.mozilla.org/issues/2509 for you.

humphd:
This is a test.

[...]

PR #2526, opened 2017-10-11
Fixes issue #2355, adds no index message when updating project without index.html

MPierre9:
As a continuation of issue #2355 this allows the no index.html message to show when updating a publish. 

Mainly with jQuery I show `#publish-buttons` then remove the publish and cancel buttons to avoid having two publish buttons (update published version and publish). 

no index message

### DIFF
--- a/locales/en-US/server.properties
+++ b/locales/en-US/server.properties
@@ -217,7 +217,7 @@ publishShareLink=Here's a link you can share…
 publishChangesPrompt=You've made changes since you last published this project.
 publishProjectDescription=Project Description
 publishCancelBtn=Cancel
-noIndexFound=Please add an <strong>"index.html"</strong> file before publishing your project. <a href="{{ learnMoreURL }}">Learn more</a>
+noIndexFound_2=Published projects need an <strong>"index.html"</strong> page, please add it and try again. <a href="{{ learnMoreURL }}">Learn more</a>
 
 ##################
 ## Project List ##
--- a/public/editor/scripts/ui/index.js
+++ b/public/editor/scripts/ui/index.js
@@ -445,6 +445,11 @@ function init(bramble, csrfToken, appUrl) {
     _escKeyHandler = null;
   }
   function showPublishDialog() {
+    $("#updateDialog").hide();
+    $("#no-index-update")
+      .addClass("hide")
+      .removeClass("show");
+
     publishDialogUnderlay = new Underlay("#publish-dialog", hidePublishDialog);
     $("#publish-dialog").show();
 
--- a/public/editor/scripts/ui/publisher.js
+++ b/public/editor/scripts/ui/publisher.js
@@ -145,13 +145,12 @@ Publisher.prototype.publish = function(bramble) {
       publisher.dialogEl.removeClass("cannot-publish");
       publisher.dialogEl.width(publisher.dialogEl.width());
     }
-    if ($('#publish-button-update').is(":visible"))
-    {
-      $('#publish-buttons').show();
-      $('#publish-buttons').css({ 'border-top' : '0px' });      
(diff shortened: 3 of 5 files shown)


flukeout:
Hello @MPierre9 - thanks for opening the PR. I'll test it out now.

flukeout:
Okay this is working well as far as I can tell, and it's nice that we can just re-use the old UI we already designed for this. One issue I have though is that the wording feels a bit off...

``Please add an "index.html" file before publishing your project.``

Since in the Update case the project is already published, it doesn't feel right. Would it be possible to change the text that we are displaying there to something like?

``Before updating your published project, please add an "index.html" file.``

I wonder if there is a phrase we could use that would cover both instances. Anyway, let me know and we can finalize the wording.

MPierre9:
Thanks @flukeout, I think its a good idea to make the phrase cover both instances. In that event it could be something like: 

`Please add an "index.html" to continue publishing`

Or

`Oops you forgot to add an index.html, please add one to continue`

I think the workflow to add to a separate message for updating would be:

1. Add an noIndexUpdate message to messages.json 

2. In publish.html make a div similar to `<div id="publish-buttons"></div>` however without the buttons and including the noIndexUpdate message from messages.json

3. Use jQuery to show the "noIndexUpdate" div in the event that the user is updating and has no index.html

^ Is that kind of the right track? 

Overall I think either one could work.

flukeout:
Thanks @MPierre9 - I like your suggestions, so let's go with just one phrase for both instances, I think that will be easier overall.

How about... `Published projects need an "index.html" page, please add it and try again. Learn more.` - what do you think?

Since we'll be making a change to a localized string, we'll need to update the name of the key in the strings file, this one will be in `server.properties`. It's the `noIndexFound` string, and we'll need to update the key name to something like `noIndexFound_2` - this key name will also need to be updated in the dialog wherever it is referenced. The reason we do this is to ensure that our translation system picks up the change and then our localization community will re-translate it for other languages.

MPierre9:
@flukeout I think that message is great for both use cases. I added it in my newest commit. Thanks for the feedback!

flukeout:
Great, trying it out again to make sure it all works as expected.

flukeout:
@MPierre9 - do you know the steps for generating the localization strings locally? The way I used to do it doesn't work anymore.

MPierre9:
@flukeout - I don't think so, I'm using Visual Studio Code so I searched for `noIndexFound` in the project and changed the strings manually. There was a bunch of `messages.json` that I had to change.

humphd:
@gideonthomas can you help here?  ^^

gideonthomas:
@flukeout it's just `npm start`. It will generate everything for you. If you're looking for the specific command that only generates the l10n files, it's `npm run build:l10n`

MPierre9:
@flukeout Is there anything I need to correct?

flukeout:
@MPierre9 - no I think we're all good here. I just want to run it locally and confirm everything is 👍

MPierre9:
Thanks @gideonthomas! I'll trying to fix that issue with publish buttons not showing with what you said. Good thing you noticed it 👍

MPierre9:
After trying to fix the issue @gideonthomas outlined it ended up being less of a headache to just write separate html and jQuery to deal with the update no index found use case. This way all the existing functionality with the no index found publish use case is kept intact. I haven't found any bugs with this so far but an extra set of eyes always helps.

MPierre9:
Hello @gideonthomas just wondering if I'm on the right track with my latest fix (a47ae31). Thanks :)

gideonthomas:
Thanks @MPierre9, this works like a charm!

MPierre9:
Thanks @gideonthomas! :)

Issue #2479, opened 2017-09-11
Allow user to change the project description without needing to publish their project.


flukeout:
Currently in the publish dialog, there is a Description field. However, it only saves what you put in there at the moment of publishing, which isn't ideal. Sometimes people don't want to publish a project, but want to leave a description.

I think we should just auto-save whatever is in the description when a user closes the Publish dialog.

jmrodriguesgoncalves:
Hello, I am a Seneca student working on a project for my Open Source course. I am looking around for ways in which to help out, and this looks like something I want to try working on. This would be my first bug so I am pretty new. Guidance would be much appreciated!

humphd:
@flukeout, @gideonthomas can you give some technical insights here on what you want @jmrodriguesgoncalves to do in order to begin?  Thanks.

gideonthomas:
It's all yours @jmrodriguesgoncalves.  I would suggest starting by setting up Thimble on your machine for development. You can find instructions on how to do that in the [README](https://github.com/mozilla/thimble.mozilla.org/blob/master/README.md). 

Once you have Thimble setup, make sure you are able to reproduce the issue mentioned above - projects descriptions are not saved until you publish a project. 

To fix this issue, I believe you need to add a function to `Publisher.prototype` (like https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/ui/publisher.js#L47) that calls the `Metadata.update` function from the `project/metadata` module. A similar call can be found here: https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/ui/publisher.js#L47. Then, in the `init` function of the `Publisher` module, setup an event listener on `this.dialogEl` and call the function you created below.

I think that should fix your issue, but let us know if you have any issues and we'd be happy to help out.

jmrodriguesgoncalves:
@gideonthomas many thanks, I'm pretty sure you just about answered all of the questions I had in my mind. I will report back if I run into any issues!

Avedis777:
I'm a student at Seneca college learning open source, and I was hoping to work on this bug for my course.  If no one else is currently working on it, I'd like to give it a try.

jmrodriguesgoncalves:
hey, @Avedis777 currently working on this one. Keep your notifications opened in case I run into some trouble but for now I got it. You can feel free to come and have a look at the code when we are in class!

jmrodriguesgoncalves:
[...]

PR #2528, opened 2017-10-14
Fixing #2479: Allow user to change the project description without needing to publish their project.

jmrodriguesgoncalves:
Description wasn't saving if the project wasn't published. Now, by clicking cancel while the project is unpublished, the description is saved. Please review this and let me know if there's something that can be fixed.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -11,7 +11,7 @@ Vagrant.configure("2") do |config|
   # VM, but users are encouraged to test this and make adjustments below (or file PRs)
   # if you find the VM lagging or unresponsive.
   config.vm.provider "virtualbox" do |v|
-    v.memory = 1536
+    v.memory = 8000
     v.cpus = 1
     v.customize ["setextradata", :id, "VBoxInternal2/SharedFoldersEnableSymlinksCreate/vagrant", "1"]
   end
--- a/public/editor/scripts/ui/publisher.js
+++ b/public/editor/scripts/ui/publisher.js
@@ -29,6 +29,7 @@ function Publisher() {
     buttons: {
       publish: $("#publish-button-publish"),
       update: $("#publish-button-update"),
+      saveDescription: $("#publish-button-cancel"), /*Added Cancel button*/
       unpublish: $("#publish-button-unpublish"),
       parent: $("#publish-buttons"),
       indexMessage: $("#no-index")
@@ -54,6 +55,7 @@ Publisher.prototype.init = function(bramble) {
   publisher.handlers = {
     publish: publisher.publish.bind(publisher, bramble),
     unpublish: publisher.unpublish.bind(publisher),
+    setDesc: publisher.setDesc.bind(publisher),
     unpublishedChangesPrompt: unpublishedChangesPrompt.bind(publisher)
   };
 
@@ -78,6 +80,10 @@ Publisher.prototype.init = function(bramble) {
     publisher.showUnpublishedChangesPrompt();
   });
 
+  //saves the description and displays message to user telling them they have made changes
+  //and suggesting they should publish their project.
+  dialog.buttons.saveDescription.on("click", publisher.handlers.setDesc);
+  
   dialog.buttons.publish.on("click", publisher.handlers.publish);
 
   // Were there any files that were updated and not published?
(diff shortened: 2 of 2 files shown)


Vagrantfile
```diff
@@ -11,7 +11,7 @@ Vagrant.configure("2") do |config|
   # VM, but users are encouraged to test this and make adjustments below (or file PRs)
   # if you find the VM lagging or unresponsive.
   config.vm.provider "virtualbox" do |v|
-    v.memory = 1536
+    v.memory = 8000
```
flukeout: Does this change need to be part of this PR?

jmrodriguesgoncalves: Not at all, I should actually have left that as it was, that's something I forgot to change. When testing, the loading between pages took a little too long so I thought giving the VM a bunch of space would make things faster on my end. It didn't change anything, and it slipped right by me. We can change it to the way it was before.

flukeout: Okay yeah, please change it back, thanks!

flukeout:
OK - thanks for submitting this PR @jmrodriguesgoncalves!

My suggestion for this is to save the description when someone clicks outside of the Publish dialog and the dialog closes it. Currently, if I change the description and then click outside of the popup dialog, the change is lost. Hitting the `Cancel` button doesn't feel like it should save the description, which it does currently. So, in summary I'd recommend...

* Save the new description if someone clicks outside of the publish dialog
* Throw away the description of someone clicks the `Cancel` button

Let me know your thoughts!

jmrodriguesgoncalves:
I think that sounds good, I can try to do what you are asking me. Clicking outside would save the description as you said, so we'd use the same code we have now, except, when clicking anywhere outside the Publish dialog. 

Now, if we click the cancel button, should the description be completely erased, as in, making it blank, or should we have it be the way it was before, which basically kept the string intact while editing the project (so one would still see it there if they were to open the publish dialog again), but not saved if not published (so it would not show up when looking at their list of projects, or if closed and opened again for editing)?

flukeout:
Yeah the `Cancel` case is a bit strange for us. I just thought of a great solution however...

What if we just remove the `Cancel` button from the UI altogether? That way, you HAVE to click outside of the popup to close it, at which point we'll just save whatever changes you made.

What do you think? Simplifies things a bit for us.

jmrodriguesgoncalves:
I think that is a great idea that not only would simplify the code, but would also be much less confusing for the user.

If we are to go ahead with this idea, I have a quick question for you since you have more familiarity with this code. Just like we talked about on the chat, I am struggling to find the function that deals with the click outside of the dialog box. I have tried setting an event listener for the div under it, "#click-underlay", however that does not seem to work.

If you have any input and if we are to go ahead with the UI change as well, let me know and I will get working on it. Thanks again!

flukeout:
Hey @jmrodriguesgoncalves - yeah let's def go ahead with removing the `Cancel` button then. I'll dig into the code and look for the click-underlay stuff

flukeout:
OK @jmrodriguesgoncalves - I think I found what we need.

Check out [this line here](https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/ui/index.js#L437). 

That's where `hidePublishDialog()` is defined, which is then passed into the `new Underlay` constructor on line `448` - when the Underlay is clicked, it removes the the underlay and fires that function - I tested this briefly with a `console.log` to make sure it works. I think you can add your logic there. Hope that helps - let me know.

jmrodriguesgoncalves:
@flukeout great stuff, never mind the latest 2 pushes, I will try this!

flukeout:
Yeah do what you need, you can just keep updating this PR, we can squash it all into one commit when we merge it.

jmrodriguesgoncalves:
@flukeout So currently, the code does everything that we decided to go with. It saves the description when the user clicks outside of the dialog box, and since the cancel button isn't there anymore, the user won't be as confused.

On the back-end however, the solution I used to make this work may not be the most elegant. In order to save the description, the function checks whether or not the project the user is working on is currently published or not. If it isn't, an ajax request is made to "unpublish" the unpublished project. This of course throws an error message, but the description is saved perfectly. If the project is published however, the code attempts an ajax request to publish with that same ID it is published on, changing only the description, which means nothing but that field will change. What this does is basically allow the user to change the description whether or not their project is published.

flukeout:
Hey awesome @jmrodriguesgoncalves - I'll test out the functionality now, and let's ping @gideonthomas so he can provide his thoughts on the back-end piece you asked about.

humphd:
@jmrodriguesgoncalves looks like there are some things you can fix in `publisher.js`, see https://travis-ci.org/mozilla/thimble.mozilla.org/builds/289235540#L483

flukeout:
@jmrodriguesgoncalves let us know when this is ready to re-test!

jmrodriguesgoncalves:
@humphd got them sorted out, there were some spacing issues!

@flukeout, I haven't changed the logic from what you have tested so there shouldn't be any surprises, but let me know! This latest commit has the same functionality as before, just the spacing is now done correctly as to pass the two tests from below

flukeout:
OK, I'll wait on @humphd to weigh in on the changes you made and then I'll do one last functional test before we get this landed. Cheers!

public/editor/scripts/ui/publisher.js
```diff
@@ -130,6 +131,50 @@ Publisher.prototype.showUnpublishedChangesPrompt = function(callback) {
   });
 };
 
+//Using generateRequest("unpublish") in order to sync description
+//throws error since nothing is being unpublished, however this works
+//to save description. This is currently a "hack", a more elegant
+//solution may be implemented after
+Publisher.prototype.setDescription = function() {
+  var publisher = this;
```
gideonthomas: Let's not force a "publish" just to save the description. Instead, let's make the following changes:

To this function:
```js
const publisher = this;
const oldDescription = Project.getDescription();
const description = publisher.dialog.description.val();

if(oldDescription === description) {
  return;
}

Project.setDescription(description);

const data = {
  title: Project.getTitle(),
  description,
  dateUpdated: new Date().toISOString()
};

Metadata.update({
  update: true,
  csrfToken: publisher.csrfToken,
  host: Project.getHost(),
  id: Project.getId(),
  data 
}, error => {
  console.error("[Thimble] Failed to update project description with: ", error);
});
```

public/editor/scripts/ui/publisher.js
```diff
@@ -283,6 +328,7 @@ Publisher.prototype.unpublish = function() {
   request.always(function() {
     SyncState.completed();
     publisher.unpublishing = false;
+    publickCheck = false;
```
gideonthomas: what's this?

public/editor/scripts/ui/publisher.js
gideonthomas: you'll also need to require in the `metadata` module from https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/project/metadata.js

public/editor/scripts/ui/publisher.js
jmrodriguesgoncalves: Nothing, it has been removed.

public/editor/scripts/ui/publisher.js
jmrodriguesgoncalves: This does not to work, even with the metadata module included. After inspecting, it seems this is producing a bad request.

capture

PR #2532, opened 2017-10-19
tool-tips for settings menu works for the whole block now -issue 2152

scarecrow21:
Mouse-over for a settings switch tells you what it does. "Mouse overing" worked only for the toggle-switch. This is fixed in this PR. Tool tips now show if you mouse over on the text-label too.


views/editor/nav-options.html
```diff
@@ -112,14 +112,14 @@ <h2 class="snippets-menu-title">{{ gettext("snippetMenuTitle") }}</h2>
         </div>
         <div id="editor-pane-nav-options-menu" class="hide">
           <ul>
-            <li id="editor-option-textsize">
+            <li title="{{ gettext("Change the editor font size") }}" id="editor-option-textsize">
```
humphd: We need to add a new string to the `server.properties` file, which each locale can translate into the appropriate language.  You can add something like `navTextSizeTitle` with this string in https://github.com/mozilla/thimble.mozilla.org/blob/224ecbe154f5ab855e500db36fbecc003d31c272/locales/en-US/server.properties#L383-L414

views/editor/nav-options.html
```diff
@@ -112,14 +112,14 @@ <h2 class="snippets-menu-title">{{ gettext("snippetMenuTitle") }}</h2>
...
               <span>{{ gettext("navTextSize") }}</span>
               <div class="text-size-ui">
                 <div title="{{ gettext("navDecreaseTextSizeIconTitle") }}" id="editor-pane-nav-decrease-font" class="font-size-button"></div>
                 <div title="{{ gettext("navIncreaseTextSizeIconTitle") }}" id="editor-pane-nav-increase-font" class="font-size-button"></div>
               </div>
             </li>
-            <li id="editor-option-colortheme">
+            <li title="{{ gettext("Change the editor color theme") }}" id="editor-option-colortheme">
```
humphd: Same thing here, just add a new key/value, maybe `navColorThemeTitle` with this string.

scarecrow21:
I added a key-value pair in the server.properties folder and added the keys to the nav-options.html file. However, the "value" doesn't show when I mouse over. Only the "key" is displayed.

flukeout:
Testing this now

flukeout:
OK, this looks good to me. The tooltips show up as expected for all of the options in the dropdown menu. Good work @scarecrow21 

@humphd Can you re-review, I think the file you wanted removed is gone?

flukeout:
Ping @humphd for re-review.

humphd:
Yup, this looks great now. @gideonthomas, can you confirm the l10n side of this, and then land?  r+ from me, great job @scarecrow21.

Issue #2518, opened 2017-10-05
Home page content overflowing container on window resize


marcobeltempo:
OS: Windows 10
Browser: Firefox 57, Google Chrome v61.0

**Description**

When you resize the browser on the home screen with a width less than 765 pixels, the "Welcome video" and footer begin to overflow the content-wrapper.

![thimble_video_overflow_capture](https://user-images.githubusercontent.com/15523758/31206117-ea0e6f62-a943-11e7-9585-ec7efce23f21.PNG)

![thimble_footer_overflow_capture](https://user-images.githubusercontent.com/15523758/31206126-f51b97cc-a943-11e7-95a3-d3b510a01509.PNG)

**Steps to reproduce**

1. Go to home page [Thimble by Mozilla](https://thimble.mozilla.org/en-US/)
2. Resize window with a width less than 765px

![thimble_home_resize_overflow_capture](https://user-images.githubusercontent.com/15523758/31206109-e11e4f80-a943-11e7-8297-592aa7aa672f.gif)

humphd:
@flukeout thoughts?  Perhaps @marcobeltempo can also fix this.

Thanks for filing this!

marcobeltempo:
@humphd that would be great if I could work on this as well.

gideonthomas:
@flukeout based on your screenshot, the content wrapper is cropped, wouldn't fixing that fix this issue as well?

flukeout:
I'll check things out and report back within the hour.

flukeout:
@marcobeltempo If you want to tackle this, please do! I welcome any and all changes that improve the layout of the site when the window gets narrower.

A simple fix would just be to make sure we don't have the weird empty white area on the right that you can scroll to.

A fancier fix would make it so that the homepage gracefully adapts to a narrower viewport.

I think a simple fix would be good for now.

marcobeltempo:
@flukeout I would love to tackle this one! I’ll work on a simple fix and experiment with some other options in the future.

flukeout:
Great, it's yours

marcobeltempo:
@flukeout I followed your suggestion by adding `min-width:981px` to the body tag. There's still a horizontal scroll but it seems to solve the overflow issue.

![thimble_issue2518_capture](https://user-images.githubusercontent.com/15523758/31590096-b338cd1e-b1d8-11e7-9534-aabacd637dbe.gif)

[...]

PR #2534, opened 2017-10-20
Fixing #2518: Added a min-width to the homepage body to prevent content overflow

marcobeltempo:
### Issue
If you resize the browser window while on the home screen, elements such as the video, and footer will begin to overflow.

31206109-e11e4f80-a943-11e7-8297-592aa7aa672f

### Fix
I simply added a `max-width: 981px;` to the body tag in public/homepage/stylesheets/style.less

### Result
31590096-b338cd1e-b1d8-11e7-9534-aabacd637dbe


humphd:
Flagging @flukeout for review.

flukeout:
Testing this out now.

flukeout:
Thanks @marcobeltempo - tested this in a couple of browsers and it does the trick! Merging!

PR #2544, opened 2017-10-26
mouse-style is now a pointer in the settings menu. Issue-2543

scarecrow21:
The mouse-style is now a "pointer" throughout the settings  menu.

### DIFF
--- a/public/editor/stylesheets/editor.less
+++ b/public/editor/stylesheets/editor.less
@@ -539,10 +539,6 @@ body {
       cursor: pointer;
     }
 
-    #editor-option-textsize {
-      cursor: default;
-    }
-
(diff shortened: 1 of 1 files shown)


gideonthomas:
now that I think about it, I think this was done intentionally since the text size is not a toggle. When you click on the "Text Size" label, what do you do? Do you increase or decrease font size. Even if we choose one it won't really be intuitive so I think we made the label itself unclickable.

What do you think @flukeout?

flukeout:
Yeah, we left the cursor as a ``default`` cursor for the text size toggle since clicking on the label doesn't do anything, unlike for the other options, which are toggles.

PR #2568, opened 2017-12-02
[WIP] Extending light theme to UI above Editor and Files panes.

flukeout:
Looks like this so far...

image

Fixes #2746

### DIFF
--- a/public/editor/stylesheets/editor.less
+++ b/public/editor/stylesheets/editor.less
@@ -76,10 +76,14 @@ body {
   left: 0;
   z-index: 2;
   height: @toolbarHeight;
-  background-color: #212121;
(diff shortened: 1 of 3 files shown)


flukeout:
Progress is slow but steady...

image

public/editor/stylesheets/editor.less
```diff
@@ -52,11 +51,11 @@ body {
...
-  position: absolute;
-  display: block;
-  top: 0;
-  width: 100%;
-  height: 0;
+  // position: absolute;
+  // display: block;
+  // top: 0;
+  // width: 100%;
+  // height: 0;
```
humphd: Let's not do this.  Can you remove instead?

PR #2571, opened 2017-12-05
update /healthcheck to note the build commit - Issue 2181

### LINKED ISSUE #2181: update /healthcheck to note the build commit rather than version
we're not versioning thimble in a meaningful way: we should update healthcheck so that it lists its current commit instead, so that we actually know which version of Thimble we're looking at on staging or production.

ericsuhn:
This is for fix #2181 using [git-rev-sync](https://github.com/kurttheviking/git-rev-sync-js) to record the current commit in /healthcheck.

### DIFF
--- a/server/request.js
+++ b/server/request.js
@@ -7,7 +7,7 @@ let cookieParser = require("cookie-parser");
 let cookieSession = require("cookie-session");
 
 let version = require("../package").version;
-const gitRevSync = require("git-rev-sync").long();
+const commit = require("git-rev-sync").long();
 const Logger = require("./logger");
 
 const logFormats = {
@@ -104,7 +104,7 @@ Request.prototype = {
   },
   healthcheck() {
     this.server.get("/healthcheck", (req, res) => {
-      res.json({ http: "okay", version: version, commit: gitRevSync });
+      res.json({ http: "okay", version: version, commit: commit });
     });
 
     return this;
--- a/server/request.js
+++ b/server/request.js
@@ -7,7 +7,7 @@ let cookieParser = require("cookie-parser");
 let cookieSession = require("cookie-session");
 
 let version = require("../package").version;
-let commit = require("../node_modules/git-rev-sync/").long();
+const gitRevSync = require("git-rev-sync").long();
 const Logger = require("./logger");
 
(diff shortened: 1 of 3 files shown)


humphd:
@gideonthomas can you review this one please?

server/request.js
```diff
@@ -7,6 +7,7 @@ let cookieParser = require("cookie-parser");
 let cookieSession = require("cookie-session");
 
 let version = require("../package").version;
+let commit = require("../node_modules/git-rev-sync/").long();
```
humphd: You don't need to use `../node_modules/`, just do:

```js
let gitRevSync = require("git-rev-sync").long();
```

services/login.webmaker.org/app/http/controllers/site.js
```diff
@@ -38,14 +38,16 @@ exports.js = function (filename) {
  * GET health check for app
  */
 var version = require("../../../package").version;
+var commit = require("../../../node_modules/git-rev-sync/").long;
```
humphd: Same thing here, no pathing to get to `git-rev-sync`

gideonthomas: undo changes in this file...we shouldn't update services in here.

server/request.js
gideonthomas: +1 to not using "node_modules" in the path. But let's call `long()` only once.
So, it should look like: 
```js
const commit = require("git-rev-sync").long();
...
res.json({ http: "okay", version, commit });
```

ericsuhn:
Thanks for the review! I've committed the requested changes.

server/request.js
```diff
@@ -7,6 +7,7 @@ let cookieParser = require("cookie-parser");
 let cookieSession = require("cookie-session");
 
 let version = require("../package").version;
+const gitRevSync = require("git-rev-sync").long();
```
gideonthomas: @ericsuhn can you rename this to `commit`...then I think this should be good to land

ericsuhn:
done, thanks @gideonthomas

gideonthomas:
Thanks @ericsuhn! This looks great

PR #2586, opened 2017-12-23
Improve UX of clicking preview icons: Prevent drag

### LINKED ISSUE #1966: Improve UX of clicking small icons in the preview & editor bars.
Check out the screengrab below. I click the icon, then move the mouse only a few pixels, which causes the ```` used in the icon to star dragging, and also stops my click from registering. 

icon-drag

This leads the UI to sometimes not doing what you wanted, which is annoying. We should fix this. 

* Maybe instead of click events, we fire the UI on ``mousedown``
* Maybe we shouldn't use ```` tags in there, and just rely on background images
  * This will prevent the dragging behaviour
* Maybe something else...

pradeepgangwar:
- Prevents preview icons from being dragged.

Closes #1966

### DIFF
--- a/public/editor/scripts/ui/index.js
+++ b/public/editor/scripts/ui/index.js
@@ -366,7 +366,10 @@ function init(bramble, csrfToken, appUrl) {
   });
 
   // Preview Mode Toggle
-  $("#preview-pane-nav-desktop").click(function() {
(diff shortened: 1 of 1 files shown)


humphd:
I think this looks right.  @flukeout, are you cool with this?

PR #2652, opened 2018-08-14
[enhancement - docs] - mozilla/thimble.mozilla.org - thimble.mozilla.org/README.md

tmm2018:
#### [enhancement - indentation] - mozilla/thimble.mozilla.org - thimble.mozilla.org/README.md


Pomax:
This is actually correct english. "setup" is a noun, where as in this sentence it's used as verb "set" with preposition "up". They're very similar, but correctly used in this case.

Issue #2656, opened 2018-08-28
Export CTA dialog on homepage


gideonthomas:
**Behind feature flag - GLITCH_EXPORT_ENABLED**

Layout: https://projects.invisionapp.com/share/6YN7CXIKHQN#/screens/312632667_Homepage_Colour_Option_2
Comp reference: https://drive.google.com/file/d/1LRqVEFARgJ3o7J7RNm6sVlYplB4_jLnx/view?usp=sharing

A dialog should show up only for authenticated users when they land on the homepage that shows the CTA for migrating their projects. It should show up after a 2s delay (play around with it to see what looks smooth). 

Notes:
1. The date/month/year string in the text should be an environment variable (GLITCH_MIGRATION_DATE) so that we don't need any more code changes.
2. The button should take you to the project list page.
3. On dismissing the dialog, the dialog and overlay vanish and instead a banner (that cannot be closed) shows up (comps/implementation in a subsequent ticket).
4. Link at the bottom should be an environment variable (GLITCH_HOMEPAGE_DIALOG_BLOG_POST).
5. Strings should be localized.

Pomax:
closed by PR.

PR #2662, opened 2018-09-05
homepage CTA after 2 seconds

Pomax:
Implements the homepage CTA as per, but does not yet close, https://github.com/mozilla/thimble.mozilla.org/issues/2656

- still needs the additional two screens from the comp

### DIFF
--- a/public/homepage/scripts/glitch.js
+++ b/public/homepage/scripts/glitch.js
@@ -4,11 +4,14 @@ var $ = require("jquery");
 
 module.exports = {
   init: function() {
+    let banner = $(".glitch-banner");
     let cta = $(".glitch-cta.underlay");
     if (cta) {
-      setTimeout(() => cta.removeClass("hidden"), 2000);
       var close = $(".glitch-cta .cta-close");
-      close.click(() => cta.addClass("hidden"));
+      close.click(() => {
+        banner.removeClass("hidden");
+        cta.addClass("hidden");
+      });
     }
   }
 };
--- a/public/homepage/scripts/glitch.js
+++ b/public/homepage/scripts/glitch.js
@@ -1,0 +1,14 @@
+/* globals $: true */
+
+var $ = require("jquery");
+
+module.exports = {
+  init: function() {
+    let cta = $(".glitch-cta.underlay");
+    if (cta) {
+      setTimeout(() => cta.removeClass("hidden"), 2000);
+      var close = $(".glitch-cta .cta-close");
+      close.click(() => cta.addClass("hidden"));
+    }
+  }
+};
--- a/public/homepage/stylesheets/glitch-notice.less
+++ b/public/homepage/stylesheets/glitch-notice.less
@@ -61,3 +61,98 @@
     }
   }
 }
+
+.glitch-cta.underlay {
+  position: fixed;
+  top: 0;
+  left: 0;
+  right: 0;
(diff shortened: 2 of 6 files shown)


public/homepage/stylesheets/glitch-notice.less
```diff
@@ -61,3 +61,98 @@
...
+      display: inline-block;
+      font-family: Helvetica;
+      font-size: 18px;
+      font-weight: bold;
+      line-height: 1.2;
+      text-decoration: none;
+      color: black !important;
+      background: #7effcc;
+      border: 3px solid black;
+      border-radius: 3px;
```
gideonthomas: can this be `5px`?

public/homepage/stylesheets/glitch-notice.less
```diff
@@ -61,3 +61,98 @@
...
+      box-shadow: 3px 3px 0 0 black;
+      padding: 17px;
```
gideonthomas: can this be `15px 17px`?

public/homepage/stylesheets/glitch-notice.less
```diff
@@ -61,3 +61,98 @@
...
+      line-height: 1.4;
+      margin: 30px auto;
+      padding: 0;
+    }
+
+    p:last-child {
+      margin-bottom: 0;
+    }
+
+    .export-button {
```
gideonthomas: can you also have a hover state to have a background of `#70e8b8`?

public/homepage/scripts/glitch.js
```diff
@@ -0,0 +1,14 @@
+/* globals $: true */
+
+var $ = require("jquery");
+
+module.exports = {
+  init: function() {
+    let cta = $(".glitch-cta.underlay");
+    if (cta) {
+      setTimeout(() => cta.removeClass("hidden"), 2000);
```
gideonthomas: can we change this to `1000`?

views/homepage/glitch-notice.html
```diff
@@ -14,4 +14,19 @@
       </a>
     </div>
   </div>
+
+  <div class="glitch-cta underlay hidden">
+    <div class="glitch-cta content">
+      <img class="cta-close" src="/img/clear.svg" tabindex="0">
+      <i class="glitch-logo"></i>
+      <h1>{{ gettext("weAreMoving") | safe }}</h1>
+      <p>{{ gettext("takeAction")  | safe }}</p>
```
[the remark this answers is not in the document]
Pomax: it does, well spotted.

public/homepage/stylesheets/glitch-notice.less
Pomax: it can but I've deviated from the text in the comp in order to actually _match_ the comp so if I make it 5px, it'll actually look different. If we're fine with that, I'll change it.

public/homepage/stylesheets/glitch-notice.less
Pomax: I can, but that's the kind of stuff that should be in a comp, too because if that's what you want the button to do, I assume you want every button to do that, not just this one?

public/homepage/scripts/glitch.js
Pomax: I tried that, but this causes the dialog to pop up even before all the content has visually loaded, which is an odd experience.

Issue #2659, opened 2018-08-28
Export button on projects list page


gideonthomas:
**Behind feature flag - GLITCH_EXPORT_ENABLED**

Layout: https://projects.invisionapp.com/share/6YN7CXIKHQN#/screens/312632662_Project_List_Copy
Comp reference: https://drive.google.com/file/d/1LRqVEFARgJ3o7J7RNm6sVlYplB4_jLnx/view?usp=sharing

On rendering the project list, each project will also have a `glitch_migrated` boolean flag (will be implemented in publish, default to `false` for now). The remix and published version links will be replaced by either the "hooray" text or the button to export based on the boolean flag.

Notes:
1. The new sorting order for projects on this page will be based on whether the project has been exported or not.
2. The button when clicked will open up a dialog (implementation in subsequent issue) which will need to know the project id, and whether the project was published or not.
3. Strings should be localized.

PR #2663, opened 2018-09-05
Project list banner and export button stubs

### LINKED ISSUE #2658: Export CTA banner on project list page
**Behind feature flag - GLITCH_EXPORT_ENABLED**

Layout: https://projects.invisionapp.com/share/6YN7CXIKHQN#/screens/312632662_Project_List_Copy
Comp reference: https://drive.google.com/file/d/1LRqVEFARgJ3o7J7RNm6sVlYplB4_jLnx/view?usp=sharing

A persistent banner should show up on top of the projects list page.

Notes:
1. The date/month/year string in the text should be an environment variable (GLITCH_MIGRATION_DATE) so that we don't need any more code changes.
2. Link should be an environment variable (GLITCH_HOMEPAGE_DIALOG_BLOG_POST).
3. Strings should be localized.

Pomax:
closed #2658
stubs out #2659 

I had to guess the styling for the "hooray!" line, but far as I can tell it looks near identical.

This also moves the .less file so that it's used by both homepage and project list, but duplicates the actual glitch template HTML because they're slightly different and given these are effectively final changes to Thimble, having a bit of duplication in the glitch-specific code feels fine?

### DIFF
--- a/views/projects-list/index.html
+++ b/views/projects-list/index.html
@@ -68,20 +68,24 @@
           </a>
 
           <div class="project-ui glitch">
-            {% if project.publish_url %}
-            <a class="published-link" target="_blank" href="{{ project.publish_url }}">{{ gettext("publishedLink") }}</a>
-            <a class="remix-link" href="/projects/{{ project.published_id }}/remix">{{ gettext("remix") }}</a>
-            {% endif %}
-            {% if project.glitch_migrated %}
-            <div class="hooray">
-              Hooray! Your project is now on Glitch
-            </div>
+            {% if glitchExportEnabled %}
+              {% if project.glitch_migrated %}
+                <div class="hooray">
+                  Hooray! Your project is now on Glitch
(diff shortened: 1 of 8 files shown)


views/projects-list/index.html
```diff
@@ -63,11 +67,21 @@ <h2>{{ gettext("yourProjectsLink") }}</h2>
             <div class="project-description">{% if project.description %}{{ project.description | trim }}{% endif %}</div>
           </a>
 
-          <div class="project-ui">
+          <div class="project-ui glitch">
             {% if project.publish_url %}
```
gideonthomas: the remix and published link shouldn't show up anymore. So your logic might look like:
```django
{% if glitchExportEnabled %}
    {% if project.glitch_migrated %}
    ...
{% else %}
    {% if project.glitch_migrated %}
    ...
{% endif %}
```

Issue #2664, opened 2018-09-06
Export dialog in project list


gideonthomas:
**Behind feature flag - GLITCH_EXPORT_ENABLED**

**Layouts:** 
1. https://projects.invisionapp.com/share/6YN7CXIKHQN#/screens/312645378
2. https://invis.io/6YN7CXIKHQN#/312645377_Project_List_Copy_2
3. https://projects.invisionapp.com/share/6YN7CXIKHQN#/screens/312645379

**Comp reference:** https://drive.google.com/file/d/1LRqVEFARgJ3o7J7RNm6sVlYplB4_jLnx/view?usp=sharing

When the Export button is clicked in the project list, a dialog should pop up as per the layout above letting the user know about the export process for Glitch, and show buttons to continue with the export.

**Notes:**
1. Based on whether the export button was clicked for a project that has a published version or not, either layout 1 or 2 will be shown. If the project has a published version, layout 1 will be shown, otherwise layout 2 will be shown.
2. In layout 1, once one of the buttons is clicked, the other button will be greyed out as seen in layout 3 and the clicked button will say "Exporting..." (this text is also the same for layout 2).
3. On clicking the button a bunch of XHR requests happen as detailed below.
4. Strings should be localized.

**XHR Requests and Redirects**
1. An XHR POST request is made to the Thimble server to `/projects/<project id>/export/start/ ` (implementation pending) if the "Export project" button was clicked or to `/publishedprojects/<published project id>/export/start/` if the "Export published project" button was clicked.
   - The response from this route will be in the format: `{ "token": "<random hash>" }`. 
2. Once the token is received, the page will redirect to `<GLITCH_IMPORT_PROJECT_APP_URL>` (this should be an env var) with the following query params - 
   - `token` - The token received in the previous request
   - `id` - The project (or published project) id of the project that needs to be exported
   - `published` - A boolean indicating whether we're exporting the published project or the unpublished project.

Pomax:
It would be good if we also had some messaging around how to let the user know that our `.../start` route(s) failed:

- calling the URL works, but the result does not have a `{token:...}`
- calling the URL fails (e.g. the user's router has a melt-down after loading the projects page)

gideonthomas:
Hey @jgmac1106 this is something that Glitch might be able to do on their end when importing each project - you can probably [email them](mailto:support@glitch.com) and ask if they're able to support this.

Issue #2665, opened 2018-09-06
Export FAQ


gideonthomas:
**Behind feature flag - GLITCH_EXPORT_ENABLED**

**Layout:** https://invis.io/6YN7CXIKHQN#/312632663_Moving_To_Glitch_Copy

**Comp reference:** https://drive.google.com/file/d/1LRqVEFARgJ3o7J7RNm6sVlYplB4_jLnx/view?usp=sharing

This is an FAQ page that will answer questions that users might have about the migration and why we're doing it.

**Notes:**
1. Add a nav item called "Moving to Glitch".
2. For the styling, follow the same rules as the "Get Involved" page that already exists (as opposed to being strict to the layout above). So that means:
   - font-size for top paragraph and questions: 1.125em
   - font-size for rest of the text: 1em
   - line-height: 150%
   - color: rgba(0,0,0,.65)
   - font: Open Sans
   - regular font-weight: 400
   - link color: #268cb9
   - horizontal line color: #cacaca
   - top/bottom padding between questions - 20px
The rest can be approximations.
3. The email should be an env var - `GLITCH_SUPPORT_EMAIL`.
4. Strings should be localized.

For now use the following gif: 
![glitch-assets](https://user-images.githubusercontent.com/5816130/45183091-19082900-b1f1-11e8-9093-eeda5c4314eb.gif)

gideonthomas:
# Copy

Mozilla has made a strategic decision to sunset Thimble within the course of 2018-2019. To continue supporting our Thimble community, we have partnered with Glitch (created by [Fog Creek Software](https://www.fogcreek.com/)). [Read our blog post](https://about:blank) to learn more about this partnership.

To make this process simple for you, we have answered some common questions you might have. If you don’t find your question here, shoot us an email at [email@mozillafoundation.org](https://about:blank) and we’d be happy to chat with you.

### FAQ

**What do I need to do?**
We encourage you to export your projects over to Glitch. In the project list, you can transfer each project individually. Don’t forget to do this by December 1, 1989.

---

**What is Glitch?**
Glitch is a creative community that gives you the tools to create any app or website that you can imagine — and it's **free**! Like Thimble, it allows you to remix and tweak web projects to learn and build new content and put your own spin on it. [Click here](https://glitch.com/about/) to learn more about Glitch and what it can do.

---

**Is Glitch available in my language?**
Currently Glitch is translated in English. However, the Glitch team is committed to supporting more languages soon.

---

[...]

PR #2673, opened 2018-09-17
Project list export

Pomax:
See https://github.com/mozilla/thimble.mozilla.org/issues/2664

Strings have not yet been `locale/en-US`'d


views/projects-list/index.html
```diff
@@ -94,10 +96,43 @@ <h2>{{ gettext("yourProjectsLink") }}</h2>
...
+              <!-- click-through view: two buttons for 'last publish'  and 'current project' -->
+              <button class="btn export-button published hidden">
+                <i class="export-icon"></i><span class="label">Export published project</span>
+              </button>
+
+              <button class="btn export-button unpublished hidden">
+                <i class="export-icon"></i><span class="label">Export unpublished project</span>
+              </button>
+            </p>
+            <p><a class="" href="">Read Glitch's privacy policy</a></p>
```
gideonthomas: `href` is https://glitch.com/legal

public/resources/stylesheets/glitch.less
```diff
@@ -193,4 +160,89 @@
...
     }
   }
+
+  /* rule overrides for the projects page */
+  &.project-list {
+    margin-bottom: 5px;
+
+    .glitch-cta.underlay {
+      .glitch-cta.content {
+        width: 540px;
```
gideonthomas: can we change this to `480px`?

public/resources/stylesheets/glitch.less
```diff
@@ -193,4 +160,89 @@
...
+        padding: 55px 80px;
+
+        h1 {
+          font-size: 24px;
+          font-weight: 900;
+          line-height: 30px;
+          margin: 0;
+          padding-left: 1em; /* match the <li> indent */
```
gideonthomas: can we remove this? Commented on removing the `li` indent below.

gideonthomas: can we also set the `text-align` to `left`?

public/resources/stylesheets/glitch.less
```diff
@@ -193,4 +160,89 @@
...
+        }
+
+        ul {
+          text-align: left;
+          font-size: 14px;
+          line-height: 20px;
+
```
gideonthomas: can we add `padding-left` to `17px`?

public/resources/stylesheets/glitch.less
```diff
@@ -193,4 +160,89 @@
...
+          li {
+            padding-left: 2em;
```
gideonthomas: can we remove this?

public/resources/stylesheets/glitch.less
```diff
@@ -193,4 +160,89 @@
...
+            margin-bottom: 1em;
+          }
+        }
+
+        .export-button {
+          user-select: none;
+          width: 100%;
+
```
gideonthomas: can we add:
```css
    margin-left: 0;
    text-align: left;
    padding-left: 21px;
```

public/projects-list/scripts/glitch.js
```diff
@@ -4,10 +4,95 @@ var $ = require("jquery");
...
+      exportUnpublished = $("button.export-button.unpublished", content),
+      csrfToken = $("meta[name='csrf-token']").attr("content"),
+      importURL = $("meta[name='glitch-url']").attr("content");
+
+    if (banner.length && cta.length) {
       banner.removeClass("hidden");
     }
+
+    function getToken(url, onSuccess, onError) {
+      try {
```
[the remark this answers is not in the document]
Pomax: hmm, I was seeing it do weird things with the chain `getToken` is in without the try/catch, but I'll double check what's going on there.

public/projects-list/scripts/glitch.js
```diff
@@ -4,10 +4,101 @@ var $ = require("jquery");
 
 module.exports = {
   init: function() {
-    let banner = $(".glitch-banner");
-    let cta = $(".glitch-cta.underlay");
-    if (banner && cta) {
+    let banner = $(".glitch-banner"),
+      cta = $(".glitch-cta.underlay"),
+      content = $(".content", cta),
+      close = $(".cta-close", content),
```
gideonthomas: you probably need to add a listener for this too so that the text for the buttons resets.

STR:
1. Click on "export" for a project.
2. Click on "Start exporting", and watch it fail.
3. Close the dialog.
4. Reopen the dialog and it will still show the button saying "Exporting..."

views/projects-list/index.html
```diff
@@ -74,13 +76,13 @@ <h2>{{ gettext("yourProjectsLink") }}</h2>
...
-                <div class="btn export-button" title="{{ gettext("prjListMigrateProjectBtnTitle") }}">
+                <button class="btn export-button" data-project-id="{{ project.id }}" data-project-url="{{ project.publish_url }}" title="{{ gettext("prjListMigrateProjectBtnTitle") }}">
                   <i class="export-icon"></i>
                   {{ gettext("prjListMigrateProjectBtnTitle") }}
-                </div>
+                </button>
               {% endif %}
             {% else %}
-              {% if project.publish_url %}
+              {% if project.publish_url != None %}
```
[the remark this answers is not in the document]
Pomax: this is Jinja, not Python. `not ...` is not supported.

public/projects-list/scripts/glitch.js
Pomax: Hm, https://github.com/mozilla/thimble.mozilla.org/pull/2673/files/f18a7c98d140ab867696e5ffcc9d54ed8d733a9a#diff-188a9152d1759a40d1395fa702a49491R77 should already do that, I'll have a look at why it's not.

locales/en-US/server.properties
```diff
@@ -235,6 +235,18 @@ prjListMigrateProjectBtnTitle=Export to Glitch
 prjListCancelDeleteProjectBtn=Cancel
 publishedLink=Published Version
 deleteProjects=Delete Projects
+startExportingProject=Start exporting your project
+exportPublished=Export published project
+exportUnpublished=Export unpublished project
+readPP=Read Glitch’s privacy policy
+beforeYouStart=Before you continue, we would like you to know&hellip;
+exportOops=Oops, something went wrong
+retryOrMail=Please try again later or contact us at <a href="mailto:email@mozillafoundation.org">email@mozillafoundation.org</a>.
```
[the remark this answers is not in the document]
Pomax: hmm, which one? The locale file has `mailto:thimble@...` hardcoded in a million places, should I just use that one?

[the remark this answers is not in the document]
gideonthomas: Oh sorry, I thought https://github.com/mozilla/thimble.mozilla.org/issues/2665 was already done. In that ticket, we use an email that will be an env var called `GLITCH_SUPPORT_EMAIL`. We'll want to use the same one here.

public/projects-list/scripts/glitch.js
```diff
@@ -4,10 +4,102 @@ var $ = require("jquery");
...
       banner.removeClass("hidden");
     }
+
+    // We track the projectId "globally" for a dialog, because there
+    // can only ever be one dialog open at a time, triggered for
+    // one specific project to be exported.
+
+    let runOperation = (evt, button, published) => {
+      let root = published ? `publishedprojects` : `projects`;
+      let url = `/${root}/${projectId}/export/start`;
```
gideonthomas: you should only use project ID for the projects/ route, and for the publishedprojects/ route you should use the `publishedProjectId`. You'll have access to that on the server so you can render it in a data attribute or just parse it out of the publish_url which you already have access to

Pomax: Wait, why would those two not be linked? We update the content when you "re"publish, so your old page is gone and now it's the new page. Why would that need a separate publication id?

(I'll happily update the code but needing to call the API with a different id seems odd. The server should be able to do the thimbledb.project.id -> publishdb.project.id lookup without the client having to care?)

PR #2716, opened 2019-07-24
Setup environment variables for shutdown

gideonthomas:
The goal of this PR is to remove all elements from the UI that allow:
1. Creating new accounts
2. Publishing, Remixing, or creating new projects
based on checking their respective environment variables.

TO test: set the two new env vars to `true` to make sure they are working, run thimble locally with all dependent services (good luck), and then go through every page and ensure that there aren't any ways to use the above mentioned features using any UI. Also ensure that when you view a published project (publish one before testing with the env vars flipped to true), that the remix button doesn't appear.

Ignore the locale files (checked with Theo to make sure we can do a direct edit), the changes are meant to only remove the "Sign Up" text on the homepage.

### DIFF
--- a/locales/ach/server.properties
+++ b/locales/ach/server.properties
@@ -24,6 +24,7 @@ pageTitleFeatures=Thimble by Mozilla - Jami
 
 # Userbar
 signInPromptHomepage=<a href="#" title="Sign in to Thimble" id="login-link" data-loginUrl="{{ loginURL }}" class="navbar-login">Dony iyie</a> onyo <a href="#" title="Create an account" id="signup-link" class="navbar-login">Cwe akaunt</a>
+signInPromptHomepage2=<a href="#" title="Sign in to Thimble" id="login-link" data-loginUrl="{{ loginURL }}" class="navbar-login">Dony iyie</a>
 
 # Reload Preview Page
 errorCantLoadPreviewTitle=Aii, gin mo tye marac!
--- a/public/resources/remix/style.less
+++ b/public/resources/remix/style.less
@@ -21,6 +21,7 @@
 }
 
 .details-bar {
+  display: none !important;
   position: fixed !important;
   top: 0 !important;
   right: 0 !important;
--- a/public/resources/remix/style.less
+++ b/public/resources/remix/style.less
@@ -21,7 +21,6 @@
 }
 
 .details-bar {
-  display: none !important;
   position: fixed !important;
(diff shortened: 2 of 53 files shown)


server/routes/projects/read.js
```diff
@@ -44,12 +44,14 @@ module.exports = function(config, req, res, next) {
...
 
       if (response.statusCode === 404) {
-        // If there aren't any projects for this user, create one with a redirect
-        res.redirect(307, "/" + locale + "/projects/new" + qs);
-        return;
-      }
-
-      if (response.statusCode !== 200) {
+        if (config.shutdownNewProjectsAndPublishing) {
+          body = "[]";
```
Pomax: it feels like it might be more useful to have this be a redirect to the "we've moved to Glitch" page rather than simply presenting an empty list?

public/resources/remix/style.less
```diff
@@ -21,6 +21,7 @@
 }
 
 .details-bar {
+  display: none !important;
```
Pomax: This should not be necessary. Instead, the associated DOM element should probably simply not even get generated (if the env vars say it shouldn't be)?

locales/ach/server.properties
```diff
@@ -24,6 +24,7 @@ pageTitleFeatures=Thimble by Mozilla - Jami
 
 # Userbar
 signInPromptHomepage=<a href="#" title="Sign in to Thimble" id="login-link" data-loginUrl="{{ loginURL }}" class="navbar-login">Dony iyie</a> onyo <a href="#" title="Create an account" id="signup-link" class="navbar-login">Cwe akaunt</a>
+signInPromptHomepage2=<a href="#" title="Sign in to Thimble" id="login-link" data-loginUrl="{{ loginURL }}" class="navbar-login">Dony iyie</a>
```
Pomax: Do these still get updated automatically? (with the repo archived, I assume "no"). If not, do we need to get all of these locale strings confirmed before we land this? (not every language has grammar in which a subclause and that clause as full sentence use the same words, conjugations, and inflections...)

gideonthomas:
@TheoChevalier if you don't mind, would you be able to review the l10n part of this please?

public/resources/remix/style.less
gideonthomas: yeah I forgot that I need to do this based on an env var

gideonthomas:
thanks all!

